PR #3, opened 2023-06-28
API endpoint definition and openAPI spec import

nathanrussell-moj-digital:
This PR is some basic scaffolding in respect of defining a couple of API endpoints that we will need to call and importing their openAPI specs to generate the stubs and model classes


mattwills-moj-digital (approved):
LGTM

PR #6, opened 2023-06-28
Cypress tests now do basic accessibility tests

nathanrussell-moj-digital:
This PR adds cypress-axe and configures the integration tests to perform basic accessibility tests on each rendered page.
This will mean that we catch and can fix basic accessibility tests at development time, rather than waiting for the formal accessibility tests as part of GDS sign off etc

### DIFF
--- a/integration_tests/mockApis/auth.ts
+++ b/integration_tests/mockApis/auth.ts
@@ -61,7 +61,7 @@ const redirect = () =>
         'Content-Type': 'text/html',
         Location: 'http://localhost:3007/sign-in/callback?code=codexxxx&state=stateyyyy',
       },
-      body: '<html><body>SignIn page<h1>Sign in</h1></body></html>',
+      body: '<html lang="en"><head><title>Mock SignIn page</title></head><body><h1>Sign in</h1><span class="govuk-visually-hidden" id="pageId" data-qa="sign-in"></span></body></html>',
     },
   })
 
@@ -76,7 +76,7 @@ const signOut = () =>
       headers: {
         'Content-Type': 'text/html',
       },
-      body: '<html><body>SignIn page<h1>Sign in</h1></body></html>',
+      body: '<html lang="en"><head><title>Mock SignIn page</title></head><body><h1>Sign in</h1><span class="govuk-visually-hidden" id="pageId" data-qa="sign-in"></span></body></html>',
     },
   })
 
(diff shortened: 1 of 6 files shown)


integration_tests/mockApis/auth.ts
```diff
@@ -61,7 +61,7 @@ const redirect = () =>
         'Content-Type': 'text/html',
         Location: 'http://localhost:3007/sign-in/callback?code=codexxxx&state=stateyyyy',
       },
-      body: '<html><body>SignIn page<h1>Sign in</h1></body></html>',
+      body: '<html lang="en"><head><title>Mock SignIn page</title></head><body><h1>Sign in</h1><span class="govuk-visually-hidden" id="pageId" data-qa="sign-in"></span></body></html>',
```
nathanrussell-moj-digital, line 64: I had to update the wiremock served stub pages for hmpps-auth (ie. sign in page) so that they have the bare minimum to pass accessibility tests

integration_tests/pages/page.ts
```diff
@@ -5,14 +5,38 @@ export default abstract class Page {
...
 
-  constructor(private readonly title: string) {
+  constructor(
+    private readonly title: string,
+    private readonly options: { axeTest?: boolean } = {
+      axeTest: true,
+    },
+  ) {
     this.checkOnPage()
+    this.checkCsfrTokenForFormBasedPages()
```
nathanrussell-moj-digital, line 15: After checking we are on the correct page (based on the h1 element), we check that pages with forms have a csrf token (so we cant forget to do it 👍 ); plus if the axe option is set it runs the axe tests 👍

integration_tests/support/index.ts
```diff
@@ -1 +1,2 @@
 import './commands'
+import 'cypress-axe'
```
nathanrussell-moj-digital, line 2: This adds (imports) cypress-axe support into the integration tests

package.json
```diff
@@ -143,6 +143,7 @@
     "concurrently": "^8.2.0",
     "cookie-session": "^2.0.0",
     "cypress": "^12.14.0",
+    "cypress-axe": "^1.4.0",
```
nathanrussell-moj-digital, line 146: Add the dependency to `package.json`
(the massive change to`package-lock.json` above is normal - this always happens when a change is made to the dependencies - it basically re-writes the whole file)

mattwills-moj-digital (approved):
LGTM

PR #14, opened 2023-07-03
RR-02 Add createGoalForm

mattwills-moj-digital:


server/@types/forms/index.d.ts
```diff
@@ -0,0 +1,15 @@
+declare module 'forms' {
+  export interface CreateGoalForm {
+    prisonNumber: string
+    title: string
+    reviewDate: Date
+    steps: [
+      {
+        title: string
+        targetDate?: Date
+        sequenceNumber: number
```
mattwills-moj-digital, line 10: I guess we could make Step an object in its own right, but what would we call it? I don't think `CreateStepForm` would be right.

server/@types/forms/index.d.ts
```diff
@@ -0,0 +1,15 @@
+declare module 'forms' {
+  export interface CreateGoalForm {
+    prisonNumber: string
```
nathanrussell-moj-digital, line 3: I wonder if we can make this field readonly?
Like this: https://github.com/ministryofjustice/send-legal-mail-to-prisons/blob/main/server/%40types/forms/index.d.ts#L33

Also, if you look at the other forms in SLM; I think all the fields (except the readonly prisoner number / name, which are known at instantiation time) are nullable.

I agree this doesn't sit well with us coming from a strongly typed background, but it kinda makes sense. We're going to new one of these up before the initial page is rendered, so all of the user entered fields will be null. Its not until the submit handler (that we are yet to write) that we will populate some of the fields into the object

I guess we need to not look at it as part of the formal API spec (which we would do if it were the stateless API), and more of a transient bucket of data that we will build up over several screen interactions.

nathanrussell-moj-digital (approved):
LGTM 👍

PR #18, opened 2023-07-03
RR-02 Submit Add Step page

mattwills-moj-digital:


integration_tests/pages/goal/AddNotePage.ts
```diff
@@ -19,7 +19,7 @@ export default class AddNotePage extends Page {
     this.submitButton().click()
   }
 
-  noteField = (): PageElement => cy.get('#goal-note')
+  noteField = (): PageElement => cy.get('#notes')
```
mattwills-moj-digital, line 22: I note we're using a mix of `note` and `notes`. The latter is used in the API and I thought it makes sense to have the form field match it, but the singular is used on the screens itself.

server/@types/forms/index.d.ts
```diff
@@ -15,4 +15,8 @@ declare module 'forms' {
     'targetDate-month'?: string
     'targetDate-year'?: string
   }
+
+  export interface AddNoteForm {
+    notes?: string
```
mattwills-moj-digital, line 20: mmm...this probably ought to be `note` and we'll map it to `notes` in the API.

integration_tests/e2e/goal/createGoal.cy.ts
```diff
@@ -29,18 +30,24 @@ context('Create a goal', () => {
...
 
     addStepPage.setStepTitle('Book French course')
     addStepPage.setStepTargetDate(23, 12, 2024)
+
+    // When
+    addStepPage.submitPage()
+
+    // Then
+    const addNotePage = Page.verifyOnPage(AddNotePage)
+    addNotePage.isForPrisoner(prisonNumber)
```
nathanrussell-moj-digital, line 49: I like where this is going as an E2E test; but we are losing some of the intermediate page assertions that we had previously. IE. we've lost the assertions of `isForPrisoner` on the CreateGoal page and Add Note page

I'm not sure I want to add assertions like that to _this_ test, because otherwise it will become massive (and the test is about the E2E journey, not about whether interim pages are rendered correctly ..... Hmmmm, we'll need to give this some thought .... 🤔

nathanrussell-moj-digital (approved):
LGTM 👍

PR #19, opened 2023-07-04
RR-2 - Added validation for Create Goal

nathanrussell-moj-digital:
This PR adds validation on the submission of the first Create Goal screen; errors display; and an integration test

Screenshot 2023-07-04 at 08 54 29

### DIFF
--- a/integration_tests/e2e/goal/createGoal.cy.ts
+++ b/integration_tests/e2e/goal/createGoal.cy.ts
@@ -24,6 +24,25 @@ context('Create a goal', () => {
     page.isForPrisoner(prisonNumber)
   })
 
+  it('should not proceed to add step page given validation errors on create goal page', () => {
+    // Given
+    const prisonNumber = 'G6115VJ'
+    cy.signIn()
+    cy.visit(`/plan/${prisonNumber}/goals/create`)
+
+    const page = Page.verifyOnPage(CreateGoalPage)
+    page.setGoalTitle('Learn French')
+    page.clearGoalReviewDate()
+
+    // When
+    page.submitPage()
+
+    // Then
+    Page.verifyOnPage(CreateGoalPage)
+    page.hasErrorCount(1)
+    page.hasFieldInError('reviewDate')
+  })
+
   it('should create a valid goal', () => {
     // Given
     const prisonNumber = 'G6115VJ'
--- a/integration_tests/e2e/goal/createGoal.cy.ts
+++ b/integration_tests/e2e/goal/createGoal.cy.ts
@@ -31,16 +31,18 @@ context('Create a goal', () => {
     cy.visit(`/plan/${prisonNumber}/goals/create`)
 
     const page = Page.verifyOnPage(CreateGoalPage)
-    page.setGoalTitle('Learn French')
-    page.clearGoalReviewDate()
+    page //
+      .setGoalTitle('Learn French')
+      .clearGoalReviewDate()
 
     // When
     page.submitPage()
 
     // Then
     Page.verifyOnPage(CreateGoalPage)
-    page.hasErrorCount(1)
-    page.hasFieldInError('reviewDate')
+    page //
+      .hasErrorCount(1)
+      .hasFieldInError('reviewDate')
   })
 
   it('should create a valid goal', () => {
@@ -50,15 +52,17 @@ context('Create a goal', () => {
     cy.visit(`/plan/${prisonNumber}/goals/create`)
 
     const createGoal = Page.verifyOnPage(CreateGoalPage)
-    createGoal.setGoalTitle('Learn French')
-    createGoal.setGoalReviewDate(23, 12, 2024)
-    createGoal.submitPage()
+    createGoal //
+      .setGoalTitle('Learn French')
+      .setGoalReviewDate(23, 12, 2024)
+      .submitPage()
 
     const addStepPage = Page.verifyOnPage(AddStepPage)
     addStepPage.isForGoal('Learn French')
 
-    addStepPage.setStepTitle('Book French course')
-    addStepPage.setStepTargetDate(23, 12, 2024)
+    addStepPage //
+      .setStepTitle('Book French course')
+      .setStepTargetDate(23, 12, 2024)
 
     // When
     addStepPage.submitPage()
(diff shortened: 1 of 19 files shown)


integration_tests/e2e/goal/createGoal.cy.ts
```diff
@@ -24,22 +24,45 @@ context('Create a goal', () => {
...
   })
 
+  it('should not proceed to add step page given validation errors on create goal page', () => {
+    // Given
+    const prisonNumber = 'G6115VJ'
+    cy.signIn()
+    cy.visit(`/plan/${prisonNumber}/goals/create`)
+
+    const page = Page.verifyOnPage(CreateGoalPage)
+    page //
```
nathanrussell-moj-digital, line 34: Minor "tidying" ... seeing as the methods all return "this" we can chain them .... but unfortunately prettier feels we can and should put it all on one line (which makes it harder to read IMHO), so we have to trick prettier by adding the comment slashes !

integration_tests/pages/page.ts
```diff
@@ -40,4 +40,19 @@ export default abstract class Page {
   signOut = (): PageElement => cy.get('[data-qa=signOut]')
 
   manageDetails = (): PageElement => cy.get('[data-qa=manageDetails]')
+
+  hasNoErrors() {
```
nathanrussell-moj-digital, line 44: These methods are in the superclass `Page` (which all page classes inherit from), so we can put common methods like this in here 👍

server/filters/findErrorFilter.ts
```diff
@@ -0,0 +1,9 @@
+type Error = {
+  href?: string
+  text?: string
+}
+
+export default function findErrorFilter(array: Error[], formFieldId: string): Error {
```
nathanrussell-moj-digital, line 6: Probably easier if I walk you through how Errors work

server/views/pages/goal/create/index.njk
```diff
@@ -27,14 +36,20 @@
         <div class="govuk-form-group">
           <input type="hidden" name="_csrf" value="{{ csrfToken }}" />
 
+          {% set titleLabelHtml %}
```
nathanrussell-moj-digital, line 39: I took the opportunity to update the page content from Jo's content docs ... though I think the slack conversation about it is making me wish I hadn't have!

integration_tests/e2e/goal/createGoal.cy.ts
mattwills-moj-digital: Ok, but I'm not sure if the chaining adds any value here 🤔

nathanrussell-moj-digital: Thats a fair challenge. I can put these back to how they were if we like, but I think in subsequent tests where we might want to assert several things on the page the chaining will come into its own 👍 
Shall I put these ones back for now, on the basis that we dont really need the chaining syntax here and now?

server/routes/createGoal/createGoalFormValidator.ts
```diff
@@ -0,0 +1,12 @@
+import type { CreateGoalForm } from 'forms'
+import validateGoalReviewDate from '../../validators/goalReviewDateValidator'
+import formatErrors from '../errorFormatter'
+import validateGoalTitle from '../../validators/goalTitleValidator'
+
+export default function validateCreateGoalForm(createGoalForm: CreateGoalForm): Array<Record<string, string>> {
+  const errors: Array<Record<string, string>> = []
+
+  errors.push(...formatErrors('title', validateGoalTitle(createGoalForm.title)))
+  errors.push(...formatErrors('reviewDate-day', validateGoalReviewDate(createGoalForm.reviewDate)))
```
mattwills-moj-digital, line 10: Have you chosen `-day` because you have to select one of the fields for the errors to work at the top of the page? Interestingly however, I note that the review date fields don't have a red border around them like the goal title does.

server/validators/goalReviewDateValidator.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export default function validateGoalReviewDate(reviewDate?: Date): Array<string> {
+  const errors: Array<string> = []
+
+  if (!reviewDate) {
+    errors.push('Enter a review date for the goal')
+  } else if (Number.isNaN(reviewDate.getTime())) {
+    errors.push('Enter a review date in the correct format')
+  } else if (reviewDate < moment().toDate()) {
+    errors.push('Enter a review date in the future')
```
mattwills-moj-digital, line 11: Are you thinking of making this more generic for when we validate a step's target date?

integration_tests/e2e/goal/createGoal.cy.ts
mattwills-moj-digital: Nah, I'd leave it now you've done it 👍

mattwills-moj-digital (approved):
LGTM. Just a couple of questions.

integration_tests/e2e/goal/createGoal.cy.ts
nathanrussell-moj-digital: Cool, will do 👍

server/routes/createGoal/createGoalFormValidator.ts
nathanrussell-moj-digital: That is a good question and a good observation ...
Honestly .... I copied it from what we did in SLM (surprise!!!! 😁 ), but there was some potentially nasty hacky stuff that I deliberately didnt copy that I believe was about styling the whole field set with a red border 

I didn't like the apparent hackyness of it, certainly considering this is the GDS templates, components and styling ... so should just work ...

Let me look at it again and see what I can do / remember / hack together 😁

server/validators/goalReviewDateValidator.ts
nathanrussell-moj-digital: Not quite. I've made it generic (more specifically agnostic) to the form where a goal review date will need to be validated (ie. create goal, and later on, edit goal)

I'm envisaging we will have a very similar validator function, but for step target date, and it will have different error message wording. (or, perhaps we wont have it at all, and it will be a "step target date range validator" 🙄 )

mattwills-moj-digital: I was thinking it would be nice if the message came from somewhere else. The validator would know how to get the message based on the field name. For example, I'm thinking of what we used to do with `ResourceBundles`. It's been a long time since I used one of those! 😄

PR #21, opened 2023-07-04
Try using user token rather than systemToken when calling prisoner-search

nathanrussell-moj-digital:
This PR attempts to call prisoner-search using the user token rather than the system token.
If this works then we can consider which is the right token that we _should_ be using (it might be that the system token is the right token, but we need to give it some additional roles. Or we might decide the user token is the right token)

### DIFF
--- a/server/routes/createGoal/createGoalController.ts
+++ b/server/routes/createGoal/createGoalController.ts
@@ -1,7 +1,7 @@
 import type { RequestHandler } from 'express'
 import type { Prisoner } from 'prisonRegisterApiClient'
 import type { PrisonerSummary } from 'viewModels'
-// import createError from 'http-errors'
+import createError from 'http-errors'
 import PrisonerSearchService from '../../services/prisonerSearchService'
 import CreateGoalView from './createGoalView'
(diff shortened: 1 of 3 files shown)


server/routes/createGoal/createGoalController.ts
```diff
@@ -13,13 +13,11 @@ export default class CreateGoalController {
   getCreateGoalView: RequestHandler = async (req, res, next): Promise<void> => {
     const { prisonNumber } = req.params
     let prisoner: Prisoner
-    // eslint-disable-next-line no-useless-catch
     try {
-      prisoner = await this.prisonerSearchService.getPrisonerByPrisonNumber(prisonNumber)
+      prisoner = await this.prisonerSearchService.getPrisonerByPrisonNumber(prisonNumber, req.user.token)
```
nathanrussell-moj-digital, line 17: Put all this back from the attempt in my last PR (that didnt work!!!)

nathanrussell-moj-digital: (and pass in the user token from the request, on line 17)

PR #25, opened 2023-07-05
RR-2 - Basic screen routing to support adding other steps

nathanrussell-moj-digital:
This PR adds the basic screen flow and processing to support "Add another step"
All this PR does is address the screen flow - it does not attempt to do anything with the submitted step

### DIFF
--- a/integration_tests/e2e/goal/addStep.cy.ts
+++ b/integration_tests/e2e/goal/addStep.cy.ts
@@ -37,6 +37,37 @@ context('Add a step', () => {
       .hasFieldInError('title')
   })
 
+  it('should not proceed to add step page given user chooses to add another step', () => {
+    // Given
+    const prisonNumber = 'G6115VJ'
+    cy.signIn()
+    cy.visit(`/plan/${prisonNumber}/goals/create`)
+
+    const createGoal = Page.verifyOnPage(CreateGoalPage)
+    createGoal //
+      .setGoalTitle('Learn French')
+      .setGoalReviewDate(23, 12, 2024)
+      .submitPage()
+
+    let addStepPage = Page.verifyOnPage(AddStepPage)
+    addStepPage //
+      .isForGoal('Learn French')
+      .isStepNumber(1)
+
+    addStepPage //
+      .setStepTitle('Book French course')
+      .setStepTargetDate(23, 12, 2024)
+
+    // When
+    addStepPage.addAnotherStep()
+
+    // Then
+    addStepPage = Page.verifyOnPage(AddStepPage)
+    addStepPage //
+      .isForGoal('Learn French')
+      .isStepNumber(2)
+  })
+
   it.skip('should move to add note page', () => {
     // Given
     const prisonNumber = 'G6115VJ'
--- a/integration_tests/pages/goal/AddStepPage.ts
+++ b/integration_tests/pages/goal/AddStepPage.ts
@@ -15,6 +15,11 @@ export default class AddStepPage extends Page {
     return this
   }
 
+  isStepNumber(expected: number) {
+    this.stepNumberLabel().should('contain.text', `Step ${expected}`)
+    return this
+  }
+
   setStepTitle(title: string) {
     this.titleField().clear().type(title)
     return this
@@ -36,6 +41,10 @@ export default class AddStepPage extends Page {
     this.submitButton().click()
   }
 
+  addAnotherStep() {
+    this.addAnotherStepButton().click()
+  }
+
   goalTitle = (): PageElement => cy.get('h1')
 
   titleField = (): PageElement => cy.get('#title')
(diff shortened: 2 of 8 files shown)


integration_tests/e2e/goal/addStep.cy.ts
```diff
@@ -37,6 +37,37 @@ context('Add a step', () => {
...
+      .setStepTargetDate(23, 12, 2024)
+
+    // When
+    addStepPage.addAnotherStep()
+
+    // Then
+    addStepPage = Page.verifyOnPage(AddStepPage)
+    addStepPage //
+      .isForGoal('Learn French')
+      .isStepNumber(2)
```
mattwills-moj-digital, line 68: Nice 👍

integration_tests/pages/goal/AddStepPage.ts
```diff
@@ -46,9 +55,11 @@ export default class AddStepPage extends Page {
...
   targetDateYearField = (): PageElement => cy.get('#targetDate-year')
 
-  addStepButton = (): PageElement => cy.get('#add-step-button')
+  addAnotherStepButton = (): PageElement => cy.get('#add-another-step-button')
 
   submitButton = (): PageElement => cy.get('#submit-button')
 
   prisonNumberLabel = (): PageElement => cy.get('[data-qa=prison-number]')
+
+  stepNumberLabel = (): PageElement => cy.get('label[for=title]')
```
mattwills-moj-digital, line 64: You can explain this one to me tomorrow 🤯

server/@types/forms/index.d.ts
```diff
@@ -9,11 +9,13 @@ declare module 'forms' {
   }
 
   export interface AddStepForm {
+    stepNumber: number
     title?: string
     targetDate?: Date
     'targetDate-day'?: string
     'targetDate-month'?: string
     'targetDate-year'?: string
+    action?: string
```
mattwills-moj-digital, line 18: I'm not sure about the name `action` (it's a bit vague), but I can't think of a better one right now! 🤔

server/routes/createGoal/createGoalController.test.ts
```diff
@@ -0,0 +1,117 @@
...
+import { SessionData } from 'express-session'
+import CreateGoalController from './createGoalController'
+import { PrisonerSearchService } from '../../services'
+import validateCreateGoalForm from './createGoalFormValidator'
+import validateAddStepForm from './addStepFormValidator'
+
+jest.mock('./addStepFormValidator')
+jest.mock('./createGoalFormValidator')
+
+describe('createGoalController', () => {
```
mattwills-moj-digital, line 11: I spotted we didn't have a test for the controller earlier and wondered if we were just going to rely on the int-tests. However, you've added more logic in it now, so it makes sense 👍

server/routes/createGoal/createGoalController.ts
```diff
@@ -66,13 +66,24 @@ export default class CreateGoalController {
...
       req.flash('errors', errors)
       return res.redirect(`/plan/${prisonNumber}/goals/add-step`)
     }
 
+    // Redirect to the desired page based on the form action
+    if (addStepForm.action === 'add-another-step') {
+      // The next PR will work out where to store steps as each is submitted
+
+      // Initialize a new AddStepForm with the next step number
+      const nextStepNumber = Number(addStepForm.stepNumber) + 1
```
mattwills-moj-digital, line 82: Why not just `addStepForm.stepNumber + 1`? 🤔

mattwills-moj-digital (approved):
LGTM

integration_tests/pages/goal/AddStepPage.ts
nathanrussell-moj-digital: lol - it's a css selector (that's what all the args in the `cy.get` method calls are - they are all css selectors to select 1 or more elements on the page)
Specifically its the selector to get a `label` element whose attribute `for` has the value `title`. IE. it selects this element in the markup:
```
<label class="govuk-label govuk-label--m" for="title">
```
Screenshot 2023-07-06 at 08 41 13

server/routes/createGoal/createGoalController.ts
nathanrussell-moj-digital: Excellent question 😁 - see my screenshot above which shows "Step 1111"
Even though our form type (in `@types`) defines the field as a `number`, at the point of it being rendered in the markup in the various `input` fields etc, and then subsequently submitted back in the HTTP request, all the fields become `string`s ... which is annoying !
It's because `request.params` (which is the map of form fields that are submitted on the POST request) is typed as a `Map<String, String>` (in java speak) 
When the controller constructs the form object from the request params (ie. `req.session.addStepForm = { ...req.body, targetDate }`) it annoying doesnt coerce the data types into the target object. It changes the types of the target object !!!! (because at runtime its all javascript, and js doesnt have type safety in the same way that typescript does. All the neat typescript stuff with types etc is all compile time and developer convenience. At runtime its all plain old javascript!)

So .... to answer your question ... at runtime at this line 82, `addStepForm.stepNumber` is a string, so `addStepForm.stepNumber + 1` concatenates it as a string (hence `Step 1111` in my screenshot!)

server/@types/forms/index.d.ts
nathanrussell-moj-digital: I did consider `formAction`, but decided against it because a) its in a form type anyway so felt that was already implied; and b) "form action" means something subtly different in context of a HTML form (it has an `action` attribute which tells the browser where to post the data to)

server/routes/createGoal/createGoalController.ts
mattwills-moj-digital: I miss Kotlin! 😄

PR #26, opened 2023-07-05
RR-02 Send new goal to API

mattwills-moj-digital:


server/routes/createGoal/createGoalFormToCreateGoalDtoMapper.test.ts
```diff
@@ -0,0 +1,47 @@
+import type { AddStepDto, CreateGoalDto } from 'dto'
+import type { AddNoteForm, AddStepForm, CreateGoalForm } from 'forms'
+import moment from 'moment'
+import CreateGoalFormToCreateGoalDtoMapper from './createGoalFormToCreateGoalDtoMapper'
+
+describe('createGoalFormToCreateGoalDtoMapper', () => {
+  it.skip('should map to CreateGoalDto given valid form data', () => {
+    // Given
+    const createGoalDtoMapper = new CreateGoalFormToCreateGoalDtoMapper()
```
mattwills-moj-digital, line 9: I'll create a const for this before the test, like with `educationAndWorkPlanService` in educationAndWorkPlanService.test below.

server/data/educationAndWorkPlanApiMapper.test.ts
```diff
@@ -0,0 +1,43 @@
...
+      title: 'Book Spanish course',
+      targetDate: moment('2123-01-31', 'YYYY-MM-DD').toDate(),
+    } as AddStepDto
+    const createGoalDto = {
+      prisonNumber,
+      title: 'Learn Spanish',
+      reviewDate: moment('2123-06-31', 'YYYY-MM-DD').toDate(),
+      steps: [addStepDto],
+      note: 'Prisoner is not good at listening',
+    } as CreateGoalDto
```
mattwills-moj-digital, line 21: A test data builder would be nice 🤔

server/data/educationAndWorkPlanApiMapper.test.ts
```diff
@@ -0,0 +1,43 @@
+import type { AddStepDto, CreateGoalDto } from 'dto'
+import type { CreateGoalRequest, CreateStepRequest } from 'educationAndWorkPlanApiClient'
+import moment from 'moment'
+import EducationAndWorkPlanApiMapper from './educationAndWorkPlanApiMapper'
+
+describe('educationAndWorkPlanApiMapper', () => {
+  it.skip('should map to CreateGoalDto given valid form data', () => {
```
mattwills-moj-digital, line 7: I need to figure out why this isn't working...

server/@types/dto/index.d.ts
```diff
@@ -0,0 +1,14 @@
+declare module 'dto' {
+  export interface CreateGoalDto {
+    prisonNumber: string
+    title: string
+    reviewDate: Date
+    steps: [AddStepDto]
```
nathanrussell-moj-digital, line 6: I don't know the answer to this one ... but is that the right syntax for declaring the array vs. instantiating the array?
I've seen in other places (in this codebase) where we use `Array<type>` , so I wonder if your line 6 should be:
```
    steps: Array<AddStepDto>
```

I could be wrong, but to me it looks like what you have is the instantiation of an array with 1 element which is the `AddStepDto` type (rather than an array of `AddStepDto` instances)

Like I said, I'm no expert here,  but thats what it looks like to me?

server/@types/forms/index.d.ts
```diff
@@ -1,6 +1,5 @@
 declare module 'forms' {
   export interface CreateGoalForm {
-    readonly prisonNumber: string
```
nathanrussell-moj-digital, line 3: Was this causing you some grief somewhere?

TBH I was kinda thinking we might use that in some near-future dev for this story .... IE. I'm thinking we need to write something that says "when you arrive on the AddStep page (and the AddNote page), check you have a CreateGoalForm in the session AND its for the same prison number as in the URL address)

My thinking is we need to make sure you dont just land on the AddStep or AddNote page without having gone through the previous pages (ie. people bookmarking / typing urls etc); and also we should check that if you've started the journey for prisoner X you complete it for prisoner X

server/data/educationAndWorkPlanApiMapper.test.ts
```diff
@@ -0,0 +1,43 @@
...
+      reviewDate: moment('2123-06-31', 'YYYY-MM-DD').toDate(),
+      steps: [expectedAddStepRequest],
+      note: createGoalDto.note,
+    } as CreateGoalRequest
+
+    // When
+    const createGoalRequest = educationAndWorkPlanApiMapper.createGoalDtoToCreateGoalRequest(createGoalDto)
+
+    // Then
+    expect(createGoalRequest).toEqual(expectedCreateGoalRequest)
```
nathanrussell-moj-digital, line 41: I cant be sure, but it might be because the objects you are comparing are "deep" / ie. not simple flat objects
I think theres a `toDeepEquals` or similarly named method you might be able to try?

server/data/educationAndWorkPlanApiMapper.ts
```diff
@@ -0,0 +1,22 @@
+import type { CreateGoalRequest, CreateStepRequest } from 'educationAndWorkPlanApiClient'
+import type { CreateGoalDto } from 'dto'
+
+export default class EducationAndWorkPlanApiMapper {
```
nathanrussell-moj-digital, line 4: I guess we need to think about where our mappers should live 🤔 
I guess this is OK in `data` because its creating an object that will be used by a client class (that also lives in `data` - yuck! 🤣 )
But I wonder if perhaps we should put it in `data/mappers` or similar ?

server/data/educationAndWorkPlanApiMapper.ts
```diff
@@ -0,0 +1,22 @@
...
+  createGoalDtoToCreateGoalRequest(createGoalDto: CreateGoalDto): CreateGoalRequest {
```
nathanrussell-moj-digital, line 5: Rather than making this a class, I wonder if you'd be better off doing away with the class, and simply exporting the function, and naming the source file the same
Which makes me wonder what the function name should be? 🤔 

How about:
`educationAndWorkPlanApiMappers.ts` (plural, because perhaps it exports several functions that can map stuff relating to this API ?)
And then its contents might be something like:
```
const mapCreateGoalDtoToCreateGoalRequest(createGoalDto: CreateGoalDto): CreateGoalRequest {
...
...
}

export default {
  mapCreateGoalDtoToCreateGoalRequest,
}
```
(and then you could have several functions in the module (source file), all being exported as a named function)

Which then makes the Client class below a little easier in respect of using it (and mocking it in its tests)

server/data/educationAndWorkPlanClient.ts
```diff
@@ -0,0 +1,24 @@
...
+  private static restClient(token: string): RestClient {
+    return new RestClient('Education and Work Plan API Client', config.apis.educationAndWorkPlan, token)
+  }
+
+  private static educationAndWorkPlanApiMapper(): EducationAndWorkPlanApiMapper {
+    return new EducationAndWorkPlanApiMapper()
+  }
+
+  async createGoal(createGoalDto: CreateGoalDto, token: string): Promise<void> {
+    const createGoalRequest =
```
nathanrussell-moj-digital, line 16: If you did as I suggested above re: the mapper, you can tidy this up. You can remove the static function altogether; and line 16 becomes:
```
const createGoalRequest = mapCreateGoalDtoToCreateGoalRequest(createGoalDto)
```

server/routes/createGoal/createGoalController.ts
```diff
@@ -2,16 +2,26 @@ import type { RequestHandler } from 'express'
...
+import logger from '../../../logger'
 
 export default class CreateGoalController {
-  constructor(private readonly prisonerSearchService: PrisonerSearchService) {}
+  constructor(
+    private readonly prisonerSearchService: PrisonerSearchService,
+    private readonly educationAndWorkPlanService: EducationAndWorkPlanService,
+  ) {}
+
+  private static createDtoMapper(): CreateGoalFormToCreateGoalDtoMapper {
```
nathanrussell-moj-digital, line 22: Same comment as above re: defining your mapper as a simple exported function rather than a class, then you dont need this static etc

nathanrussell-moj-digital (commented):
I think there's a lot of good stuff in here, and the comments I've left are really around some of the more 'unusual' features of typescript that as java/kotlin devs are not as immediately obvious to us.
See what you think about my comments - you might have other thoughts or views on them?
We can chat it through tomorrow if you like

server/@types/dto/index.d.ts
mattwills-moj-digital: I think you might be right 👍

server/@types/forms/index.d.ts
mattwills-moj-digital: It wasn't causing a problem, just some confusion as it was `prisonNumber` was effectively duplicated in the controller. I can revert it if you like, or we can add it when we need it.

server/data/educationAndWorkPlanApiMapper.ts
mattwills-moj-digital: Nice idea. I did wonder about introducing different mapper files and where to put them :)

server/data/educationAndWorkPlanApiMapper.test.ts
mattwills-moj-digital: Unfortunately that doesn't exist and I've tried a few different methods (like `toStrictEqual`), but no joy so far...

server/@types/dto/index.d.ts
mattwills-moj-digital: Updated 👍

server/data/educationAndWorkPlanApiMapper.test.ts
mattwills-moj-digital: I'm using `JSON.stringify` for now and have enabled the test.

server/data/educationAndWorkPlanClient.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    nock.cleanAll()
+  })
+
+  describe('createGoal', () => {
+    it('should create goal', async () => {
+      // Given
+      const prisonNumber = 'A1234BC'
+      const systemToken = 'a-system-token'
+
+      const addStepDto = {
```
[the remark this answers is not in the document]
mattwills-moj-digital: On the basis we put the tests alongside the main classes, I guess we could put them in the same folder as the classes they're relevant to?

server/data/educationAndWorkPlanApiMapper.test.ts
nathanrussell-moj-digital: I've found the problem; and you are absolutely gonna kick yourself!!!! (though I blame this on one of Javascript's weirder parts; of which there are many!)
I'll talk you through it in a huddle this morning 👍

server/@types/forms/index.d.ts
nathanrussell-moj-digital: If you could revert it that would be good. I think I might start on the code that will use it today ...

server/data/educationAndWorkPlanApiMapper.test.ts
```diff
@@ -29,7 +29,7 @@ describe('educationAndWorkPlanApiMapper', () => {
       prisonNumber,
       title: 'Learn Spanish',
       category: 'WORK',
-      reviewDate: moment('2123-06-31', 'YYYY-MM-DD').toDate(),
+      reviewDate: moment('2123-06-30', 'YYYY-MM-DD').toDate(),
```
mattwills-moj-digital, line 32: FFS! 😂

server/@types/dto/index.d.ts
```diff
@@ -3,7 +3,7 @@ declare module 'dto' {
     prisonNumber: string
     title: string
     reviewDate: Date
-    steps: Array[AddStepDto]
+    steps: Array<AddStepDto>
```
mattwills-moj-digital, line 6: Thanks! 🤦‍♂️

server/data/educationAndWorkPlanClient.test.ts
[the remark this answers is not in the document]
mattwills-moj-digital: As discussed, I've created a separate `testsupport` folder.

nathanrussell-moj-digital (approved):
LGTM 👍 
Thanks for considering and working through my earlier comments - I know some of it was a little frustrating!

Great stuff! Lets get it merged! 👍

PR #27, opened 2023-07-06
RR-2 - Implement form session handling to prevent page navigation out of sequence

nathanrussell-moj-digital:
This PR implements form session handling code to prevent the user navigating to pages in the Create A Goal journey out of sequence.

See the integration tests for specific examples of the user interactions that this code protects against

Currently the redirect is to the first page in our Create Goal journey. Once we have more screens, perhaps a better screen to redirect to would be the Overview page for the prisoner identified in the URL . Perhaps we make that change when we do RR-69 🤔

### DIFF
--- a/integration_tests/e2e/goal/addNote.cy.ts
+++ b/integration_tests/e2e/goal/addNote.cy.ts
@@ -56,6 +56,28 @@ context('Add a note', () => {
     createGoalPage.isForPrisoner(someOtherPrisonNumber)
   })
 
+  it.skip('should not be able to arrive on add step page, then change the URL to go straight to add-note without submitting add-step', () => {
+    // Given
+    const prisonNumber = 'G6115VJ'
+    cy.signIn()
+    cy.visit(`/plan/${prisonNumber}/goals/create`)
+
+    let createGoalPage = Page.verifyOnPage(CreateGoalPage)
+    createGoalPage //
+      .setGoalTitle('Learn French')
+      .setGoalReviewDate(23, 12, 2024)
+      .submitPage()
+
+    Page.verifyOnPage(AddStepPage)
+
+    // When
+    cy.visit(`/plan/${prisonNumber}/goals/add-note`)
+
+    // Then
+    createGoalPage = Page.verifyOnPage(CreateGoalPage)
+    createGoalPage.isForPrisoner(prisonNumber)
+  })
+
   it.skip('should move to review goals page', () => {
     // Given
     const prisonNumber = 'G6115VJ'
--- a/integration_tests/e2e/goal/addNote.cy.ts
+++ b/integration_tests/e2e/goal/addNote.cy.ts
@@ -8,7 +8,52 @@ context('Add a note', () => {
     cy.task('reset')
     cy.task('stubSignInAsUserWithEditAuthority')
     cy.task('stubAuthUser')
-    cy.task('getPrisonerById')
+    cy.task('getPrisonerById', 'G6115VJ')
+    cy.task('getPrisonerById', 'H4115SD')
+  })
+
+  it('should not be able to navigate directly to add note given Create Goal and Add Step has not been submitted', () => {
+    // Given
+    const prisonNumber = 'G6115VJ'
+    cy.signIn()
+
+    // When
+    cy.visit(`/plan/${prisonNumber}/goals/add-note`)
+
+    // Then
+    const createGoalPage = Page.verifyOnPage(CreateGoalPage)
+    createGoalPage.isForPrisoner(prisonNumber)
+  })
+
+  it('should not be able to arrive on add note page, then change the prison number in the URL', () => {
+    // Given
+    const prisonNumber = 'G6115VJ'
+    cy.signIn()
+    cy.visit(`/plan/${prisonNumber}/goals/create`)
+
+    let createGoalPage = Page.verifyOnPage(CreateGoalPage)
+    createGoalPage //
+      .setGoalTitle('Learn French')
+      .setGoalReviewDate(23, 12, 2024)
+      .submitPage()
+
+    const addStepPage = Page.verifyOnPage(AddStepPage)
+    addStepPage //
+      .setStepTitle('Book French course')
+      .setStepTargetDate(23, 12, 2024)
+      .submitPage()
+
+    const addNotePage = Page.verifyOnPage(AddNotePage)
+    addNotePage.isForPrisoner(prisonNumber)
+
+    const someOtherPrisonNumber = 'H4115SD'
+
+    // When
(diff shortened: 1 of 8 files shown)


integration_tests/e2e/goal/addNote.cy.ts
```diff
@@ -8,7 +8,52 @@ context('Add a note', () => {
     cy.task('reset')
     cy.task('stubSignInAsUserWithEditAuthority')
     cy.task('stubAuthUser')
-    cy.task('getPrisonerById')
+    cy.task('getPrisonerById', 'G6115VJ')
```
nathanrussell-moj-digital, line 11: We need to mock 2 calls to getPrisonerByID

The previous mock was for G6115VJ (its the default in the wiremock stub)
But we now need to mock calls for 2 different prisoners. The prisoner number can be passed as an argument like this, which overrides the default

integration_tests/e2e/goal/addNote.cy.ts
```diff
@@ -8,7 +8,52 @@ context('Add a note', () => {
...
+
+    const someOtherPrisonNumber = 'H4115SD'
+
+    // When
+    cy.visit(`/plan/${someOtherPrisonNumber}/goals/add-note`)
+
+    // Then
+    createGoalPage = Page.verifyOnPage(CreateGoalPage)
+    createGoalPage.isForPrisoner(someOtherPrisonNumber)
   })
```
nathanrussell-moj-digital, line 58: There is a scenario that this code does not yet address - that being where the user has arrived on page 1 (Describe a Goal); they have successfully submitted it and arrived on page 2 (Add a Step). But instead of submitting the Add A Step page, then change the URL in the address bar to `/add-note` (but for the same prisoner)
In this scenario we would arrive on the Add Note page. The CreateGoalForm object exists in the session for the correct prisoner 👍 , but there is no populated AddStepForm object in the session because the user never actually submitted page 2 (Add a Step)

I decided against protecting against this until we have our collection of AddStepForms in the session (rather than just one). Once that work is complete we can come back here and implement it

server/@types/forms/index.d.ts
```diff
@@ -1,6 +1,6 @@
 declare module 'forms' {
   export interface CreateGoalForm {
-    readonly prisonNumber: string
+    prisonNumber: string
```
nathanrussell-moj-digital, line 3: I've changed this from a readonly property; and have added it as a hidden field on the HTML form (see the change to the nunjucks view below)
It makes the controller much simpler in respect of populating the form in the session

server/routes/createGoal/createGoalController.ts
```diff
@@ -20,10 +20,15 @@ export default class CreateGoalController {
 
   getCreateGoalView: RequestHandler = async (req, res, next): Promise<void> => {
     const { prisonNumber } = req.params
+    if (!req.session.createGoalForm) {
```
nathanrussell-moj-digital, line 23: I've moved the creation of the form and putting it in the session up towards the top of the GET method. It makes more sense for it to be here 👍

server/routes/createGoal/createGoalController.ts
```diff
@@ -37,17 +42,15 @@ export default class CreateGoalController {
...
+    const view = new CreateGoalView(prisonerSummary, req.session.createGoalForm, req.flash('errors'))
     res.render('pages/goal/create/index', { ...view.renderArgs })
   }
 
   submitCreateGoalForm: RequestHandler = async (req, res, next): Promise<void> => {
     const { prisonNumber } = req.params
 
     const reviewDate = parseDate(req, 'reviewDate')
-    req.session.createGoalForm = { ...req.body, prisonNumber, reviewDate }
+    req.session.createGoalForm = { ...req.body, reviewDate }
```
nathanrussell-moj-digital, line 53: Now when the CreateGoalForm is submitted we can simply replace the session object, and it will correctly populate the prison number (because it was submitted in the HTML form submission via the hidden field)

server/views/pages/goal/create/index.njk
```diff
@@ -35,6 +35,7 @@
       <form class="form" method="post" novalidate="">
         <div class="govuk-form-group">
           <input type="hidden" name="_csrf" value="{{ csrfToken }}" />
+          <input type="hidden" name="prisonNumber" value="{{ form.prisonNumber }}" />
```
mattwills-moj-digital, line 38: 👍

server/routes/createGoal/routerRequestHandlers.ts
```diff
@@ -0,0 +1,30 @@
...
+ */
+const checkCreateGoalFormExistsInSession = async (req: Request, res: Response, next: NextFunction) => {
+  if (!req.session.createGoalForm) {
+    logger.warn(
+      `No CreateGoalForm object in session - user attempting to navigate to path ${req.path} out of sequence. Redirecting to start of Create Goal journey.`,
+    )
+    res.redirect(`/plan/${req.params.prisonNumber}/goals/create`)
+  } else if (req.session.createGoalForm.prisonNumber !== req.params.prisonNumber) {
+    logger.warn(
+      `CreateGoalForm object in session references a different prisoner. Redirecting to start of Create Goal journey.`,
```
mattwills-moj-digital, line 21: Minor - this is ok, but perhaps it could be expressed in less technical "business" terms (i.e. so that someone who doesn't know what the `CreateGoalForm` is might be able to understand the issue just from the logs)?

Having said that, I'm finding it difficult to choose the right words! 😆

server/routes/createGoal/index.ts
```diff
@@ -1,26 +1,42 @@
...
-  get('/plan/:prisonNumber/goals/add-note', createGoalController.getAddNoteView)
-  post('/plan/:prisonNumber/goals/add-note', createGoalController.submitAddNoteForm)
+  router.get('/plan/:prisonNumber/goals/add-note', [
+    checkCreateGoalFormExistsInSession,
+    createGoalController.getAddNoteView,
+  ])
+  router.post('/plan/:prisonNumber/goals/add-note', [
+    checkCreateGoalFormExistsInSession,
+    createGoalController.submitAddNoteForm,
+  ])
```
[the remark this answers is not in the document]
mattwills-moj-digital: Sounds like a `Resolver` to me! 😂

mattwills-moj-digital (approved):
GS!

server/routes/createGoal/routerRequestHandlers.ts
nathanrussell-moj-digital: Thats a fair shout.
I think we'll be adding a few more of these type of handler functions (fairly soon we will do the one that I hinted at in this PR where we need to check you have at least 1 AddStepForm when you arrive on Add Note)
So we can consider how we word these messages etc then I guess

PR #28, opened 2023-07-06
RR-02 Submit multiple steps

mattwills-moj-digital:


server/routes/createGoal/createGoalController.ts
```diff
@@ -82,6 +83,11 @@ export default class CreateGoalController {
       return res.redirect(`/plan/${prisonNumber}/goals/add-step`)
     }
 
+    if (!req.session.addStepForms) {
+      req.session.addStepForms = new Array<AddStepForm>()
```
mattwills-moj-digital, line 87: I'm sure there might be a more succinct way of doing this?!

server/routes/createGoal/createGoalController.test.ts
```diff
@@ -57,6 +56,7 @@ describe('createGoalController', () => {
         'targetDate-year': '2024',
         action: 'submit-form',
       }
+      req.session.addStepForms = new Array<AddStepForm>()
```
nathanrussell-moj-digital, line 59: When you new up a new array here and on L86 and L107; I think you can use the shorter syntax:
```
req.session.addStepForms = []
```

server/routes/createGoal/createGoalController.ts
```diff
@@ -65,6 +66,7 @@ export default class CreateGoalController {
     const { prisonerSummary } = req.session
     const { createGoalForm } = req.session
     const addStepForm = req.session.addStepForm || { stepNumber: 1 }
+    req.session.addStepForms = req.session.addStepForms || new Array<AddStepForm>()
```
nathanrussell-moj-digital, line 69:
```
req.session.addStepForms = req.session.addStepForms || []
```
Would that work?

mattwills-moj-digital: Yep. I understand more about how type safety works (or doesn't work) in TS.

server/routes/createGoal/createGoalController.ts
```diff
@@ -75,6 +77,8 @@ export default class CreateGoalController {
     const targetDate = parseDate(req, 'targetDate')
     req.session.addStepForm = { ...req.body, targetDate }
     const { addStepForm } = req.session
+    const { addStepForms } = req.session
+    addStepForms.push(req.session.addStepForm)
```
nathanrussell-moj-digital, line 81: I think we only want to push to the array after the form has been validated
As you have it I _think_ you could submit an invalid form and it would get added to the array on L81, before we validate it
The validation fails, we re-render the page, and the user submits again
The new form will be added to the array, so now there are 2 elements in the array, even though its the same StepForm

I think you'd want to move L81 to L88

What do you think?

server/routes/createGoal/createGoalController.ts
```diff
@@ -107,10 +111,10 @@ export default class CreateGoalController {
   submitAddNoteForm: RequestHandler = async (req, res, next): Promise<void> => {
     const { prisonNumber } = req.params
     const { createGoalForm } = req.session
-    const { addStepForm } = req.session
+    const { addStepForms } = req.session
     req.session.addNoteForm = { ...req.body }
 
-    const createGoalDto = toCreateGoalDto(createGoalForm, addStepForm, req.body)
+    const createGoalDto = toCreateGoalDto(createGoalForm, addStepForms, req.body)
     await this.educationAndWorkPlanService.createGoal(createGoalDto, req.user.token)
```
nathanrussell-moj-digital, line 119: I think we should nuke the forms in the session here as well - ie. the CreateGoalForm, the AddStepForm, the collection of AddStepForms; and the AddNoteForm

If we dont it means you can go back to create a new goal , perhaps for a new prisoner, and all the goal data from prisoner 1 will still be there! 🤣

nathanrussell-moj-digital (approved):
LGTM

server/routes/createGoal/createGoalController.test.ts
```diff
@@ -57,6 +58,7 @@ describe('createGoalController', () => {
         'targetDate-year': '2024',
         action: 'submit-form',
       }
+      req.session.addStepForms = []
 
       mockedValidateAddStepForm.mockReturnValue([])
```
nathanrussell-moj-digital, line 64: (The stupid diff wont let me add a comment where I want to!), but is it worth adding an assertion in the `// Then` block about the `addStepForms` array in the session?
Presumably we expect the `addStepForms` array to have had a new step added to it?
Something like:

```
// Then
...
...
expect(req.session.addStepForms).toEqual(....)
```

server/routes/createGoal/createGoalController.test.ts
```diff
@@ -83,6 +85,7 @@ describe('createGoalController', () => {
         'targetDate-year': '2024',
         action: 'add-another-step',
       }
+      req.session.addStepForms = []
```
nathanrussell-moj-digital, line 88: As above re: the expected state/content of the `addSessionsForms` array in session

server/routes/createGoal/createGoalController.test.ts
```diff
@@ -103,6 +106,7 @@ describe('createGoalController', () => {
       // Given
       req.params.prisonNumber = 'A1234GC'
       req.body = {}
+      req.session.addStepForms = []
```
nathanrussell-moj-digital, line 109: In this test, in the `// Then` block, we're going to want to assert the `addStepForms` array is still empty?

server/routes/createGoal/createGoalController.test.ts
```diff
@@ -121,5 +125,71 @@ describe('createGoalController', () => {
...
+
+      // When
+      await controller.submitAddStepForm(
+        req as undefined as Request,
+        res as undefined as Response,
+        next as undefined as NextFunction,
+      )
+
+      // Then
+      expect(req.session.addStepForms).toHaveLength(2)
```
nathanrussell-moj-digital, line 146: Nice 👍

server/routes/createGoal/createGoalController.test.ts
```diff
@@ -121,5 +125,71 @@ describe('createGoalController', () => {
...
+        req as undefined as Request,
+        res as undefined as Response,
+        next as undefined as NextFunction,
+      )
+
+      // Then
+      expect(req.session.addStepForms).toHaveLength(1)
+      expect(req.session.addStepForms[0].title).toEqual('Find a Spanish course')
+      expect(res.redirect).toHaveBeenCalledWith('/plan/A1234GC/goals/add-note')
+    })
```
nathanrussell-moj-digital, line 193: A nice collection of tests; 👏 👍

server/routes/createGoal/createGoalController.ts
```diff
@@ -75,17 +76,26 @@ export default class CreateGoalController {
     const targetDate = parseDate(req, 'targetDate')
     req.session.addStepForm = { ...req.body, targetDate }
     const { addStepForm } = req.session
+    const { addStepForms } = req.session
```
nathanrussell-moj-digital, line 79: Minor, and I'm not going to lose any sleep over this as its more weird TS syntax !!!, but because the 2 lines above are destructuring from the same variable, I think you can combine them into 1 statement, like this:
```
const { addStepForm, addStepForms } = req.session
```

server/routes/createGoal/createGoalController.ts
```diff
@@ -75,17 +76,26 @@ export default class CreateGoalController {
...
       return res.redirect(`/plan/${prisonNumber}/goals/add-step`)
     }
 
+    // check to see if this step has already been added (e.g. after the user clicks the back button)
+    const existingAddStepForm = addStepForms.find(step => step.stepNumber === addStepForm.stepNumber)
+    if (!existingAddStepForm) {
+      addStepForms.push(req.session.addStepForm)
+    } else {
+      // update it in case the user has clicked back and changed it
+      existingAddStepForm.title = addStepForm.title
```
nathanrussell-moj-digital, line 93: Based on your unit tests I can see this works, though it surprises me slightly because theres no obvious "update the original object / array element"

I guess this tells us something about how TS/JS works .... in that it looks like variables are passed by reference rather than by value ..... perhaps something to watch out for in the future .... 👀 

(Everyday's a school day! 🤣 )

nathanrussell-moj-digital (dismissed):
This looks great; and really good work getting the step update code working to handle the browser back button!
Nice 👍

nathanrussell-moj-digital (approved):

PR #31, opened 2023-07-10
RR-69 - Add the initial basic route, controller and view

nathanrussell-moj-digital:
Very much a basic WIP, but this PR defines the route, controller and basic view for the overview page.

It is based on / modelled on what CIAG Inductions are doing in respect of the tabbed interface and how the views work in this respect (each tab content is a partial)

There is still work to do such as retrieving the prisoner summary to put their name etc on the view; and also adding the Add A Goal button
But as a bare bones skeleton this is a good starter for 10 🤞 

Screenshot 2023-07-10 at 16 45 44

### DIFF
--- a/server/views/pages/overview/partials/navigation.njk
+++ b/server/views/pages/overview/partials/navigation.njk
@@ -1,0 +1,53 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
+    <nav class="moj-sub-navigation" aria-label="Sub navigation">
+      <ul class="moj-sub-navigation__list">
+        <li class="moj-sub-navigation__item">
+          <a class="moj-sub-navigation__link"
+            data-qa="tab-overview"
+            {% if tab === 'overview' %}aria-current="page"{% endif %}
+            href=""
+          >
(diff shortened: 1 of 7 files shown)


server/views/pages/overview/partials/navigation.njk
```diff
@@ -0,0 +1,53 @@
...
+            href=""
+          >
+            Overview
+          </a>
+        </li>
+        <li class="moj-sub-navigation__item">
+          <a class="moj-sub-navigation__link"
+            data-qa="tab-support-needs"
+            {% if tab === 'support-needs' %}aria-current="page"{% endif %}
+            href=""
```
mattwills-moj-digital, line 18: Minor - I would have been tempted to leave the other tabs out for now, but I'm not going to lose sleep over it.

nathanrussell-moj-digital: I know what you mean. I was tempted, but decided to include them because I wanted to make sure the markup rendered correctly; and our tab headings are slightly different to CIAG's (which is where I copied from 😬 )

PR #33, opened 2023-07-10
Refactor integration tests to use pageId rather than page title

nathanrussell-moj-digital:
This PR refactors the integrations tests (before we get too far with them!) so that the page assertion is based on pageId rather than page title. We did this in SLM, and the thinking was that asserting the current page on page title was fragile as content is likely to change

### DIFF
--- a/integration_tests/pages/overview/OverviewPage.ts
+++ b/integration_tests/pages/overview/OverviewPage.ts
@@ -1,0 +1,21 @@
+import Page, { PageElement } from '../page'
+
+export default class OverviewPage extends Page {
+  constructor() {
+    super('overview')
+  }
+
+  isForPrisoner(expected: string) {
+    this.prisonNumberLabel().should('have.text', expected)
+    return this
+  }
+
+  activeTabIs(expected: string) {
+    this.activeTab().should('have.text', expected)
+    return this
+  }
+
+  prisonNumberLabel = (): PageElement => cy.get('[data-qa=prison-number]')
+
+  activeTab = (): PageElement => cy.get('.moj-sub-navigation__link[aria-current=page]')
+}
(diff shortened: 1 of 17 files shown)


integration_tests/pages/overview/OverviewPage.ts
```diff
@@ -0,0 +1,21 @@
+import Page, { PageElement } from '../page'
+
+export default class OverviewPage extends Page {
```
nathanrussell-moj-digital, line 3: We have no integration tests using this yet, but I've created it ready 👍

integration_tests/pages/page.ts
```diff
@@ -55,4 +55,9 @@ export default abstract class Page {
     cy.get(`#${field}-error`).should('exist')
     return this
   }
+
+  hasMainHeading(expectedHeading: string) {
```
nathanrussell-moj-digital, line 59: If the test does care about page title we can call this assertion method

server/views/partials/layout.njk
```diff
@@ -26,6 +27,25 @@
 {% endblock %}
 
 {% block bodyStart %}
+  <span class="govuk-visually-hidden" id="pageId" data-qa="{{ pageId }}"></span>
```
nathanrussell-moj-digital, line 30: This is where the pageId is added to the page, in a hidden span

integration_tests/pages/page.ts
```diff
@@ -19,7 +19,7 @@ export default abstract class Page {
   }
 
   checkOnPage(): void {
-    cy.get('h1').contains(this.title)
+    cy.get('#pageId').should('have.attr', 'data-qa').should('equal', this.pageId)
```
mattwills-moj-digital, line 22: 👍

mattwills-moj-digital (approved):
GS

PR #37, opened 2023-07-12
RR-69 - Integrate Overview page in journey of other pages

nathanrussell-moj-digital:
This PR integrates the new Overview page with the other pages in our journey.
It is now set as being the first page in the Create Goal journey, in that you have to start on this page and click the `Add a goal` button. You can no longer navigate directly to the Create Goal page.

The only consideration with this PR is when we should merge it in relation to the testing that the team are doing for RR-2 - we will probably need to liaise with them to let them know about the screen flow change


mattwills-moj-digital (approved):
ATOB

PR #42, opened 2023-07-14
RR-105 - Refactored to use the systemToken with the authenticated username

nathanrussell-moj-digital:
The PR is based on the outcome of a number slack conversations, but [mainly this one](https://mojdt.slack.com/archives/CDK7ECU3W/p1689326272000459)

The PR refactors our integration with `prisoner-search` so that it uses the system client token rather than the authenticated users token.
The reason for this change is that we don't want to have to give all the possible user's of this service the roles `GLOBAL_SEARCH` and `VIEW_PRISONER_DATA` as this is both an admin burden, but more importantly has the potential to expose prisoner data to users who are not entitled to view it.

The preferred approach (as per the slack conversation above) is to use the system token, but with the authenticated users username. This means the roles are only applied to the system client (being requested separately in https://dsdmoj.atlassian.net/browse/HAAR-1623). We use the system token, but "in the guise of" the authenticated user. This means that the logs etc for prisoner-search will know about and can reference the authenticated username, rather than some randomly named system token that will mean nothing in the logs 👍

### DIFF
--- a/server/routes/overview/overviewController.test.ts
+++ b/server/routes/overview/overviewController.test.ts
@@ -37,8 +37,8 @@ describe('overviewController', () => {
 
   it('should get overview view given prisoner is retrieved from prisoner-search API', async () => {
     // Given
-    const userToken = 'a-user-token'
-    req.user.token = userToken
+    const username = 'a-dps-user'
+    req.user.username = username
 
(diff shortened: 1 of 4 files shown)


mattwills-moj-digital (approved):
LGTM

PR #44, opened 2023-07-14
RR-81 Fix to targetDateRange error message

mattwills-moj-digital:


server/validators/stepTargetDateRangeValidator.ts
```diff
@@ -1,14 +1,8 @@
-const MAX_TARGET_DATE_RANGE_LENGTH = 50
```
mattwills-moj-digital, line 1: I don't see much point in checking the max length is exceeded, given the values come from radio buttons, plus we'd like to switch to using a enum soon.

server/views/pages/goal/add-step/index.njk
```diff
@@ -54,9 +54,6 @@
             errorMessage: errors | findError('title')
           }) }}
 
-          {% set targetDateError = errors | findError('targetDate') %}
-          {% set targetDateErrorClass = " govuk-input--error" if targetDateError %}
```
mattwills-moj-digital, line 58: This was a hangover from when we were using 3 separate fields to represent a date. My bad for not updating this properly (partly because I knew even less then that I do now!).

nathanrussell-moj-digital (approved):
LGTM 👍

PR #49, opened 2023-07-17
implement the new summary list card component and remove old table row filter logic

chrimesdev:
- Use the new approved summary list card component on the review add goal screen
- Display "Notes" on the review add goal screen
- Remove old logic for the table row filter
- Create a new Nunjucks filter to form the date form snake case value into a human readable date range


server/views/pages/goal/review/index.njk
```diff
@@ -18,36 +17,46 @@
...
-          caption: 'Goal 1: ' + data.title + '',
-          captionClasses: "govuk-table__caption--m",
-          firstCellIsHeader: false,
-          head: [
-            { text: "Steps" },
-            { text: "Target date" }
-          ],
-          rows: data.steps | stepsTableRowFilter,
-          attributes: { id: 'steps-table' }
+      <div class="govuk-summary-card">
```
chrimesdev, line 21: This block is where I've used vanilla HTML instead of a Nunjucks macro. The HTML outputted is the exact same as the Nunjucks macro, however it means that can handle all the view logic in this HTML file opposed to having another complex file to populate the Nunjucks macro fields.

server/filters/formatDateFormValue.ts
```diff
@@ -0,0 +1,11 @@
+export default function formatDateFormValue(value: string): string {
```
mattwills-moj-digital, line 1: Much better than the original! 🙌

server/filters/formatDateFormValue.ts
```diff
@@ -0,0 +1,11 @@
...
+  const targetDateRange = TargetDateRange[value as keyof typeof TargetDateRange]
+  return targetDateRange
+}
+
+export enum TargetDateRange {
```
mattwills-moj-digital, line 6: I would be tempted to remove the word "target" to make it a bit more generic (i.e. not just for a step's "target" date range).

chrimesdev:
Forgot to actually use the Nunjucks custom filter, just added that in https://github.com/ministryofjustice/hmpps-education-and-work-plan-ui/pull/49/commits/92410c42b67ffef189a0bdd7761455efd8ddf37a

server/filters/formatDateFormValue.ts
nathanrussell-moj-digital: Agreed; that would be a good refactoring 👍

server/views/pages/goal/review/index.njk
nathanrussell-moj-digital: I also dont feel best qualified to say which is better. I guess its on a case by case basis, and probably based around the amount of code we'd have to write either setting variables in the view (😬 ) and/or filters and such like

What you've done looks fine though, and I can see why you are proposing this approach.

My only caution here would be that we dont take this approach as our defacto / go-to approach. I think we should use the govuk macros where ever possible/practical; and only look to do what we've done here when there is a clear benefit / reason for doing so 👍

nathanrussell-moj-digital (approved):
LGTM 👍

mattwills-moj-digital (dismissed):

server/filters/formatDateFormValue.ts
chrimesdev: Done in https://github.com/ministryofjustice/hmpps-education-and-work-plan-ui/pull/49/commits/6dbed7e2d02d5258ee96d1d53d769fc6bb52d1ba

server/views/pages/goal/review/index.njk
chrimesdev: Yep completely agree with that!

mattwills-moj-digital (approved):
LGTM

PR #53, opened 2023-07-18
RR-135 Service and view classes for Curious data

mattwills-moj-digital:
A few things to resolve. Very much WIP at the moment.


server/routes/overview/mappers/supportNeedsMapper.test.ts
```diff
@@ -0,0 +1,51 @@
...
+      neurodivergenceSelfDeclared: ['Dyslexia'],
+      selfDeclaredDate: '2022-02-18',
+      neurodivergenceAssessed: ['No Identified Neurodiversity Need'],
+      assessmentDate: '2022-05-18',
+      neurodivergenceSupport: ['Writing support'],
+      supportDate: '2022-02-18',
+    }
+
+    const expectedSupportNeeds = {
+      languageSupportNeeded: false,
```
mattwills-moj-digital, line 31: I've no idea how to map this yet.

server/services/curiousService.ts
```diff
@@ -0,0 +1,31 @@
...
+export default class CuriousService {
+  constructor(private readonly hmppsAuthClient: HmppsAuthClient, private readonly curiousClient: CuriousClient) {}
+
+  async getPrisonerSupportNeeds(
+    prisonNumber: string,
+    establishmentId: string,
+    username: string,
+  ): Promise<SupportNeeds> {
+    // TODO - is this the correct token?
+    const systemToken = await this.hmppsAuthClient.getSystemClientToken(username)
```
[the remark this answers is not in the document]
nathanrussell-moj-digital: lol - yes, that is the correct token.
We want to use the system token in this case (and also when searching/retrieving prisoners in prisoner-search). This is because both APIs require a role(s), and we dont want the overheard of having to ensure all our users (the CIAGs etc) have these roles. It is OK to ensure that CIAGs have the role for _our_ service UI and API (which we do already), but to say that our users need roles from other services is an overhead and a security risk (they might get access to more than our service allows by directly hitting the other API; or other services they use might suddenly give them access to stuff they are not entitled to in context of that app)

So, the general pattern I'm led to believe is that the UI service itself calls these other APIs, using the system token.
As you can see in what you've done here (assume copied from prisoner-search) is that we also pass in the logged in username when getting the system token. This is quite neat - what it does is get the system token, but with a "modified" username claim in the JWT, representing the authenticated user. This means that logs etc in the other system will detail the logged in username, so gives some more context to the logs etc of those other systems.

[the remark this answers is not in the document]
mattwills-moj-digital: I thought as much - I knew you said we didn't want the overhead of ensuring each user has specific roles for other systems. I just wasn't sure if this was the correct "system" token :)

server/data/curiousClient.test.ts
```diff
@@ -51,27 +52,28 @@ describe('curiousClient', () => {
...
-      curiousApi.get(`/learnerProfile/${prisonNumber}`).reply(200, prisonerProfile)
+      curiousApi.get(`/learnerProfile/${prisonNumber}?establishmentId=${establishmentId}`).reply(200, learnerProfile)
 
       // When
-      const actual = await curiousClient.getPrisonerProfile(prisonNumber, systemToken)
+      const actual = await curiousClient.getLearnerProfile(prisonNumber, establishmentId, systemToken)
 
       // Then
-      expect(actual).toEqual(prisonerProfile)
+      expect(actual).toEqual(learnerProfile.pop())
```
nathanrussell-moj-digital, line 61: Ref: my comment below in the implementation re: popping from the array; I wonder if we need to make the test clearer - perhaps make the stub return an array of several elements, and then be explicit about which one we expect?

server/routes/overview/mappers/supportNeedsMapper.ts
```diff
@@ -0,0 +1,32 @@
...
+      selfDeclaredNeurodiversity: learnerNeurodivergence.neurodivergenceSelfDeclared,
+      assessedNeurodiversity: learnerNeurodivergence.neurodivergenceAssessed,
+    }
+  }
+  return undefined
+}
+
+const toLddAndHealthNeeds = (learnerProfile: LearnerProfile): Array<string> => {
+  if (learnerProfile && learnerProfile.primaryLDDAndHealthProblem) {
+    return [learnerProfile.primaryLDDAndHealthProblem, ...learnerProfile.additionalLDDAndHealthProblems.sort()]
```
nathanrussell-moj-digital, line 27: Do we know what property the `sort` operation is on?

server/routes/overview/overviewController.ts
```diff
@@ -7,16 +7,18 @@ export default class OverviewController {
...
     res.render('pages/overview/index', { ...view.renderArgs })
   }
 
   getSupportNeedsView: RequestHandler = async (req, res, next): Promise<void> => {
     const { prisonNumber } = req.params
     const { prisonerSummary } = req.session
+    const { supportNeeds } = req.session
 
-    const view = new OverviewView(prisonerSummary, 'support-needs', prisonNumber)
+    const view = new OverviewView(prisonerSummary, 'support-needs', prisonNumber, supportNeeds)
```
nathanrussell-moj-digital, line 21: Ah ... thats my bad from my last PR ....
I was not envisaging that `OverviewView` would need changing for this, as its the view for the Overview tab
In my last PR I created a new view class `SupportNeedsView` which I'm envisaging is used as the view for the support tab. Clearly I didnt wire it up in the controller! Sorry! My bad!

So .... I think you can revert your changes to `OverviewView` because it does not need to know about the support needs
And instead put the new properties for support needs in the `SupportNeedsView` class; then use it here on this line 21:
```
const view = new SupportNeedsView(prisonerSummary, 'support-needs', prisonNumber, supportNeeds)
```

server/routes/overview/overviewController.ts
```diff
@@ -7,16 +7,18 @@ export default class OverviewController {
     req.session.createGoalForm = undefined
 
     const { prisonerSummary } = req.session
+    const { supportNeeds } = req.session
```
nathanrussell-moj-digital, line 10: See my comment below - I think you can revert these lines 10 and 12

nathanrussell-moj-digital: (which will mean you test above needs to change)

server/data/curiousClient.ts
```diff
@@ -7,15 +7,23 @@ export default class CuriousClient {
...
 
-  async getPrisonerProfile(prisonNumber: string, token: string): Promise<Array<LearnerProfile>> {
-    return (await CuriousClient.restClient(token).get({
+  async getLearnerProfile(prisonNumber: string, establishmentId: string, token: string): Promise<LearnerProfile> {
+    const learnerProfiles = (await CuriousClient.restClient(token).get({
       path: `/learnerProfile/${prisonNumber}`,
-    })) as Promise<Array<LearnerProfile>>
+      query: `establishmentId=${establishmentId}`,
+    })) as Array<LearnerProfile>
+    return learnerProfiles.pop()
```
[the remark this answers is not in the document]
mattwills-moj-digital: The array will only contain one element, since we are passing in establishmentId. Therefore I think it makes no sense to return an array from the client.

server/routes/overview/overviewView.ts
```diff
@@ -1,21 +1,24 @@
-import type { PrisonerSummary } from 'viewModels'
+import type { PrisonerSummary, SupportNeeds } from 'viewModels'
```
nathanrussell-moj-digital, line 1: Re: my comment above - revert the changes in this class, but make the same changes to `SupportNeedsView` instead 👍

server/routes/overview/mappers/supportNeedsMapper.ts
mattwills-moj-digital: I don't follow? My intention is that the primary LDD is listed first and the additional ones are ordered alphabetically. I'm basing this on what Mons sent me: https://mojdt.slack.com/archives/C02497NS0HK/p1689662633047629?thread_ts=1689607205.601469&cid=C02497NS0HK

server/routes/overview/prisonerSupportNeedsRequestHandler.test.ts
```diff
@@ -0,0 +1,112 @@
...
+      next as undefined as NextFunction,
+    )
+
+    // Then
+    expect(curiousService.getPrisonerSupportNeeds).not.toHaveBeenCalled()
+    expect(req.session.supportNeeds).toEqual(supportNeeds)
+    expect(next).toHaveBeenCalled()
+  })
+
+  it('should call next function with error given retrieving support needs fails', async () => {
```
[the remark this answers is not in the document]
nathanrussell-moj-digital: Yes, as discussed earlier I think we need to handle and treat a 404 as a "happy path", in that for our app it means the prisoner has not had any support needs assessments done yet

server/routes/overview/prisonerSupportNeedsRequestHandler.ts
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Class exposing middleware functions related to ensuring we have a prisoner's support needs.
+ */
+export default class PrisonerSupportNeedsRequestHandler {
+  constructor(private readonly curiousService: CuriousService) {}
+
+  /**
+   *  Middleware function to look up the prisoner's Support Needs, and store them in the session
+   */
+  getPrisonerSupportNeeds: RequestHandler = async (req: Request, res: Response, next: NextFunction) => {
```
nathanrussell-moj-digital, line 14: This has not been wired in to the route yet? Is that right?

server/routes/overview/prisonerSupportNeedsRequestHandler.ts
```diff
@@ -0,0 +1,31 @@
...
+      if (!req.session.supportNeeds) {
+        req.session.supportNeeds = await this.curiousService.getPrisonerSupportNeeds(
+          prisonNumber,
+          prisonerSummary.establishmentId,
+          req.user.username,
+        )
+      }
+      next()
+    } catch (error) {
+      next(createError(404, 'Prisoner Support Needs not found')) // TODO - what should happen here?
```
nathanrussell-moj-digital, line 28: Agreed - we need to work out what to do in the case of a 404 vs. other error status codes
Perhaps leave it for now in this PR, and we do it afterwards (next PR), otherwise this PR could become pretty big?

server/@types/viewModels/index.d.ts
```diff
@@ -1,9 +1,22 @@
...
     lastName: string
   }
+
+  export interface SupportNeeds {
+    languageSupportNeeded: boolean
+    lddAndHealthNeeds: Array<string>
+    neurodiversity: Neurodiversity
+  }
+
+  export interface Neurodiversity {
```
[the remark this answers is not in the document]
mattwills-moj-digital: Good point. I forgot about those dates!

server/data/curiousClient.ts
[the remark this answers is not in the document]
mattwills-moj-digital: Changed back to an array of one or more elements.

server/routes/overview/mappers/prisonerSupportNeedsMapper.test.ts
```diff
@@ -0,0 +1,107 @@
...
+        neurodivergenceAssessed: ['ADHD'],
+        assessmentDate: '2022-05-18',
+        neurodivergenceSupport: ['Writing support'],
+        supportDate: '2022-02-18',
+      },
+      {
+        prn: 'G6123VU',
+        establishmentId: 'DNI',
+        establishmentName: 'DONCASTER (HMP)',
+        neurodivergenceSelfDeclared: [],
```
mattwills-moj-digital, line 48: I've asked if this will ever be null/undefined. I know this is only test data, but in the arrays below that we have `No identified...` type values, so I'm wondering if we can expect that kind of value as a default.

server/routes/overview/mappers/supportNeedsMapper.test.ts
mattwills-moj-digital: I'm doing a straight mapping from `languageStatus` for now. The business are going to clarify what the requirements are here.

server/routes/overview/mappers/prisonerSupportNeedsMapper.test.ts
mattwills-moj-digital: I have defensively coded around the date fields and arrays (where relevant), expecting them to be null.

nathanrussell-moj-digital (approved):
We've paired on this a couple of times today, and it looks good
A few things to wire up in your next PRs, but this is good to merge as a building block for getting it all wired 👍

PR #58, opened 2023-07-20
Feature/RR-135 Support needs data retrieval

mattwills-moj-digital:
This PR pulls in some great stuff from @nathanrussell-moj-digital's [PR](https://github.com/ministryofjustice/hmpps-education-and-work-plan-ui/pull/61) earlier:

* Retrieval of the support needs data on request, rather than storing it in the session.
* Cheerio tests for the support needs view
* Integration test for when curious been down


integration_tests/pages/overview/OverviewPage.ts
```diff
@@ -37,7 +37,7 @@ export default class OverviewPage extends Page {
   }
 
   hasCuriousUnavailableMessageDisplayed() {
-    cy.get('h2').contains('Sorry, Functional Skills data from Curious is unavailable')
+    cy.get('h2').contains('Sorry, the Curious service is currently unavailable')
```
nathanrussell-moj-digital, line 40: 🤐 
😁

server/routes/overview/overviewController.ts
```diff
@@ -18,9 +18,10 @@ export default class OverviewController {
   }
 
   getSupportNeedsView: RequestHandler = async (req, res, next): Promise<void> => {
+    const { prisonNumber } = req.params
     const { prisonerSummary } = req.session
-    const { supportNeeds } = req.session
 
+    const supportNeeds = await this.curiousService.getPrisonerSupportNeeds(prisonNumber, req.user.username)
```
nathanrussell-moj-digital, line 24: 👍 
The more I think about it, the more I think this the right way to do data like this (for now at least)
By not storing it in the session and retrieving it every time, it means that if curious is temporarily down, then trying again shortly afterwards would work (if curious was back up), and would not get the duff curious data from the session.

Efficient to call it on each page load? Perhaps not, though that is unproven at this stage. We can consider some kind of caching as and when this is proven to be a problem 👍

server/services/curiousService.test.ts
```diff
@@ -69,25 +69,58 @@ describe('curiousService', () => {
...
 
       // Then
-      expect(actual).toEqual(Error('Not Found'))
+      expect(actual).toEqual(expectedSupportNeeds)
+      expect(curiousClient.getLearnerProfile).toHaveBeenCalledWith(prisonNumber, systemToken)
+      expect(curiousClient.getLearnerNeurodivergence).not.toHaveBeenCalled()
     })
   })
 
+  // TODO figure out how to throw a 404 not found error
```
nathanrussell-moj-digital, line 98: Nice - I think we need a little story on the backlog to address 404's from Curious 👍 
I'll add one shortly 👍

server/views/pages/overview/partials/educationAndTrainingTabContents.njk
```diff
@@ -10,7 +10,7 @@
 
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
-      <h2 class="govuk-heading-m">Sorry, Functional Skills data from Curious is unavailable</h2>
+      <h2 class="govuk-heading-m">Sorry, the Curious service is currently unavailable.</h2>
```
nathanrussell-moj-digital, line 13: 🤐

server/views/pages/overview/partials/supportNeedsTabContents.njk
```diff
@@ -2,7 +2,7 @@
   <div class="govuk-grid-column-full">
 
     {% if supportNeeds.problemRetrievingData === true %}
-      There was a problem retrieving the data from Curious.
+      <h2 class="govuk-heading-m">Sorry, the Curious service is currently unavailable.</h2>
```
nathanrussell-moj-digital, line 5: Minor (and we can do it another time), but I wonder if a neater & easier to maintain approach to these views that contain error messages is like what I did in `educationAndTrainingTabContents.njk`

I put the if block right at the top, and I made the true branch handle the happy path (ie. render the data in the various components etc)
I put the error state (sad path) right down the bottom in the else block

The other thing I did was make the if/else/endif statements the outer extent / scope of the template (rather than putting them within the divs).
The reason I did this is because it means the true and false blocks are self contained in respect of their markup .... I'm basically thinking that potentially the error markup might get pulled out into a partial or similar at some stage, so you'd kinda want it "complete" if that makes sense?

nathanrussell-moj-digital (dismissed):
LGTM (apart from you changing my beautiful wording on my error message! Took me ages to come up with that, and you just trash it and throw it away! 🤣 )

The only other minor comment is about the if/else block in your njk template - but its really minor

server/services/curiousService.test.ts
nathanrussell-moj-digital: I've created RR-155

server/views/pages/overview/partials/supportNeedsTabContents.njk
mattwills-moj-digital: I think that's a good idea and I've done the same 👍

server/routes/overview/overviewController.ts
mattwills-moj-digital: Yeah I think we might want to look at some caching at some stage.

nathanrussell-moj-digital (approved):
AToAB 👍 😁

PR #59, opened 2023-07-20
use the correct page title and breadcrumbs from the dps team

chrimesdev:


server/views/partials/breadCrumb.njk
```diff
@@ -1,11 +1,11 @@
 {{ govukBreadcrumbs({
```
nathanrussell-moj-digital, line 1: Minor - do we need this file as a partial at all now?
Are any pages actually bringing it in now?

server/views/pages/goal/add-note/index.njk
```diff
@@ -4,7 +4,23 @@
...
+        text: "Digital Prison Services",
+        href: dpsUrl
+      },
+      {
+        text: "Learning and work progress",
+        href: "/"
+      },
+      {
+        text: "Manage " + prisonerSummary.firstName | capitalize + " " + prisonerSummary.lastName | capitalize + "'s learning plan",
+        href: "/"
```
nathanrussell-moj-digital, line 19: Minor - should this href be linking to `/` ? or should it be linking to the prisoner overview page as per your other changes?

server/views/partials/breadCrumb.njk
chrimesdev: Yep makes sense to remove as it was only used on 1 page

server/views/pages/goal/add-note/index.njk
chrimesdev: Good spot resolved

chrimesdev:
Should be ready for a final review @nathanrussell-moj-digital

chrimesdev:
I've included the captions as part of this PR too as it a was a minor change and slightly related to the changing of page titles. 

I've not added it to all pages as theres an outstanding question about whether it should be added to question pages, see: https://mojdt.slack.com/archives/C03C8KV2A7N/p1690207112001479?thread_ts=1689772474.096239&cid=C03C8KV2A7N

nathanrussell-moj-digital (approved):
LGTM 👍 

My only question is whether we need to include the caption on the Goals creation pages? (create goal, add step, add note, review goal) - do we need to do it on those as well?

chrimesdev:
> LGTM 👍
> 
> My only question is whether we need to include the caption on the Goals creation pages? (create goal, add step, add note, review goal) - do we need to do it on those as well?

I've left it off the question pages for now because it isn't clear and goes against the GOV.UK Design System, I've asked the DPS team: https://mojdt.slack.com/archives/C03C8KV2A7N/p1690207112001479?thread_ts=1689772474.096239&cid=C03C8KV2A7N

PR #61, opened 2023-07-20
RR-116 - Wired in service; added curious unavailable message; added tests

nathanrussell-moj-digital:
This PR adds the service call to the controller method in order to retrieve the functional skills from Curious
It also includes error handling to catch errors from Curious and display a temp error message
And I have added unit tests for the logic in the view; and an integration test for the new tab on the Overview page

Screenshot 2023-07-21 at 09 12 22

### DIFF
--- a/feature.env
+++ b/feature.env
@@ -2,9 +2,13 @@ PORT=3007
 HMPPS_AUTH_URL=http://localhost:9091/auth
 TOKEN_VERIFICATION_API_URL=http://localhost:9091/verification
 TOKEN_VERIFICATION_ENABLED=true
+
 EDUCATION_AND_WORK_PLAN_API_URL=http://localhost:9091
 PRISONER_SEARCH_API_URL=http://localhost:9091
+CURIOUS_API_URL=http://localhost:9091
+DPS_URL=https://digital-dev.prison.service.justice.gov.uk
 NODE_ENV=development
+
 API_CLIENT_ID=clientid
 API_CLIENT_SECRET=clientsecret
 SYSTEM_CLIENT_ID=clientid
--- a/integration_tests/e2e/overview/educationAndTrainingTab.cy.ts
+++ b/integration_tests/e2e/overview/educationAndTrainingTab.cy.ts
@@ -1,0 +1,27 @@
+import OverviewPage from '../../pages/overview/OverviewPage'
+import Page from '../../pages/page'
+
+context('Prisoner Overview page - Education And Training tab', () => {
+  beforeEach(() => {
+    cy.task('reset')
+    cy.task('stubSignInAsUserWithEditAuthority')
+    cy.task('stubAuthUser')
+    cy.task('getPrisonerById')
+  })
+
+  it('should display curious unavailable message given curious is unavailable', () => {
+    // Given
+    cy.signIn()
+    const prisonNumber = 'G6115VJ'
+    cy.visit(`/plan/${prisonNumber}/view/overview`)
+    const overviewPage = Page.verifyOnPage(OverviewPage)
+
+    // When
+    overviewPage.selectTab('Education and training')
+
+    // Then
+    overviewPage //
+      .activeTabIs('Education and training')
+      .hasCuriousUnavailableMessageDisplayed()
+  })
+})
--- a/integration_tests/e2e/overview/educationAndTrainingTab.cy.ts
+++ b/integration_tests/e2e/overview/educationAndTrainingTab.cy.ts
@@ -11,6 +11,8 @@ context('Prisoner Overview page - Education And Training tab', () => {
 
   it('should display curious unavailable message given curious is unavailable', () => {
     // Given
+    cy.task('stubLearnerProfile401Error')
+
     cy.signIn()
     const prisonNumber = 'G6115VJ'
     cy.visit(`/plan/${prisonNumber}/view/overview`)
--- a/integration_tests/mockApis/curiousApi.ts
+++ b/integration_tests/mockApis/curiousApi.ts
@@ -138,6 +138,23 @@ const stubLearnerProfile = (prisonNumber = 'G6115VJ'): SuperAgentRequest =>
     },
   })
 
+const stubLearnerProfile401Error = (prisonNumber = 'G6115VJ'): SuperAgentRequest =>
+  stubFor({
+    request: {
+      method: 'GET',
+      urlPattern: `/learnerProfile/${prisonNumber}`,
+    },
+    response: {
+      status: 401,
+      headers: { 'Content-Type': 'application/json;charset=UTF-8' },
+      jsonBody: {
+        errorCode: 'VC4001',
+        errorMessage: 'Invalid token',
+        httpStatusCode: 401,
+      },
+    },
+  })
+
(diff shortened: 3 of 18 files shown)


feature.env
```diff
@@ -2,9 +2,13 @@ PORT=3007
 HMPPS_AUTH_URL=http://localhost:9091/auth
 TOKEN_VERIFICATION_API_URL=http://localhost:9091/verification
 TOKEN_VERIFICATION_ENABLED=true
+
 EDUCATION_AND_WORK_PLAN_API_URL=http://localhost:9091
 PRISONER_SEARCH_API_URL=http://localhost:9091
+CURIOUS_API_URL=http://localhost:9091
+DPS_URL=https://digital-dev.prison.service.justice.gov.uk
 NODE_ENV=development
+
```
nathanrussell-moj-digital, line 11: The env file for `start-feature` (which is used by our cypress integration tests) was missing a couple of entries, notably for this PR the endpoint for curious API pointing at wiremock

integration_tests/e2e/overview/educationAndTrainingTab.cy.ts
```diff
@@ -0,0 +1,29 @@
...
+    const overviewPage = Page.verifyOnPage(OverviewPage)
+
+    // When
+    overviewPage.selectTab('Education and training')
+
+    // Then
+    overviewPage //
+      .activeTabIs('Education and training')
+      .hasCuriousUnavailableMessageDisplayed()
+  })
```
nathanrussell-moj-digital, line 28: Happy path tests can be added to this once the Functional Skills data has been wired into the view (which is not in this PR)

integration_tests/mockApis/curiousApi.ts
```diff
@@ -138,13 +138,32 @@ const stubLearnerProfile = (prisonNumber = 'G6115VJ'): SuperAgentRequest =>
     },
   })
 
+const stubLearnerProfile401Error = (prisonNumber = 'G6115VJ'): SuperAgentRequest =>
```
nathanrussell-moj-digital, line 141: We needed an error stub for getting the curious learner profile, in order to display the 'curious is down, please try later' type message

integration_tests/pages/overview/OverviewPage.ts
```diff
@@ -31,6 +31,16 @@ export default class OverviewPage extends Page {
     return this
   }
 
+  selectTab(targetTab: string) {
```
nathanrussell-moj-digital, line 34: This method on the Overview page class allows us to click tab headings to move from tab to tab 👍

package.json
```diff
@@ -141,6 +141,7 @@
     "@typescript-eslint/eslint-plugin": "^5.59.11",
     "@typescript-eslint/parser": "^5.59.11",
     "audit-ci": "^6.6.1",
+    "cheerio": "^1.0.0-rc.12",
```
nathanrussell-moj-digital, line 144: Added `cheerio` as a dev dependency. This library is what we use in the unit tests that test the logic in the nunjucks templates (see below 👇 )

server/routes/overview/educationAndTrainingView.ts
```diff
@@ -1,21 +1,17 @@
...
-    private readonly tab: string,
-    private readonly prisonNumber: string,
-  ) {}
+  constructor(private readonly prisonerSummary: PrisonerSummary, private readonly functionalSkills: FunctionalSkills) {}
 
   get renderArgs(): {
     prisonerSummary: PrisonerSummary
     tab: string
-    prisonNumber: string
+    functionalSkills: FunctionalSkills
```
nathanrussell-moj-digital, line 9: In these view classes, the prison number is not needed (except the main `OverviewView` class); and the tab name can be hardcoded because thats essentially what the whole class is for (ie. you wouldnt use the `EducationAndTrainingView` class when displaying the Support Needs tab)

server/routes/overview/index.ts
```diff
@@ -11,17 +11,19 @@ import PrisonerSupportNeedsRequestHandler from './prisonerSupportNeedsRequestHan
 export default (router: Router, services: Services) => {
   const prisonerSummaryRequestHandler = new PrisonerSummaryRequestHandler(services.prisonerSearchService)
   const prisonerSupportNeedsRequestHandler = new PrisonerSupportNeedsRequestHandler(services.curiousService)
-  const overViewController = new OverviewController()
+  const overViewController = new OverviewController(services.curiousService)
```
nathanrussell-moj-digital, line 14: I'm exploring a slightly different pattern here .... I'm wiring the curious service into the controller, so that the controller is responsible for getting (in this case) functional skills from Curious, rather than doing it in a request handler that stores it in the session

My thinking is that this data is a bit different to the main prisoner summary. The prisoner summary is used by most pages, so storing it in the session makes sense.
But data that is only shown on 1 tab (ie. functional skills on Education And Training; and possibly learning needs etc on Support Needs) does not need to go in the session (and in cases where we change prisoner etc, it might end up being wrong!)

server/routes/overview/overviewController.ts
```diff
@@ -2,32 +2,35 @@ import { RequestHandler } from 'express'
...
     res.render('pages/overview/index', { ...view.renderArgs })
   }
 
-  getEducationAndTrainingNeeds: RequestHandler = async (req, res, next): Promise<void> => {
+  getEducationAndTrainingView: RequestHandler = async (req, res, next): Promise<void> => {
     const { prisonNumber } = req.params
     const { prisonerSummary } = req.session
 
-    const view = new EducationAndTrainingView(prisonerSummary, 'support-needs', prisonNumber)
+    const functionalSkills = await this.curiousService.getPrisonerFunctionalSkills(prisonNumber, req.user.username)
```
nathanrussell-moj-digital, line 32: Here is where we get the functional skills from the curious service, right where we need it in respect of adding it to the view. It is not added to the session or cached in any way.

(I only mention caching here because I suspect we might need to do it one day, but at the moment it would be a premature optimisation; and I'm not convinced adding it to the session would be the right way to do it in this case)

server/utils/nunjucksSetup.ts
```diff
@@ -29,6 +29,10 @@ export default function nunjucksSetup(app: express.Express, applicationInfo: App
     })
   }
 
+  registerNunjucks(app)
```
nathanrussell-moj-digital, line 32: This diff is hard to read, but basically I had to refactor the `nunkucksSetup` file a little.
Previously it had 1 function in it that did everything in respect of setting up the nunjucks env.

The new test that tests the logic in the nunjucks view needs a nunjucks environment spun up (in order for it to know how to process nunjucks templates and constructs etc), but it doesnt need everything that this function did

So I split it into 2 functions, where the first (original function) calls the 2nd (new) function . The 2nd (new) function has everything that was moved out of the first function, and is also called by the test.

See test below 👇

server/routes/overview/index.ts
mattwills-moj-digital: Makes sense, I'll do the same 👍

mattwills-moj-digital (approved):
Plenty for me to steal here 😉

PR #64, opened 2023-07-21
RR-116 - Flesh out view; cypress integration test

nathanrussell-moj-digital:
This PR handles the happy path for the Education and Training tab - Functional Skills data returned from Curious is now rendered to the page.

Contrary to the design in the prototype and story (which is a hybrid GDS component), I have chosen to use a simple govuk table for this data. Semantically it's correct (it's rows of tabular data 🤷‍♂️ ), and personally I think it looks clean and elegant (but I'm no designer!) - will create a subtask for @chrimesdev to choose and implement a suitable component if necessary.
Which ever component / design we use; we will update the story accordingly

Screenshot 2023-07-21 at 13 50 24

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ To update the types from the Open API docs run the following commands:
 
 `npx openapi-typescript https://prisoner-offender-search-dev.prison.service.justice.gov.uk/v3/api-docs -o server/@types/prisonerSearchApi/index.d.ts`
 
-`npx openapi-typescript https://raw.githubusercontent.com/ministryofjustice/curious-API/main/curious-api-specification.yaml -o server/@types/curiousApi/index.d.ts`
+`npx openapi-typescript https://testservices.sequation.net/sequation-virtual-campus2-api/v3/api-docs -o server/@types/curiousApi/index.d.ts`
 
 Note that you will need to run prettier over the generated files and possibly handle other errors before compiling.
 
--- a/integration_tests/e2e/overview/educationAndTrainingTab.cy.ts
+++ b/integration_tests/e2e/overview/educationAndTrainingTab.cy.ts
@@ -9,6 +9,24 @@ context('Prisoner Overview page - Education And Training tab', () => {
     cy.task('getPrisonerById')
   })
 
+  it('should display Education and Training data', () => {
+    // Given
+    cy.task('stubLearnerProfile')
+
+    cy.signIn()
+    const prisonNumber = 'G6115VJ'
+    cy.visit(`/plan/${prisonNumber}/view/overview`)
+    const overviewPage = Page.verifyOnPage(OverviewPage)
+
+    // When
+    overviewPage.selectTab('Education and training')
+
+    // Then
+    overviewPage //
+      .activeTabIs('Education and training')
+      .hasFunctionalSkillsDisplayed()
+  })
+
   it('should display curious unavailable message given curious is unavailable', () => {
     // Given
     cy.task('stubLearnerProfile401Error')
--- a/server/views/pages/overview/partials/educationAndTrainingTabContents.njk
+++ b/server/views/pages/overview/partials/educationAndTrainingTabContents.njk
@@ -3,6 +3,30 @@
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
 
+      <table class="govuk-table" id="functional-skills-table">
+        <caption class="govuk-table__caption govuk-table__caption--m">Functional skills</caption>
+
+        <thead class="govuk-table__head">
+          <tr class="govuk-table__row">
+            <th scope="col" class="govuk-table__header">Subject</th>
+            <th scope="col" class="govuk-table__header">Assessed on</th>
+            <th scope="col" class="govuk-table__header">Level</th>
+            <th scope="col" class="govuk-table__header">Type</th>
+          </tr>
+        </thead>
+
+        <tbody class="govuk-table__body">
+        {% for assessment in functionalSkills.assessments %}
+          <tr class="govuk-table__row">
+            <td class="govuk-table__cell">{{ assessment.type | formatFunctionalSkillType }}</td>
+            <td class="govuk-table__cell">{{ assessment.assessmentDate | formatDate('D MMMM YYYY') }}</td>
+            <td class="govuk-table__cell">{{ assessment.grade }}</td>
+            <td class="govuk-table__cell">Induction</td>
+          </tr>
+        {% endfor %}
+        </tbody>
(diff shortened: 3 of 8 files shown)


server/views/pages/overview/partials/educationAndTrainingTabContents.njk
```diff
@@ -3,6 +3,30 @@
...
+          </tr>
+        </thead>
+
+        <tbody class="govuk-table__body">
+        {% for assessment in functionalSkills.assessments %}
+          <tr class="govuk-table__row">
+            <td class="govuk-table__cell">{{ assessment.type | formatFunctionalSkillType }}</td>
+            <td class="govuk-table__cell">{{ assessment.assessmentDate | formatDate('D MMMM YYYY') }}</td>
+            <td class="govuk-table__cell">{{ assessment.grade }}</td>
+            <td class="govuk-table__cell">Induction</td>
```
nathanrussell-moj-digital, line 24: I have hardcoded the word "Induction" here, as at the moment all functional skills are from the prisoner's CIAG induction

server/views/pages/overview/partials/educationAndTrainingTabContents.njk
```diff
@@ -12,8 +36,8 @@
     <div class="govuk-grid-column-two-thirds">
       <h2 class="govuk-heading-m">Sorry, Functional Skills data from Curious is unavailable</h2>
       <p class="govuk-body">
-        Retrieving data from Curious is currently unavailable. Please try again later. Other functions within this
-        service may still be available.
+        Retrieving data from Curious is currently unavailable. Please try again later. Other functions within
```
nathanrussell-moj-digital, line 39: The code formatter made this change! 😁

integration_tests/e2e/overview/educationAndTrainingTab.cy.ts
```diff
@@ -9,6 +9,24 @@ context('Prisoner Overview page - Education And Training tab', () => {
...
+    cy.visit(`/plan/${prisonNumber}/view/overview`)
+    const overviewPage = Page.verifyOnPage(OverviewPage)
+
+    // When
+    overviewPage.selectTab('Education and training')
+
+    // Then
+    overviewPage //
+      .activeTabIs('Education and training')
+      .hasFunctionalSkillsDisplayed()
```
mattwills-moj-digital, line 27: I need to add a test like this for support needs 👍

server/views/pages/overview/partials/educationAndTrainingTabContents.test.ts
```diff
@@ -22,6 +23,33 @@ describe('Education and Training tab view', () => {
     compiledTemplate = nunjucks.compile(template.toString(), njkEnv)
   })
 
+  it('should render content', () => {
```
mattwills-moj-digital, line 26: I'm just considering the benefit these 'cheerio' tests give us in addition to the cypress ones? I know we have the equivalent in Spring Boot (i.e. unit and integration tests), but our cypress tests our driven via the browser, so don't they test the view inherently?

Also, remember that in Spring Boot we tend to avoid unit tests for Controllers because they contain little logic and are tested via the REST based integration tests.

Don't get me wrong, I value the benefit of high quality test coverage, but I don't want another set of tests to write and maintain, unless they add a real benefit!

mattwills-moj-digital (dismissed):
LGTM.

I'm not 100% convinced by the need for the cheerio tests (partly because I'd hope our templates are reasonably simple and don't contain too much logic), but I can see some benefits as well 👍

server/views/pages/overview/partials/educationAndTrainingTabContents.test.ts
nathanrussell-moj-digital: 💯 You are absolutely right; and I agree 👍 
I think cypress, cheerio, and anything else we use; are just available tools in our toolbox. And our job is to use the right tool for the job 👍 

I think cheerio definitely has it's place, as do our cypress tests. So I'd be keen that we don't blindly ditch the cheerio tests we already have and remove the dependency from `package.json` (I don't think you were suggesting that)

I think the initial cheerio test I added was to introduce the library and introduce a test or two so that we could see how they are written and where they might add value etc. (to be fair I didnt make that clear though)

You are right that there is a grey line / possible overlap between the cheerio and cypress tests; and I would agree that the cypress and cheerio tests we currently have definitely overlap.

As with the Spring Boot integration tests, I would consider the cypress tests "slow", and we perhaps don't want to over do them. We don't want to do every permutation of screen flow / journey in cypress tests, just as we wouldn't with our SB integration tests. 

All the different permutations / code paths is where the unit tests sit 👍 

I see the cheerio library as part of our unit test tool box, not integration tests. With that in mind its perfect for testing any branching, looping or (possibly?) filters used in the njk templates.

As an example (and we dont have a test for this) - but the Overview page has several code branches (and I suspect there'll be more as we add more functionality):
```
Render main overview page template
If tab is Overview
    include overview tab template
If tab is Support Needs
    include support needs template
    if not problemRetrievingData
        render supper needs table
        loop through support needs
            render a row for each support need
        render selfDeclaredNeurodiversity table
        loop through selfDeclaredNeurodiversities
            render a row for each selfDeclaredNeurodiversity
        render assessedNeurodiversity table
        loop through assessedNeurodiversities
            render a row for each assessedNeurodiversity
    else
        render curious error message
If tab is Education And Training
    include education and training template
    if not problemRetrievingData
        render functional skills table
        loop through functional skills
            render a row for each functional skill
    else
        render curious error message


```

mattwills-moj-digital: A well reasoned response! 👍 😄

README.md
```diff
@@ -61,7 +61,7 @@ To update the types from the Open API docs run the following commands:
 
 `npx openapi-typescript https://prisoner-offender-search-dev.prison.service.justice.gov.uk/v3/api-docs -o server/@types/prisonerSearchApi/index.d.ts`
 
-`npx openapi-typescript https://raw.githubusercontent.com/ministryofjustice/curious-API/main/curious-api-specification.yaml -o server/@types/curiousApi/index.d.ts`
+`npx openapi-typescript https://testservices.sequation.net/sequation-virtual-campus2-api/v3/api-docs -o server/@types/curiousApi/index.d.ts`
```
nathanrussell-moj-digital, line 64: Took the opportunity to use the official location of the curious openAPI spec; rather than the one I found in a MoJ repo !!!
(they are the same file though 👍 )

mattwills-moj-digital (approved):
What a way to finish the week! 😉

PR #70, opened 2023-07-25
RR-158 - Retrieve a Prisoner's Action Plan

mattwills-moj-digital:
This PR adds the service, client and mapper for retrieving a Prisoner's Action Plan.


server/data/educationAndWorkPlanClient.test.ts
```diff
@@ -42,4 +33,21 @@ describe('educationAndWorkPlanClient', () => {
...
+      const systemToken = 'a-system-token'
+      const actionPlanResponse = aValidActionPlanResponseWithOneGoal()
+      educationAndWorkPlanApi.get(`/action-plans/${prisonNumber}`).reply(200, actionPlanResponse)
+
+      // When
+      const actualPlanDto = await educationAndWorkPlanClient.getActionPlan(prisonNumber, systemToken)
+
+      // Then
+      expect(nock.isDone()).toBe(true)
+      expect(actualPlanDto.prisonNumber).toEqual(prisonNumber)
```
nathanrussell-moj-digital, line 50: Any thoughts re: creating an expected action plan DTO, and asserting actualPlanDto equals the expected one? 🤔

server/data/mappers/actionPlanMapper.ts
```diff
@@ -0,0 +1,38 @@
...
+  return {
+    stepReference: stepResponse.stepReference,
+    title: stepResponse.title,
+    targetDateRange: stepResponse.targetDateRange,
+    status: stepResponse.status,
+    sequenceNumber: stepResponse.sequenceNumber,
+  }
+}
+
+export { toActionPlanDto, toGoalDto, toStepDto }
```
nathanrussell-moj-digital, line 38: Minor (and I'm not going to lose any sleep either way), but are `toGoalDto` and `toStepDto` used elsewhere? Do we need to export them or can they effectively be private within the module?

server/services/educationAndWorkPlanService.test.ts
```diff
@@ -45,4 +47,36 @@ describe('educationAndWorkPlanService', () => {
...
+      const userToken = 'a-user-token'
+      const actionPlan = aValidActionPlanDtoWithOneGoal()
+      educationAndWorkPlanClient.getActionPlan.mockImplementation(() => Promise.resolve(actionPlan))
+
+      // When
+      await educationAndWorkPlanService.getActionPlan(prisonNumber, userToken)
+
+      // Then
+      expect(educationAndWorkPlanClient.getActionPlan).toHaveBeenCalledWith(prisonNumber, userToken)
+    })
```
nathanrussell-moj-digital, line 64: Should we also be asserting the return value of the function?
IE. should line 60 be:
```
const actual = await educationAndWorkPlanService.getActionPlan(prisonNumber, userToken)
```
And then we make some assertions about `actual` ?

nathanrussell-moj-digital:
A couple of questions re: a couple of tests; but other than that its looking good 👍

server/data/educationAndWorkPlanClient.test.ts
mattwills-moj-digital: I considered that, but I didn't see the need to test the mapping here (above and beyond the test for the mapper itself).

Note that it would require creating a DTO with the exact same data that's in the `actionPlanResponse` (i.e. a different object), which seemed more trouble than it's worth.

server/data/mappers/actionPlanMapper.ts
mattwills-moj-digital: Another thing I considered! I'll remove them, as probably not, but I thought they could come in useful! 😂

server/services/educationAndWorkPlanService.test.ts
mattwills-moj-digital: Ok, I'll that in. I wasn't too bothered as it's just testing what's returned from the mock above, but I guess you could argue we need it for regression purposes 👍

nathanrussell-moj-digital (approved):
LGTM 👍 
AToAB 😁

PR #74, opened 2023-07-27
add reception date and DOB to the prisoner summary bar

chrimesdev:
- Used “string” in the interface as its coming out of the API as a string 
- Removed prisoner location as this is no longer used in the summary bar
- Added a visually hidden text paragraph above the prison summary bar for screen readers context (confirmed with Johanna)
- After speaking with Johanna, she said to leave the text as is if there is no data for a label, she is going to check with content community and other projects for consistency on the labelling and handling of no data


mattwills-moj-digital (approved):
LGTM

PR #75, opened 2023-07-27
RR-161 - Handle unexpected error from education-and-work-plan-api

mattwills-moj-digital:
This PR handles unexpected errors from the education-and-work-plan-api. As part of this change, I've avoided the use of a DTO in the view model, as it didn't fit the pattern we've used elsewhere.

### DIFF
--- a/integration_tests/mockData/actionPlanByPrisonNumberData.ts
+++ b/integration_tests/mockData/actionPlanByPrisonNumberData.ts
@@ -66,7 +66,7 @@ const actionPlans = {
               },
               {
                 stepReference: '32992dd1-7dc6-4480-b2fc-61bc36a6a775',
-                title: 'Complete French aid course',
+                title: 'Complete French course',
                 targetDateRange: 'SIX_TO_TWELVE_MONTHS',
                 status: 'NOT_STARTED',
                 sequenceNumber: 2,
@@ -135,7 +135,7 @@ const actionPlans = {
               },
               {
                 stepReference: '32992dd1-7dc6-4480-b2fc-61bc36a6a775',
-                title: 'Complete French aid course',
+                title: 'Complete French course',
                 targetDateRange: 'SIX_TO_TWELVE_MONTHS',
                 status: 'NOT_STARTED',
                 sequenceNumber: 2,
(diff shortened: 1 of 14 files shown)


integration_tests/mockData/actionPlanByPrisonNumberData.ts
```diff
@@ -66,7 +66,7 @@ const actionPlans = {
               },
               {
                 stepReference: '32992dd1-7dc6-4480-b2fc-61bc36a6a775',
-                title: 'Complete French aid course',
```
nathanrussell-moj-digital, line 69: 😁

server/views/pages/overview/partials/overviewTabContents.njk
```diff
@@ -1,58 +1,60 @@
 <div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
 
-    <div class="moj-page-header-actions">
-  
-      <div class="moj-page-header-actions__title">
-        <h2 class="govuk-heading-m">Goals in progress</h2>
-      </div>
+  {% if actionPlan.problemRetrievingData !== true %}
```
nathanrussell-moj-digital, line 3: This if statement is in the wrong place - it looks like you've wrapped the entire tab in it
You want the if statement only wrapping the for loop

nathanrussell-moj-digital (approved):
LGTM 👍

server/views/pages/overview/partials/overviewTabContents.njk
mattwills-moj-digital: Possibly. I was thinking we don't want to show the "Add goal" button either, since our API is down.

Regardless, this will need changing when we can't retrieve functional skills etc, so I wasn't too concerned.

Any suggestions @chrimesdev, or shall we leave this as it is for now?

nathanrussell-moj-digital: On reflection I'm happy that we leave it for now - I suspect we'll be back doing stuff on this page soon (!!), so we can perhaps consider changing/moving it as and when,
I think its probably fine for now 👍

PR #76, opened 2023-07-27
use the correct markup for textarea with label as heading

chrimesdev:
Markup for a Textarea when the question label is the page heading (`h1`) was slightly incorrect, uses the correct markup from the [GOV.UK Design System - Text area](https://design-system.service.gov.uk/components/textarea/) guidance.


nathanrussell-moj-digital (approved):
LGTM 👍

PR #79, opened 2023-07-28
RR-159 - Cypress tests for the Goals section of the Overview page

mattwills-moj-digital:


server/views/pages/overview/partials/overviewTabActionPlanColumn.njk
```diff
@@ -28,7 +28,7 @@
   </div>
 
   {% for goal in actionPlan.goals %}
-    <div class="govuk-summary-card">
+    <div class="govuk-summary-card" id="goals-summary-card">
```
nathanrussell-moj-digital, line 31: Ah, you cant do this I'm afraid 😢 (and it was looking so good up until this point! 😁  )

You are giving a div an ID ... but the div is in a loop, so you could have many of them ... all with the same ID

What I would do is:
* Remove the ID attribute from the div; replace it with `qa-data="goal-summary-card"` (singular on `goal` - you are saying this is a summary card for a goal, of which there could be many)
* Change your `goalSummaryCard` method in `OverviewPage.ts` to be:
`goalSummaryCards = (): PageElement => cy.get('[data-qa=goal-summary-card]')`
IE. this will return a PageElement reference to zero or more Goal Summary Cards
* Change `hasEmptyGoalsSection` to something like
```
  hasEmptyGoalsSection() {
    this.goalSummaryCards().should('be.empty') // or should(have.length(0)) or something like that - you'll have to have a play and see what works
  }
  ```
* Change `hasGoalsDisplayed` to the inverse of the above - basically are there 1 or more elements on the page

mattwills-moj-digital: Doh - I overlooked the fact it was in a loop 🤦‍♂️

All done, thanks for the pointers 👍

nathanrussell-moj-digital (approved):
LGTM 👍

PR #85, opened 2023-08-02
RR-3 - Client method and associated code to call Update Goal API

nathanrussell-moj-digital:
This PR introduces a method in `educationAndWorkPlanClient.ts` to invoke the Update Goal API.

Because the API is a `PUT` method operation, `restClient.ts` needed some refactoring because it does not currently support `PUT` or `DELETE` (which we will need later on in the project). The changes made to the rest client in keeping with (copy and paste 😁 ) what we did on a previous project for the same reasons (Send Legal Mail)

Unit tests have been written for the new method in `educationAndWorkPlanClient.ts`; and whilst doing so I realised that both the implementation (in respect of returning promises) and the tests for the existing methods in `educationAndWorkPlanClient.ts` and `curiousClient.ts` were incorrect, so I've taken the opportunity to fix them here.

### DIFF
--- a/server/@types/educationAndWorkPlanApiClient/index.d.ts
+++ b/server/@types/educationAndWorkPlanApiClient/index.d.ts
@@ -4,6 +4,9 @@ declare module 'educationAndWorkPlanApiClient' {
   export type CreateGoalRequest = components['schemas']['CreateGoalRequest']
   export type CreateStepRequest = components['schemas']['CreateStepRequest']
 
+  export type UpdateGoalRequest = components['schemas']['UpdateGoalRequest']
+  export type UpdateStepRequest = components['schemas']['UpdateStepRequest']
+
   export type ActionPlanResponse = components['schemas']['ActionPlanResponse']
   export type GoalResponse = components['schemas']['GoalResponse']
   export type StepResponse = components['schemas']['StepResponse']
--- a/server/data/curiousClient.test.ts
+++ b/server/data/curiousClient.test.ts
@@ -60,6 +60,29 @@ describe('curiousClient', () => {
       expect(actual).toEqual(learnerProfile)
       expect(nock.isDone()).toBe(true)
     })
+
+    it('should not get learner profile given API returns an error response', async () => {
+      // Given
+      const prisonNumber = 'A1234BC'
+      const systemToken = 'a-system-token'
+
+      const expectedResponseBody = {
+        errorCode: 'VC4001',
+        errorMessage: 'Invalid token',
+        httpStatusCode: 401,
+      }
+      curiousApi.get(`/learnerProfile/${prisonNumber}`).reply(401, expectedResponseBody)
+
+      // When
+      try {
+        await curiousClient.getLearnerProfile(prisonNumber, systemToken)
+      } catch (e) {
+        // Then
+        expect(nock.isDone()).toBe(true)
+        expect(e.status).toEqual(401)
+        expect(e.data).toEqual(expectedResponseBody)
+      }
+    })
   })
 
   describe('getLearnerNeurodivergence', () => {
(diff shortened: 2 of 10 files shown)


server/@types/educationAndWorkPlanApiClient/index.d.ts
```diff
@@ -4,6 +4,9 @@ declare module 'educationAndWorkPlanApiClient' {
   export type CreateGoalRequest = components['schemas']['CreateGoalRequest']
   export type CreateStepRequest = components['schemas']['CreateStepRequest']
 
+  export type UpdateGoalRequest = components['schemas']['UpdateGoalRequest']
```
nathanrussell-moj-digital, line 7: Export the types we need out of the imported openAPI spec types

server/data/curiousClient.test.ts
```diff
@@ -60,6 +60,29 @@ describe('curiousClient', () => {
       expect(actual).toEqual(learnerProfile)
       expect(nock.isDone()).toBe(true)
     })
+
+    it('should not get learner profile given API returns an error response', async () => {
```
nathanrussell-moj-digital, line 64: New "sad path" tests for curious client, because they didnt exist previously

server/data/curiousClient.ts
```diff
@@ -8,16 +8,16 @@ export default class CuriousClient {
   }
 
   async getLearnerProfile(prisonNumber: string, token: string): Promise<Array<LearnerProfile>> {
-    const learnerProfiles = (await CuriousClient.restClient(token).get({
+    const learnerProfiles = CuriousClient.restClient(token).get({
```
nathanrussell-moj-digital, line 11: Previously I dont think we were actually returning the promise, as we had the `await` keyword ... not even sure how it compiled or ran correctly at runtime to be honest because I'm not sure how the casting would have worked 🤷

server/data/educationAndWorkPlanClient.test.ts
```diff
@@ -23,14 +24,45 @@ describe('educationAndWorkPlanClient', () => {
       // Given
       const prisonNumber = 'A1234BC'
       const systemToken = 'a-system-token'
-      const createGoalRequest = aValidCreateGoalRequestWithOneStep()
-      educationAndWorkPlanApi.post(`/action-plans/${prisonNumber}/goals`).reply(200, createGoalRequest)
+
```
nathanrussell-moj-digital, line 27: This test was slightly wrong, in that it was using the `createGoalRequest` as the response object for the mock API .... so I've fixed it 👍

server/data/educationAndWorkPlanClient.ts
```diff
@@ -7,17 +7,23 @@ export default class EducationAndWorkPlanClient {
     return new RestClient('Education and Work Plan API Client', config.apis.educationAndWorkPlan, token)
   }
 
-  async createGoal(createGoalRequest: CreateGoalRequest, token: string): Promise<void> {
-    await EducationAndWorkPlanClient.restClient(token).post({
+  async createGoal(createGoalRequest: CreateGoalRequest, token: string): Promise<unknown> {
```
nathanrussell-moj-digital, line 10: Much like the `curiousClient` above, I'm not sure how this worked before - not sure how a method declaring it returns a `Promise` works when it uses the `await` keyword ... 🤔

server/data/restClient.ts
```diff
@@ -1,4 +1,4 @@
-import superagent from 'superagent'
+import superagent, { Response, ResponseError } from 'superagent'
```
nathanrussell-moj-digital, line 1: A lot of refactoring in this class to be able to support `PUT` and `DELETE` .... but its mostly a copy and paste job from Send Legal Mail! 😁 👍

PR #86, opened 2023-08-02
initial template and routing logic for the update goal functionality

chrimesdev:
- Add the Update a goal views and functionality

### DIFF
--- a/server/config.ts
+++ b/server/config.ts
@@ -96,6 +96,6 @@ export default {
   domain: get('INGRESS_URL', 'http://localhost:3000', requiredInProduction),
   dpsHomeUrl: get('DPS_URL', 'http://localhost:3000/', requiredInProduction),
   featureToggles: {
-    editGoalsEnabled: Boolean(get('EDIT_GOALS_ENABLED', true)),
+    editGoalsEnabled: Boolean(get('EDIT_GOALS_ENABLED', false)),
   },
 }
--- a/server/config.ts
+++ b/server/config.ts
@@ -96,6 +96,6 @@ export default {
   domain: get('INGRESS_URL', 'http://localhost:3000', requiredInProduction),
   dpsHomeUrl: get('DPS_URL', 'http://localhost:3000/', requiredInProduction),
   featureToggles: {
-    editGoalsEnabled: Boolean(get('EDIT_GOALS_ENABLED', false)),
+    editGoalsEnabled: Boolean(get('EDIT_GOALS_ENABLED', true)),
   },
 }
--- a/server/views/pages/goal/update/index.njk
+++ b/server/views/pages/goal/update/index.njk
@@ -40,9 +40,7 @@
 
       <h1 class="govuk-heading-l">You are updating a goal for {{ prisonerSummary.firstName | capitalize }} {{ prisonerSummary.lastName | capitalize }}</h1>
 
-      <div data-module="moj-add-another">
         <form class="form" method="post" novalidate="">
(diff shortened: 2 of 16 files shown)


server/config.ts
```diff
@@ -96,6 +96,6 @@ export default {
   domain: get('INGRESS_URL', 'http://localhost:3000', requiredInProduction),
   dpsHomeUrl: get('DPS_URL', 'http://localhost:3000/', requiredInProduction),
   featureToggles: {
-    editGoalsEnabled: Boolean(get('EDIT_GOALS_ENABLED', false)),
+    editGoalsEnabled: Boolean(get('EDIT_GOALS_ENABLED', true)),
```
nathanrussell-moj-digital, line 99: Any reason why you changed this to `true` ?
I'd be tempted to leave it as `false`, so that the only way you can enable the feature is by setting the env var `EDIT_GOALS_ENABLED` to `true` in your `.env` file or helm deployment?

Unless theres some subtlety here that I've missed?

chrimesdev: Oops that was me accidentally checking in my lazy way of enabling the feature toggle locally. Will use `.env` in future!

server/views/pages/goal/update/index.njk
```diff
@@ -0,0 +1,161 @@
...
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+
+      <h1 class="govuk-heading-l">You are updating a goal for {{ prisonerSummary.firstName | capitalize }} {{ prisonerSummary.lastName | capitalize }}</h1>
+
+        <form class="form" method="post" novalidate="">
+            <input type="hidden" name="_csrf" value="{{ csrfToken }}" />
+            <input type="hidden" name="prisonNumber" value="{{ form.prisonNumber }}" />
+            <input type="hidden" name="goalReference" value="{{ form.goalReference }}" />
```
nathanrussell-moj-digital, line 46: I think we'll need status as a hidden field as well (its part of the form, so has a value and needs to be submitted, but at the moment we dont allow the user to change the status ... that will come later)
Something like
```
<input type="hidden" name="status" value="{{ form.status }}" />
```

server/views/pages/goal/update/index.njk
```diff
@@ -0,0 +1,161 @@
...
+              errorMessage: errors | findError('note')
+            }) }}
+
+            <h2 class="govuk-heading-m">Edit and remove steps</h2>
+
+            {% for step in form.steps %}
+              <input type="hidden" name="stepReference" value="{{ step.reference }}" />
+
+              {{ govukTextarea({
+                name: 'title-' + step.reference,
```
nathanrussell-moj-digital, line 77: I think the `name` and `id` attributes will need to reference the form structure, and in particular the fact that the `steps` field is an array
I think it'll need to be something like:
```
name: 'steps[' + loop.index + '].title',
id: 'steps[' + loop.index + '].title',
```
Something like that (assuming we can get the loop index?)

You'll need to do this on all the step fields, so that when the form gets `POST`ed back we get rehydrate the form data into the `UpdateGoalForm` object

chrimesdev: Just noticed that too from your PR https://github.com/ministryofjustice/hmpps-education-and-work-plan-ui/pull/87#discussion_r1283193701

Will update to match

nathanrussell-moj-digital (approved):
LGTM 👍 👏 
Lets merge it 👍

PR #100, opened 2023-08-09
RR-167: Check your updates page

chrimesdev:
When a user update goals, they will receive a check your updates page before continuing to save or going back to updating the goal.

### DIFF
--- a/integration_tests/e2e/goal/reviewUpdateGoal.cy.ts
+++ b/integration_tests/e2e/goal/reviewUpdateGoal.cy.ts
@@ -1,0 +1,99 @@
+import Page from '../../pages/page'
+import UpdateGoalPage from '../../pages/goal/UpdateGoalPage'
+import ReviewUpdateGoalPage from '../../pages/goal/ReviewUpdateGoalPage'
+import OverviewPage from '../../pages/overview/OverviewPage'
+import AuthorisationErrorPage from '../../pages/authorisationError'
+
+context('Review updated goal', () => {
+  beforeEach(() => {
+    cy.task('reset')
+    cy.task('stubSignInAsUserWithEditAuthority')
+    cy.task('stubAuthUser')
+    cy.task('getPrisonerById', 'G6115VJ')
+    cy.task('getActionPlan', 'G6115VJ')
+    cy.task('stubLearnerProfile')
+    cy.task('updateGoal')
+  })
+
+  it('should not be able to navigate directly to Review Goal given previous forms have not been submitted', () => {
+    // Given
+    const prisonNumber = 'G6115VJ'
+    const goalReference = '10efc562-be8f-4675-9283-9ede0c19dade'
+    cy.signIn()
+
+    // When
+    cy.visit(`/plan/${prisonNumber}/goals/${goalReference}/update/review`)
+
+    // Then
+    const overviewPage = Page.verifyOnPage(OverviewPage)
+    overviewPage.isForPrisoner(prisonNumber)
+  })
+
+  it('should navigate back to the update goals form if the go back to edit goal button is clicked', () => {
+    // Given
+    const prisonNumber = 'G6115VJ'
+    const goalReference = '10efc562-be8f-4675-9283-9ede0c19dade'
+    cy.signIn()
+
+    // When
+    cy.visit(`/plan/${prisonNumber}/goals/${goalReference}/update`)
+
+    // Then
+    const updateGoalPage = Page.verifyOnPage(UpdateGoalPage)
+    updateGoalPage.isForGoal(goalReference).submitPage()
+
+    // Then
+    const reviewUpdateGoalPage = Page.verifyOnPage(ReviewUpdateGoalPage)
+    reviewUpdateGoalPage.isForPrisoner(prisonNumber).goBackToEditGoal()
+
+    // Then
+    Page.verifyOnPage(UpdateGoalPage)
+  })
+
+  it('should be able to submit the form if no validation errors', () => {
+    // Given
+    const prisonNumber = 'G6115VJ'
+    const goalReference = '10efc562-be8f-4675-9283-9ede0c19dade'
+    cy.signIn()
+
+    // When
+    cy.visit(`/plan/${prisonNumber}/goals/${goalReference}/update`)
+
+    // Then
+    const updateGoalPage = Page.verifyOnPage(UpdateGoalPage)
+    updateGoalPage.isForGoal(goalReference).submitPage()
+
+    // Then
+    Page.verifyOnPage(OverviewPage)
+  })
+
+  it('should redirect to auth-error page given user does not have any authorities', () => {
+    // Given
+    cy.task('stubSignIn')
+
+    const prisonNumber = 'G6115VJ'
+    const goalReference = '10efc562-be8f-4675-9283-9ede0c19dade'
+    cy.signIn()
+
+    // When
+    cy.visit(`/plan/${prisonNumber}/goals/${goalReference}/update/review`, { failOnStatusCode: false })
+
+    // Then
+    Page.verifyOnPage(AuthorisationErrorPage)
+  })
+
+  it('should redirect to auth-error page given user does not have edit authority', () => {
+    // Given
+    cy.task('stubSignInAsUserWithViewAuthority')
+
+    const prisonNumber = 'G6115VJ'
+    const goalReference = '10efc562-be8f-4675-9283-9ede0c19dade'
+    cy.signIn()
+
+    // When
+    cy.visit(`/plan/${prisonNumber}/goals/${goalReference}/update/review`, { failOnStatusCode: false })
+
+    // Then
+    Page.verifyOnPage(AuthorisationErrorPage)
+  })
+})
--- a/integration_tests/e2e/goal/reviewUpdateGoal.cy.ts
+++ b/integration_tests/e2e/goal/reviewUpdateGoal.cy.ts
@@ -29,7 +29,7 @@ context('Review updated goal', () => {
     overviewPage.isForPrisoner(prisonNumber)
   })
 
-  it('should navigate back to the update goals form if the go back to edit goal button is clicked', () => {
+  it(`should navigate back to the update goal form when the 'go back to edit goal' button is clicked`, () => {
     // Given
     const prisonNumber = 'G6115VJ'
     const goalReference = '10efc562-be8f-4675-9283-9ede0c19dade'
@@ -63,9 +63,11 @@ context('Review updated goal', () => {
     const updateGoalPage = overviewPage.clickUpdateButtonForFirstGoal()
     updateGoalPage.isForGoal(goalReference).submitPage()
 
-    // Then
     const reviewUpdateGoalPage = Page.verifyOnPage(ReviewUpdateGoalPage)
-    reviewUpdateGoalPage.isForPrisoner(prisonNumber).submitPage()
+    reviewUpdateGoalPage.isForPrisoner(prisonNumber)
+
+    // Then
+    reviewUpdateGoalPage.submitPage()
 
     // Then
     Page.verifyOnPage(OverviewPage)
--- a/integration_tests/e2e/goal/reviewUpdateGoal.cy.ts
+++ b/integration_tests/e2e/goal/reviewUpdateGoal.cy.ts
@@ -36,10 +36,10 @@ context('Review updated goal', () => {
     cy.signIn()
 
     // When
-    cy.visit(`/plan/${prisonNumber}/goals/${goalReference}/update`)
+    cy.visit(`/plan/${prisonNumber}/view/overview`)
+    const overviewPage = Page.verifyOnPage(OverviewPage)
 
-    // Then
-    const updateGoalPage = Page.verifyOnPage(UpdateGoalPage)
+    const updateGoalPage = overviewPage.clickUpdateButtonForFirstGoal()
     updateGoalPage.isForGoal(goalReference).submitPage()
 
     // Then
@@ -57,11 +57,15 @@ context('Review updated goal', () => {
     cy.signIn()
 
     // When
-    cy.visit(`/plan/${prisonNumber}/goals/${goalReference}/update`)
+    cy.visit(`/plan/${prisonNumber}/view/overview`)
+    const overviewPage = Page.verifyOnPage(OverviewPage)
+
+    const updateGoalPage = overviewPage.clickUpdateButtonForFirstGoal()
+    updateGoalPage.isForGoal(goalReference).submitPage()
 
     // Then
-    const updateGoalPage = Page.verifyOnPage(UpdateGoalPage)
-    updateGoalPage.isForGoal(goalReference).submitPage()
+    const reviewUpdateGoalPage = Page.verifyOnPage(ReviewUpdateGoalPage)
+    reviewUpdateGoalPage.isForPrisoner(prisonNumber).submitPage()
 
     // Then
     Page.verifyOnPage(OverviewPage)
--- a/integration_tests/e2e/goal/reviewUpdateGoal.cy.ts
+++ b/integration_tests/e2e/goal/reviewUpdateGoal.cy.ts
@@ -35,16 +35,16 @@ context('Review updated goal', () => {
     const goalReference = '10efc562-be8f-4675-9283-9ede0c19dade'
     cy.signIn()
 
-    // When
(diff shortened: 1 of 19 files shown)


integration_tests/pages/goal/ReviewUpdateGoalPage.ts
```diff
@@ -0,0 +1,34 @@
+import Page, { PageElement } from '../page'
+
+export default class UpdateGoalPage extends Page {
```
nathanrussell-moj-digital, line 3: Minor; but I think this class name should be `ReviewUpdateGoalPage` to match the filename

server/routes/updateGoal/reviewView.ts
```diff
@@ -0,0 +1,16 @@
+import type { PrisonerSummary } from 'viewModels'
+import type { UpdateGoalForm } from 'forms'
+
+export default class ReviewView {
```
nathanrussell-moj-digital, line 4: I'm thinking the name `ReviewView` is a bit vague / nondescript (even though it is in a folder structure which arguably gives it some context) ... 🤔 

What about something like `reviewUpdateGoalView.ts` where the classname is `ReviewUpdateGoalView` ?

What do you think? 🤔

server/routes/updateGoal/index.ts
```diff
@@ -26,4 +27,21 @@ export default (router: Router, services: Services) => {
...
   ])
+
+  router.use('/plan/:prisonNumber/goals/:goalReference/update/review', [
+    checkUserHasEditAuthority(),
+    prisonerSummaryRequestHandler.getPrisonerSummary,
+  ])
+
+  router.get('/plan/:prisonNumber/goals/:goalReference/update/review', [
+    checkPrisonerSummaryExistsInSession,
+    checkUpdateGoalFormExistsInSession,
```
nathanrussell-moj-digital, line 38: Good work finding our how our route request handlers work 👍 💯 

What you've done works, and is fine, but there is an improvement I might suggest (which might help solidify how and why some of this stuff works)

`router.use` is like a 'common' handler/hook for the given path that are then referenced in the `router.get` / `router.post` etc calls. It means we can put common code there and not have to repeat it in the `router.get` / `router.post` etc calls.
In `router.use` we want to check the user has the correct authority 👍 , but we can also use it to check the user has the Update Goal Form in the session. It therefore saves doing it in both the `router.get` / `router.post` calls.

And we can improve it even more ... 👍 ....
The call `prisonerSummaryRequestHandler.getPrisonerSummary` in the current `router.use` goes off and puts the prisoner summary in the session if it's not already there. Putting it in a `router.use` makes sense .... but only when its the start of the journey.
Our new page is the end of the journey (its the check your answers page), so we dont want to mess about getting the prisoner summary and putting it in the session for _this_ route - we want to make sure its already there. If not it means someone is trying to get to this page out of sequence.
So, you can remove the call to get the prisoner summary (line 34), and replace it with a check to see that it already exists.

A bit like this: ....

```
  router.use('/plan/:prisonNumber/goals/:goalReference/update/review', [
    checkUserHasEditAuthority(),
    checkPrisonerSummaryExistsInSession,
    checkUpdateGoalFormExistsInSession,
  ])

  router.get('/plan/:prisonNumber/goals/:goalReference/update/review', [
    updateGoalController.getReviewUpdateGoalView,
  ])

  router.post('/plan/:prisonNumber/goals/:goalReference/update/review', [
    updateGoalController.submitReviewUpdateGoal,
  ])
```

(Based on the explanation above of what `prisonerSummaryRequestHandler.getPrisonerSummary` and `checkPrisonerSummaryExistsInSession` do and why/where you would use them; you could remove lines 22 and 27 altogether. I know its not part of your change, but if you want to push the broom around whilst you are in the general vicinity that would be cool 👍 . Tidy tidy 👍 )

server/routes/updateGoal/updateGoalController.ts
```diff
@@ -60,14 +61,29 @@ export default class UpdateGoalController {
...
+  }
+
+  submitReviewUpdateGoal: RequestHandler = async (req, res, next): Promise<void> => {
+    const { prisonNumber } = req.params
+    const { updateGoalForm } = req.session
+
     const updateGoalDto = toUpdateGoalDto(updateGoalForm)
     await this.educationAndWorkPlanService.updateGoal(prisonNumber, updateGoalDto, req.user.token)
-    // TODO - RR-188 - handle API error response when updating goal
+
```
nathanrussell-moj-digital, line 85: Can you add the todo comment in please? Otherwise we will forget this is one of the areas that we need to come back to when we do RR-188
```
// TODO - RR-188 - handle API error response when updating goal
```

server/routes/updateGoal/updateGoalController.ts
```diff
@@ -60,14 +61,29 @@ export default class UpdateGoalController {
...
+    req.session.updateGoalForm = undefined
```
nathanrussell-moj-digital, line 86: 💯 Yes, good work remembering / sussing out that we need to clear the form from the session once we are done 👍 

Not only is it a clean and tidy thing to do (we only want to hold things in the session for as long as we need them, otherwise the session could get massive); but also it could cause weird bugs. For example, if we've left an UpdateGoalForm in the session; then when someone next hits the Update Goal page (either by clicking Update from the Overview page, or perhaps using browser back navigation, or bookmarking URLs etc); then your new `checkUpdateGoalFormExistsInSession` request handler will kick in and will pass! Which might result in showing/editing the wrong Goal !!

server/views/pages/goal/update/review.njk
```diff
@@ -0,0 +1,131 @@
...
+{% endblock %}
+
+{% block content %}
+
+  {% include "../../../partials/prisonerBanner.njk" %}
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-full">
+
+      <h1 class="govuk-heading-l">Check your updates for {{ prisonerSummary.firstName | capitalize }} {{ prisonerSummary.lastName | capitalize }}'s goals</h1>
```
nathanrussell-moj-digital, line 33: Minor - I think it should be "goal" (singular) ?

server/views/pages/goal/update/review.njk
```diff
@@ -0,0 +1,131 @@
...
+
+      <h2 class="govuk-heading-m govuk-visually-hidden">Goal details</h2>
+
+      {% if data.status == "COMPLETE" %}
```
nathanrussell-moj-digital, line 37: This is not quite right .... unfortunately we have 2 types of status we are dealing with here - Goal Status and Step Status .... and unfortunately they have different values! 😢 

The field you are dealing with in _this_ part of the template is the Goal Status - and it has these values:
```
'ACTIVE' | 'COMPLETED' | 'ARCHIVED'
```
And just to make it more confusing, the terminology they like to use for an Active goal is "In progress" ... not "Started" ! 
We will need to find out the complete set of corresponding screen labels/terminology (I dont know them off hand)

---

Steps have their own status enum, which is subtly different! It has these values:
```
'NOT_STARTED' | 'ACTIVE' | 'COMPLETE'
```
Where the corresponding screen labels/terminology are "Not started", "Started", "Completed"

---

Rather than putting this logic in the template as you have here for Goal Status; and further down for Step Status on line 90 - can you please write a couple of filters for these please? One for Goal Status, and one for Step Status.
Have a look at `formatDateFormValue` and it's unit test as an example.

integration_tests/e2e/goal/updateGoal.cy.ts
```diff
@@ -90,6 +91,10 @@ context('Update a goal', () => {
       .setStepTargetDateRange(3, '6 to 12 months')
       .submitPage()
 
+    // Then
```
nathanrussell-moj-digital, line 94: Minor, but we dont want 2 `// Then` blocks, plus I think the code in your new `Then` block is actually part of the `When` 
My thinking is that in context of the test (adding a step as part of adding a goal) we need to go via the new Review Goal page in order to actually save the Goal (and its new step) - so it feels like its part of the `When` ?

How about this?
```
    // When
    updateGoalPage //
      // The goal loaded from the wiremock stub data has 2 steps. After adding another step, the new step will be referenced as step 3
      .addAnotherStep()
      .setStepTitle(3, 'A brand new step')
      .setStepStatus(3, 'Started')
      .setStepTargetDateRange(3, '6 to 12 months')
      .submitPage()
    const reviewUpdateGoalPage = Page.verifyOnPage(ReviewUpdateGoalPage)
    reviewUpdateGoalPage.submitPage()

    // Then
    Page.verifyOnPage(OverviewPage)
    // Assert that the expected ...
```

server/routes/updateGoal/updateGoalController.test.ts
```diff
@@ -136,7 +136,7 @@ describe('updateGoalController', () => {
   })
 
   describe('submitUpdateGoalForm', () => {
-    it('should update goal and redirect to overview page given action is submit-form and validation passes', async () => {
+    it('should update goal and redirect to review updated goal page given action is submit-form and validation passes should redirect to the overview page', async () => {
```
nathanrussell-moj-digital, line 139: Minor (probably copy and paste typo 😁 ) - but we don't need the extra `should redirect to the overview page` on the end of the function name)

server/routes/updateGoal/updateGoalController.test.ts
```diff
@@ -157,6 +157,18 @@ describe('updateGoalController', () => {
         next as undefined as NextFunction,
       )
 
+      // Then
```
nathanrussell-moj-digital, line 160: I think this new `Then` block is in the wrong place (and might be masking a failing test - I'm not sure how this test passes as it currently stands ... 🤔 )

So, our Given When Then comment blocks are there to help us structure our tests and to aid readability etc. The basic idea is this:
* Given "a load of setup / preconditions / etc"
* When "I do one and only one thing (ie. the action under test)"
* Then "I make some assertions about what should have happened"

The idea is that we have 1 and only 1 of each comment block, and that they are in the correct sequence (G, W, T) 

So, your new assertion (expect that we've redirected to the review page) is the right thing to assert.
But we're asserting it before we've done the When .... which feels kinda weird. Other than being in the wrong place, I have idea how this passes, because surely jest should fail this expectation because we've not called `res.redirect` yet because we've not called the `submitReviewUpdateGoal` ..... oh, hang on, I've seen it!

Oh - you have 2 sets of When Then .... you have `Given, When, Then, When, Then` ! Ah ha!

We need to refactor this test because it's doing 2 things (2 `When` blocks)

Lets pair on this one later and I'll help you with it 👍

integration_tests/e2e/goal/reviewUpdateGoal.cy.ts
```diff
@@ -0,0 +1,103 @@
...
+    overviewPage.isForPrisoner(prisonNumber)
+  })
+
+  it('should navigate back to the update goals form if the go back to edit goal button is clicked', () => {
+    // Given
+    const prisonNumber = 'G6115VJ'
+    const goalReference = '10efc562-be8f-4675-9283-9ede0c19dade'
+    cy.signIn()
+
+    // When
```
nathanrussell-moj-digital, line 38: I think the scenarios you are testing in these cypress tests are good 👍 ; its just that the Given, When, Then's are very slightly confused.

The `When` is the thing you do in order to cause the behaviour and ideally should be a single function call (though in some cases thats not always possible)

In these tests, the key action is clicking the back button on the Update Goal Review page Everything before the actual click of the button (ie. how we got to the page) is all setup/pre-amble, and so lives in the `Given` block

Everything after the submit button has been clicked are your assertions so all live under a single `Then` comment block.

Like this:
```
    // Given
    const prisonNumber = 'G6115VJ'
    const goalReference = '10efc562-be8f-4675-9283-9ede0c19dade'
    cy.signIn()

    cy.visit(`/plan/${prisonNumber}/view/overview`)
    const updateGoalPage = overviewPage.clickUpdateButtonForFirstGoal()
    updateGoalPage.isForGoal(goalReference).submitPage()

    const reviewUpdateGoalPage = Page.verifyOnPage(ReviewUpdateGoalPage)
    reviewUpdateGoalPage.isForPrisoner(prisonNumber).goBackToEditGoal()

    // When
    reviewUpdateGoalPage.goBackToEditGoal()

    // Then
    Page.verifyOnPage(UpdateGoalPage).isForGoal(goalReference)
```

If you can apply that approach to the other scenarios in this test that would be great 👍 
Shout if you need help

integration_tests/pages/goal/ReviewUpdateGoalPage.ts
chrimesdev: Updated to `ReviewUpdateGoalPage`

server/routes/updateGoal/updateGoalController.ts
chrimesdev: Added the comment back in!

server/routes/updateGoal/updateGoalController.test.ts
chrimesdev: Fixed this duplicated typo

server/routes/updateGoal/index.ts
chrimesdev: Removed the duplication here for both `/update/review` and `/update`

server/routes/updateGoal/updateGoalController.ts
chrimesdev: Can't take credit for this one! Think its something you did for Update a goal

server/views/pages/goal/update/review.njk
chrimesdev: Think there is a little bit of inconsistency through the app (mainly Overview and this page) with the labeling/wording - will double check the designs and implement filters

chrimesdev: Confirmed the labels with Jo and created a filter for both Goal status and Step status 👍🏻 .

integration_tests/e2e/goal/updateGoal.cy.ts
chrimesdev: Removed the duplicated `// Then` blocks

integration_tests/e2e/goal/reviewUpdateGoal.cy.ts
chrimesdev: Removed the duplicated `// Then` blocks

server/routes/updateGoal/index.ts
chrimesdev: Hmm this change has broken the test on the pipeline but its still passing locally for me 🤔

chrimesdev: Something strange happening with the tests here, when you move `checkPrisonerSummaryExistsInSession` to the `.use` for the Update view, the test `should be able to navigate directly to update goal page` fails 🤔

chrimesdev: Works fine for the Review view

server/utils/nunjucksSetup.ts
```diff
@@ -53,6 +55,8 @@ export function registerNunjucks(app?: express.Express): Environment {
   njkEnv.addFilter('formatDate', formatDateFilter)
   njkEnv.addFilter('formatDateFormValue', formatDateFormValue)
   njkEnv.addFilter('formatFunctionalSkillType', formatFunctionalSkillType)
+  njkEnv.addFilter('formatStepStatusValue', formatStepStatusValue)
```
nathanrussell-moj-digital, line 58: 👍

server/routes/updateGoal/index.ts
nathanrussell-moj-digital: I've checked out your branch and made the edits I was suggesting to the router. Not sure what tests you had that were failing, but they are all passing now; so happy days 👍

server/routes/updateGoal/updateGoalController.test.ts
nathanrussell-moj-digital: Ah ha; I've sussed it - its actually 2 separate tests! That's often the clue - if there are 2 `Given`'s, it usually means its actually 2 tests. And that was the case here.
The first test was the one you were refactoring - the test of the `submitUpdateGoalForm` function - we simply needed to change that to not assert the service method was called to actually update the goal; and to change the page it redirects to. Because thats what the `submitUpdateGoalForm` does now 👍 

And then there is a new test for the new `submitReviewUpdateGoal` method. Bingo! That makes more sense 👍 

I've refactored your tests to reflect the above 👍

nathanrussell-moj-digital (approved):
LGTM 👍

PR #101, opened 2023-08-10
RR-140 - Initial Curious API client function and service to get prisoner education records

nathanrussell-moj-digital:
This PR adds the Curious API client function to get prisoner education records (`LearnerEducation` in their model); and a service method to call it, collating the results and mapping into `PrisonerEducation` instances (our view model)

There's a little complexity here in that the Curious API is a paged API. It returns "pages" of data, rather than the entire lot in one go. Therefore the service class needs to check the response to see if its the last page, and if not call the API again for the next page of data.

The other thing that needs working out is what our view model (`PrisonerEducation`) actually needs to contain. At the moment I've just given it a few fields from the monster set of data we get from Curious. Once we understand what our screen requirements are and how that relates to the Curious data we can add those fields to our `PrisonerEducation` class (and possibly change its name)


server/data/restClient.ts
```diff
@@ -64,17 +65,32 @@ export default class RestClient {
     return this.config.timeout
   }
 
-  private logRequest(method: string, path: string, query = '') {
+  private logRequest(method: string, path: string, query = {}) {
     if (this.token) {
-      logger.info(`${method.toUpperCase()} request using user credentials: calling ${this.name}: ${path} ${query}`)
+      const queryString = this.queryString(query)
+      logger.info(`${method.toUpperCase()} request using user credentials: calling ${this.name}: ${path}${queryString}`)
```
nathanrussell-moj-digital, line 71: The queryString wasnt logged properly before (it was being logged as `[Object Object]`)

server/data/restClient.ts
```diff
@@ -10,7 +10,7 @@ import { restClientMetricsMiddleware } from './restClientMetricsMiddleware'
 
 interface GetRequest {
   path?: string
-  query?: string
+  query?: Record<string, string>
```
nathanrussell-moj-digital, line 13: The query string params need to be a map of name value pairs; rather than a string

server/services/curiousService.test.ts
```diff
@@ -30,13 +43,13 @@ describe('curiousService', () => {
       const username = 'a-dps-user'
 
       const systemToken = 'a-system-token'
-      hmppsAuthClient.getSystemClientToken.mockImplementation(() => Promise.resolve(systemToken))
+      hmppsAuthClient.getSystemClientToken.mockResolvedValue(systemToken)
```
nathanrussell-moj-digital, line 46: Not really related to the Curious API and getting education data; but `mockResolved` and `mockRejected` is a better/cleaner approach than `mockImplementation`

server/services/curiousService.ts
```diff
@@ -30,11 +36,44 @@ export default class CuriousService {
...
+    const systemToken = await this.hmppsAuthClient.getSystemClientToken(username)
+
+    try {
+      let page = 0
+      let apiPagedResponse = { last: false } as LearnerEductionPagedResponse
+      const apiLearnerEducation: Array<LearnerEducation> = []
+
+      // loop until the API response's `last` field is `true`
+      while (apiPagedResponse.last === false) {
+        // eslint-disable-next-line no-await-in-loop
```
nathanrussell-moj-digital, line 61: `eslint` frowns upon using an `await` inside a loop because it says it is inefficient.
It's rationale is that you instead you return all your promises in a `Promise.all` ... which does indeed make sense .... if your loop is not controlled/dependent on the result of each promise.

Our loop _is_ controlled by the result of each promise - we need the result of the promise (ie. the API call to Curious to get the page of data) in order to decide whether its the last page or not, and whether to keep looping.

It basically makes our API call synchronous rather than async, which is the main objection to using `await` in a loop.
But because we need to know the result of the previous API call, we are implicitly synchronous 🤷‍♂️

chrimesdev (approved):

PR #103, opened 2023-08-11
404 error handling with a 404 page template and content

chrimesdev:
https://dsdmoj.atlassian.net/browse/RR-220

### DIFF
--- a/server/errorHandler.test.ts
+++ b/server/errorHandler.test.ts
@@ -13,25 +13,14 @@ afterEach(() => {
 })
 
 describe('GET 404', () => {
-  it('should render content with stack in dev mode', () => {
+  it('should render a 404 page with a link to the homepage when page is not found', () => {
     return request(app)
       .get('/unknown')
       .expect(404)
       .expect('Content-Type', /html/)
       .expect(res => {
-        expect(res.text).toContain('NotFoundError: Not found')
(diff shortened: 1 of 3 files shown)


nathanrussell-moj-digital (approved):
LGTM 👍 

Lets get this merged; then in another branch and PR can you write a couple of cypress tests please?
I think we want:
* User is logged in and has tried to visit a page that does not exist. Expect the 404 page to be displayed
* User is not logged in and has tried to visit a page that does not exist. Expect the auth-error page to be displayed (403)

You will need to add the pageId to the 404.njk page you just created, but should be easy enough to do

The functionality that drives both behaviour paths above already exists (or will done once you merge this), so it should just be a case of writing the cypress tests
(see here for an example, though we dont have the concept of different types of users (mailroom and legal sender users) - https://github.com/ministryofjustice/send-legal-mail-to-prisons/blob/main/integration_tests/integration/page-not-found.cy.ts ; also https://github.com/ministryofjustice/send-legal-mail-to-prisons/blob/main/integration_tests/pages/error404.ts )

nathanrussell-moj-digital (approved):

chrimesdev:
Added the `id` in this one as it felt relevant to the template changes, will add the Cypress tests in a new PR!

PR #105, opened 2023-08-11
add custom column widths to fix the styling of steps tables

chrimesdev:
Not usually a fan of adding custom CSS, but in this case it is recommended by GOVUK Design System

> If the [width override classes](https://design-system.service.gov.uk/styles/layout/#width-override-classes) do not meet your needs you can create your own width classes and apply them to the cells in the table head. These can be added using the classes option in the Nunjucks macro or adding the class directly to the individual cells within govuk-table__head as below.

Source: https://design-system.service.gov.uk/components/table/#custom-column-widths

https://dsdmoj.atlassian.net/browse/RR-219

### DIFF
--- a/assets/scss/local.scss
+++ b/assets/scss/local.scss
@@ -1,3 +1,14 @@
 .govuk-main-wrapper {
   min-height: 600px;
 }
+
+// Custom column widths
+// https://design-system.service.gov.uk/components/table/#custom-column-widths
+
+.app-u-width-20 {
+  width: 20% !important;
+}
+
+.app-u-width-60 {
+  width: 60% !important;
+}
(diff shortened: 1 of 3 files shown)


nathanrussell-moj-digital (approved):
LGTM 👍 

1 minor question - what does the `-u-` mean in the css class names? Is it some kind of convention? - either a web industry convention or a GDS convention ?

chrimesdev:
> 

It's from a CSS architecture called ITCSS (Inverted Triangle CSS), utilities – utilities and helper classes with ability to override anything which goes before in the triangle, e.g. hide helper class. The common pattern is to use `-u-` to indicate its a utility class and its the only time you should be using `!important` 

Good article on ITCSS here if you're interested: https://www.xfive.co/blog/itcss-scalable-maintainable-css-architecture

chrimesdev:
Also any custom styles should be prefixed with `app-` so its easy to see that the styles are coming from the app, rather than GOVUK styles prefixed with `govuk-`

PR #109, opened 2023-08-14
RR-140 - Add qualification grade

nathanrussell-moj-digital:
This PR maps the grade from the Curious eduction record to our view model and renders it on screen.

The rules / Curious scheme is unclear in this respect. Looking at the (scant few) example responses we have been given, I've gone with the approach that the "grade" is the `LearnerEducationDTO`s `outcomeGrade` field, else the `outcome` field, else `null`

I'm sure this question of how we've mapped the data will come out in testing, but in the absence of any clear direction at the moment this is OK for now.

Screenshot 2023-08-14 at 13 03 04


server/@types/viewModels/index.d.ts
```diff
@@ -81,7 +81,7 @@ declare module 'viewModels' {
   }
 
   /**
-   * A prisoner's record of In Prison courses and eduction, which is made up of a collection of InPrisonEducation.
+   * A prisoner's record of In-Prison courses and eduction, which is made up of a collection of InPrisonEducation.
    * TODO - needs better name and field definition once we understand the data and the requirements.
```
nathanrussell-moj-digital, line 85: @mattwills-moj-digital @chrimesdev - What do we think of the names `InPrisonEducationRecords` and `InPrisonEducation` as type names in our view model?
If we are happy with the names I'll get rid of the TODOs in this PR; else do we have any better name suggestions?; else I'll leave it for another day
What do you think?

chrimesdev: Those names seem OK to me, without looking at the code and seeing the names first - I had an good idea of what to expect from them!

server/views/pages/overview/partials/educationAndTrainingFunctionalSkills.test.ts
```diff
@@ -0,0 +1,125 @@
+import * as fs from 'fs'
```
nathanrussell-moj-digital, line 1: This is not really a new test; its a rename of the previously committed `server/views/pages/overview/partials/educationAndTrainingTabContents.test.ts` (which is shown below as a deletion)

server/views/pages/overview/partials/supportNeedsTabContents.test.ts
```diff
@@ -1,4 +1,4 @@
-import fs from 'fs'
+import * as fs from 'fs'
```
nathanrussell-moj-digital, line 1: Minor tidy of this import, prevents an IDE warning

server/@types/viewModels/index.d.ts
mattwills-moj-digital: I think it's ok too. An alternative might be `InPrisonEducation` which consists of a collection of `InPrisonCourse`, but I'm not sure if that's any better? I'll let you choose 😄

server/@types/viewModels/index.d.ts
```diff
@@ -81,7 +81,7 @@ declare module 'viewModels' {
   }
 
   /**
-   * A prisoner's record of In Prison courses and eduction, which is made up of a collection of InPrisonEducation.
+   * A prisoner's record of In-Prison courses and eduction, which is made up of a collection of InPrisonEducation.
```
mattwills-moj-digital, line 84: Minor - typo in 'eduction'

server/data/mappers/inPrisonEducationMapper.ts
```diff
@@ -13,6 +13,7 @@ const toInPrisonEducation = (apiLearnerEducation: LearnerEducation): InPrisonEdu
     courseCompletionDate: apiLearnerEducation.learningActualEndDate
       ? moment(apiLearnerEducation.learningActualEndDate).toDate()
       : null,
+    grade: apiLearnerEducation.outcomeGrade || apiLearnerEducation.outcome || null,
```
mattwills-moj-digital, line 16: Makes sense, at least for now 👍

mattwills-moj-digital (dismissed):
LGTM. Just a couple of comments 👍

server/@types/viewModels/index.d.ts
nathanrussell-moj-digital: I'll leave it for now; but will remove the TODOs
(mainly because changing the name would mean changing all the test data builders, variables that use test data instances etc etc)
😁

server/@types/viewModels/index.d.ts
nathanrussell-moj-digital: Good spot - I'll fix that

server/@types/viewModels/index.d.ts
```diff
@@ -101,6 +101,7 @@ declare module 'viewModels' {
     courseStartDate: Date
     courseCompleted: boolean
     courseCompletionDate?: Date
+    grade?: string
     source: 'CURIOUS'
```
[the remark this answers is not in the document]
nathanrussell-moj-digital: Its always Curious at the moment.
As I understand it, its not captured as part of the CIAG induction. I believe they might display it from Curious, and they _might_ save it in their database (I dont know that for sure), but as far as I know its mastered/owned in Curious.

In respect of our integration we are reading it from Curious.

We have future stories (yet to be written or refined etc yet) where we allow adding and updating qualifications. We know that as far as our project is concerned Curious is read-only, so when we add/edit qualifications it must be in some other API/database (probably ours, but yet to be defined)
The point is, when we get to that point, an In Prison qualification could have been read from Curious, or it could have be read from another API. Hence the `source` field

mattwills-moj-digital (approved):

server/@types/viewModels/index.d.ts
mattwills-moj-digital: No problem, I might change it myself! 😂

PR #111, opened 2023-08-14
add the nunjucks template for the qualifications and achievement sidebar

chrimesdev:


server/views/pages/overview/partials/functionalSkillsSidebar.njk
```diff
@@ -17,21 +17,22 @@
           {% for skill in functionalSkills.assessments %}
           <tr class="govuk-table__row">
             <th scope="row" class="govuk-table__header">
-              {{ skill.type | formatFunctionalSkillType }}
               {% if skill.assessmentDate %}
-              <br>
               <span class="govuk-body-s">Assessed on {{ skill.assessmentDate | formatDate('D MMMM YYYY') }}</span>
+              <br>
```
nathanrussell-moj-digital, line 22: Did you mean to change this markup?
Wasn't it correct before?

chrimesdev: Seems like the date keeps getting moved above and below on the designs, not much consistency in the designs atm  🤔 ... I followed the screens in RR-179 and made the date appear above on both sidebars, as the screens in the ticket will be what it gets tested against 🤷🏻‍♂️

nathanrussell-moj-digital: Oh, I see what you mean.
This whole problem of the prototype being a moving target is doing my head in! And the fact that we now apparently have 2 different designs in the same prototype! 

This is precisely why I've tried to assert the position that we only develop from what is in the ticket, and nothing else. If its wrong in the ticket then they can raise a new ticket to get it corrected (though in this case I know its a bit weird in that we had a ticket recently to change the Functional Skills box, which then put the screenshots in this ticket out of date!)

I'll raise it in tomorrow's DSU and will take it from there

nathanrussell-moj-digital (approved):

PR #116, opened 2023-08-16
RR-204 - Removed EDIT_GOALS_ENABLED feature toggle

nathanrussell-moj-digital:
This PR removes the EDIT_GOALS_ENABLED feature toggle.
It was only ever a temporary feature toggle, to prevent us delivering half-baked code whilst we did the Update Goal stories
Now those stories are delivered and have been tested successfully, we can remove the feature toggle 👍

### DIFF
--- a/server/config.ts
+++ b/server/config.ts
@@ -96,6 +96,6 @@ export default {
   domain: get('INGRESS_URL', 'http://localhost:3000', requiredInProduction),
   dpsHomeUrl: get('DPS_URL', 'http://localhost:3000/', requiredInProduction),
   featureToggles: {
-    editGoalsEnabled: Boolean(get('EDIT_GOALS_ENABLED', false)),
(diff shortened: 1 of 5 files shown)


server/config.ts
```diff
@@ -96,6 +96,6 @@ export default {
   domain: get('INGRESS_URL', 'http://localhost:3000', requiredInProduction),
   dpsHomeUrl: get('DPS_URL', 'http://localhost:3000/', requiredInProduction),
   featureToggles: {
-    editGoalsEnabled: Boolean(get('EDIT_GOALS_ENABLED', false)),
+    // someToggleEnabled: Boolean(get('SOME_TOGGLE_ENABLED', false)),
```
nathanrussell-moj-digital, line 99: I left this as an example for future use, rather than an empty `featureToggles` object which might not have been so obvious how to use in the future.

mattwills-moj-digital: Nice idea 👍

PR #117, opened 2023-08-18
RR-115 - Imported swagger spec; added basic client class

nathanrussell-moj-digital:
This PR imports the CIAG API swagger spec, sets up the config and env vars, and creates the basic client class and method.

There are no unit tests at this point as we need to understand a little more about the CIAG API response data so that we can write test data builders etc to support the tests.


mattwills-moj-digital (approved):
GS

PR #120, opened 2023-08-21
RR-115 - Basic CIAG Induction Service, with error handling

nathanrussell-moj-digital:
This PR adds the CIAG Induction Service class, with a service method to get the CIAG Induction.
The returned CIAGInduction is mapped to a WorkAndInterests object (detail still to be fleshed out)

The view has been updated to show an error if the CIAG API is down altogether (our usual "some other service is down" type message). 

We also handle the scenario where the CIAG API is up, but there is no CIAG Induction for the prisoner. The content team have specified the message they wish to show in the case (though it might need some 'flourish' 😁 )
(Note also that this is basically a 404 scenario. I've coded it based around a 404, but we know that at the moment the CIAG API returns this as a 400. I will ask them to correct this in their API)

### DIFF
--- a/server/@types/viewModels/index.d.ts
+++ b/server/@types/viewModels/index.d.ts
@@ -9,7 +9,7 @@ declare module 'viewModels' {
   }
 
   export interface PrisonerSupportNeeds {
-    problemRetrievingData?: boolean
+    problemRetrievingData: boolean
     healthAndSupportNeeds: Array<HealthAndSupportNeeds>
     neurodiversities: Array<Neurodiversity>
   }
@@ -36,7 +36,7 @@ declare module 'viewModels' {
    * A prisoner's Functional Skills, which is made up of a collection of Assessments.
    */
   export interface FunctionalSkills {
-    problemRetrievingData?: boolean
+    problemRetrievingData: boolean
     assessments: Array<Assessment>
   }
 
@@ -84,7 +84,7 @@ declare module 'viewModels' {
    * A prisoner's record of In-Prison courses and education, which is made up of a collection of InPrisonEducation.
    */
   export interface InPrisonEducationRecords {
-    problemRetrievingData?: boolean
+    problemRetrievingData: boolean
     educationRecords: Array<InPrisonEducation>
   }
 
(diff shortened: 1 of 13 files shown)


server/@types/viewModels/index.d.ts
```diff
@@ -9,7 +9,7 @@ declare module 'viewModels' {
   }
 
   export interface PrisonerSupportNeeds {
-    problemRetrievingData?: boolean
```
nathanrussell-moj-digital, line 12: These `problemRetrievingData` fields don't need to be optional - there are no cases where we don't set them (nor indeed should there be I don't think)

server/@types/viewModels/index.d.ts
```diff
@@ -102,4 +102,30 @@ declare module 'viewModels' {
     grade?: string
     source: 'CURIOUS'
   }
+
+  /**
+   * A prisoner's pre-prison work experience, skills, and future work interests.
+   */
+  export interface WorkAndInterests {
+    problemRetrievingData: boolean
+    data?: WorkAndInterestsData
```
nathanrussell-moj-digital, line 111: This `data` field is optional, because its how I'm proposing we know that we called the CIAG API correctly, but there is no data (ie. a 404)

server/data/ciagInductionClient.ts
```diff
@@ -7,10 +7,10 @@ export default class CiagInductionClient {
     return new RestClient('CIAG Induction API Client', config.apis.ciagInduction, token)
   }
 
-  async getCiagInduction(prisonNumber: string, token: string): Promise<Array<CiagInduction>> {
-    const learnerProfiles = CiagInductionClient.restClient(token).get({
+  async getCiagInduction(prisonNumber: string, token: string): Promise<CiagInduction> {
```
nathanrussell-moj-digital, line 10: Whoops! Spot the blatant previous copy and paste from another class! 🤣

server/data/mappers/workAndInterestMapper.test.ts
```diff
@@ -0,0 +1,21 @@
...
+      problemRetrievingData: false,
+      data: undefined,
+    }
+
+    // When
+    const actual = toWorkAndInterests(ciagInduction)
+
+    // Then
+    expect(actual).toEqual(expected)
+  })
```
nathanrussell-moj-digital, line 20: I will add other tests once the rest of the mapper is fleshed out

mattwills-moj-digital (approved):
LGTM

PR #124, opened 2023-08-22
RR-115 - Define our WorkExperience view model object

nathanrussell-moj-digital:
This PR define our WorkExperience view model object; and maps into it from the CIAG Induction data

This type definition (WorkExperience) and its mapping is one third of the data required for the Work and Interests screen. In total we need Work Experience, Work Interests and Skills and Interests.

The others will be done in my next PRs, but I felt that seeing / reviewing these in smaller chunks should help visualize the data a little 🤞 

Bear in mind also that this mapping and the test data builders are pure guesswork at the moment because we are yet to see any actual data the the CIAG Induction API

### DIFF
--- a/server/testsupport/ciagInductionTestDataBuilder.ts
+++ b/server/testsupport/ciagInductionTestDataBuilder.ts
@@ -1,0 +1,60 @@
+import type { CiagInduction } from 'ciagInductionApiClient'
+
+const aCiagInductionWithNoRecordOfAnyPreviousWorkExperience = (prisonNumber = 'A1234BC'): CiagInduction => {
+  return {
+    ...baseCiagInductionUsedForBuildingOtherInstances(prisonNumber),
+    workExperience: null,
+  }
+}
+
+const aCiagInductionWithNoPreviousWorkExperience = (prisonNumber = 'A1234BC'): CiagInduction => {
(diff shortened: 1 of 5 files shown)


server/testsupport/ciagInductionTestDataBuilder.ts
```diff
@@ -0,0 +1,60 @@
...
+          otherWork: 'Retail delivery',
+          role: 'Milkman',
+          details: 'Self employed franchise operator delivering milk and associated diary products.',
+        },
+      ],
+    },
+  }
+}
+
+const baseCiagInductionUsedForBuildingOtherInstances = (prisonNumber = 'A1234BC'): CiagInduction => {
```
mattwills-moj-digital, line 46: Not sure about the name - perhaps it could have `template` in it somewhere?

I think it would be useful to have a method called something like `aCiagInductionWithMinimumData()`, but we don't know what that looks like yet!

Anyway, ignore for this now. I know you're keen to get it merged, so I'll look into it on my branch.

nathanrussell-moj-digital: Yes, we can defo rename / rework this stuff moving forwards 👍

PR #128, opened 2023-08-23
Print styling

chrimesdev:
**Description**

- Use the GOVUK Design System utility class to hide common elements on print (Headers, Footers, Buttons, Links)
- Add a print stylesheet which only gets applied to print media and relevant config for compiling SCSS to CSS
- Add some basic styles to the print stylesheet to avoid breaks in middle of components, reduce spacing and ensure that colours are only black and white

**Issues**

[RR-200](https://dsdmoj.atlassian.net/browse/RR-200)
[RR-212](https://dsdmoj.atlassian.net/browse/RR-212)
[RR-256](https://dsdmoj.atlassian.net/browse/RR-256)
[RR-257](https://dsdmoj.atlassian.net/browse/RR-257)
[RR-258](https://dsdmoj.atlassian.net/browse/RR-258)

**Screenshots**

### DIFF
--- a/assets/scss/print.scss
+++ b/assets/scss/print.scss
@@ -1,0 +1,15 @@
+.govuk-hint,
+.govuk-tag,
+.govuk-caption-l {
+  color: #0b0c0c
+}
+
+.govuk-summary-card {
+  border: 4px solid #0b0c0c;
+  break-inside: avoid
+}
+
+.app-functional-skills-sidebar {
+  break-inside: avoid;
+  margin-top: 32px;
+}
--- a/server/views/partials/layout.njk
+++ b/server/views/partials/layout.njk
@@ -25,6 +25,8 @@
   <link href="/assets/stylesheets/application-ie8.css?{{ version }}" rel="stylesheet"/>
   <script src="/assets/js/html5shiv-3.7.3.min.js"></script>
   <![endif]-->
+  
+  <link href="/assets/stylesheets/print.css?{{ version }}" media="print" rel="stylesheet" />
 
   <script src="/assets/js/jquery.min.js"></script> 
 {% endblock %}
--- a/server/views/partials/layout.njk
+++ b/server/views/partials/layout.njk
@@ -6,6 +6,7 @@
 {% from "govuk/components/details/macro.njk" import govukDetails %}
 {% from "govuk/components/error-summary/macro.njk" import govukErrorSummary %}
 {% from "govuk/components/fieldset/macro.njk" import govukFieldset %}
+{% from "govuk/components/footer/macro.njk" import govukFooter %}
 {% from "govuk/components/inset-text/macro.njk" import govukInsetText %}
 {% from "govuk/components/input/macro.njk" import govukInput %}
 {% from "govuk/components/phase-banner/macro.njk" import govukPhaseBanner %}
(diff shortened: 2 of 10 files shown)


server/views/partials/prisonerBanner.njk
```diff
@@ -1,15 +1,17 @@
...
-  </div>
-  <div class="govuk-grid-column-one-quarter">
-    <p class="govuk-!-font-weight-bold govuk-!-margin-bottom-6">Reception date: <span class="govuk-body" data-qa="reception-date">{{ prisonerSummary.receptionDate | formatDate('D MMM YYYY') }}</span></p>
-  </div>
-  <div class="govuk-grid-column-one-quarter">
-    <p class="govuk-!-font-weight-bold govuk-!-margin-bottom-6">Earliest release date: <span class="govuk-body" data-qa="release-date">{{ prisonerSummary.releaseDate | formatDate('D MMM YYYY') }}</span></p>
-  </div>
-  <div class="govuk-grid-column-one-quarter">
-    <p class="govuk-!-font-weight-bold govuk-!-margin-bottom-6">DOB: <span class="govuk-body" data-qa="dob">{{ prisonerSummary.dateOfBirth | formatDate('D MMM YYYY') }}</span></p>
+<div class="app-prisoner-banner govuk-!-display-none-print">
```
chrimesdev, line 1: Wrapped this in a `<div>` with the classname `app-prisoner-banner` to follow semantics of a custom component.

server/views/partials/layout.njk
```diff
@@ -56,6 +60,12 @@
   </div>
 {% endblock %}
 
+{% block footer %}
+  {{ govukFooter({
+    classes: "govuk-!-display-none-print"
```
chrimesdev, line 65: I've included the footer in our template instead of pulling it from the GOVUK shared template, this allows us to add custom classes and it will future proof us when we need to use the new footer from the DPS team.

assets/scss/print.scss
```diff
@@ -0,0 +1,15 @@
+.govuk-hint,
+.govuk-tag,
+.govuk-caption-l {
+  color: #0b0c0c
```
chrimesdev, line 4: This is the GOVUK black rather than `#000000`

assets/js/print.js
```diff
@@ -0,0 +1,7 @@
+function showDetailsElementOnPrint() {
+  document.querySelectorAll('.govuk-details').forEach(details => {
+    details.setAttribute('open', '')
+  })
+}
+
+window.addEventListener('beforeprint', () => showDetailsElementOnPrint())
```
chrimesdev, line 7: We need to use JavaScript here, as we cannot use CSS to expand details elements as its currently not in the CSS spec.

See: https://github.com/w3c/csswg-drafts/issues/2084

nathanrussell-moj-digital: So there's a possible (albeit remote) accessibility issue here in that if the browser does not have JS enabled (I accept this is unlikely) then they wont be able to print the notes as part of the goals on the overview page?

Which leads me onto ..... I thought there was discussion/conversation about whether the notes should be printed at all? The last I heard was No as there could be sensitive into in the notes that might not be appropriate to be printed/left lying around/given to the prisoner ? , but perhaps that has changed since then?

nathanrussell-moj-digital (commented):
Look great; just one question about whether we should be printing the notes at all?

nathanrussell-moj-digital (approved):
LGTM; 👍 
Good discussion and agreement in DSU today re: whether using JS to expand all the Notes sections and then print them is the right approach. Happy with we've had the discussion and team agreement on it 👍

PR #133, opened 2023-08-24
RR-115 - Added a couple of the filters we will need to display the CIAG enum values

nathanrussell-moj-digital:
There are a couple of other filters we will need for CIAG enums, but I have some questions about differences in the enums between their UI and their API

### DIFF
--- a/server/filters/formatGoalStatusValue.ts
+++ b/server/filters/formatGoalStatusValue.ts
@@ -3,7 +3,7 @@ export default function formatGoalStatusValue(value: string): string {
   return statusValue
 }
 
-export enum GoalStatusValue {
+enum GoalStatusValue {
   ACTIVE = 'In progress',
   COMPLETED = 'Completed',
   ARCHIVED = 'Archived',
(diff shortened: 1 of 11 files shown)


server/filters/formatGoalStatusValue.ts
```diff
@@ -3,7 +3,7 @@ export default function formatGoalStatusValue(value: string): string {
   return statusValue
 }
 
-export enum GoalStatusValue {
+enum GoalStatusValue {
```
nathanrussell-moj-digital, line 6: These didnt need to be exported, and are effectively private to the internals of the filter function 👍

server/filters/formatYesNoFilter.ts
```diff
@@ -0,0 +1,27 @@
+export default function formatYesNoFilter(value: string | boolean): string {
```
nathanrussell-moj-digital, line 1: Whilst this has been written to handle CIAG enum values of YES, NO and NOT_SURE; it felt like we could make it a bit more generic so that it can also handle booleans, Y and N as well 👍

PR #139, opened 2023-08-29
RR-211 - Refactor router request handlers into a single module

nathanrussell-moj-digital:
This PR refactors the `routerRequestHandlers` modules into a single module common across all routes (ie. in the `routes` folder)

Previously we had a `routerRequestHandlers` module in a couple of route specific folders (one for `createGoal`, one for `updateGoal`)

The `routerRequestHandlers` module in `createGoal` contained a function `checkPrisonerSummaryExistsInSession` which was then subsequently used by the routes in `updateGoal`. Therefore the `updateGoal` functions had a dependency on the `createGoal` functions, which was wrong.

This PR brings both `routerRequestHandlers` modules from `createGoal` and `updateGoal` together into a single module in `routes`


mattwills-moj-digital (approved):
GS

PR #147, opened 2023-08-30
RR-33 use the correct markup for add another and add basic remove fun…

chrimesdev:
## Description

- Use the correct markup for the [Add another](https://design-patterns.service.justice.gov.uk/components/add-another/) component by adding a `tabindex` to the `<h2>`, wrap the form in a `fieldset` and move the Step [number] title into a legend and wrap the buttons in `<div class="moj-button-action">`
- Display the Remove step button if there are more than 1 steps in the loop
- Basic functionality to remove the step from the form

### To do

- [ ] The form currently validates the input fields before allowing a user to remove the step, a user should be able to remove a step which has empty fields. If a user clicks the "Add another step" button but then they decide they no longer require that step, they cannot remove it due to validation. 
- [ ] Tests

## Screenshots

Screenshot 2023-08-30 at 14 01 35

### DIFF
--- a/server/@types/forms/index.d.ts
+++ b/server/@types/forms/index.d.ts
@@ -35,5 +35,6 @@ declare module 'forms' {
       | 'MORE_THAN_TWELVE_MONTHS'
     stepNumber: number
     status: 'NOT_STARTED' | 'ACTIVE' | 'COMPLETE'
+    action?: 'delete-step'
   }
 }
--- a/server/routes/updateGoal/updateGoalController.ts
+++ b/server/routes/updateGoal/updateGoalController.ts
@@ -41,6 +41,7 @@ export default class UpdateGoalController {
   submitUpdateGoalForm: RequestHandler = async (req, res, next): Promise<void> => {
     const { prisonNumber, goalReference } = req.params
     const updateGoalForm: UpdateGoalForm = { ...req.body }
+    const updateStepForm: UpdateStepForm = { ...req.body }
     req.session.updateGoalForm = updateGoalForm
 
     const errors = validateUpdateGoalForm(updateGoalForm)
@@ -58,6 +59,15 @@ export default class UpdateGoalController {
       updateGoalForm.steps.push(newStep)
       // Redirect back to the Update Goal page with named anchor taking the user straight to the new step
       return res.redirect(`/plan/${prisonNumber}/goals/${goalReference}/update#steps[${nextStepNumber - 1}][title]`)
+    }
+
+    // Remove the desired step on the action delete step
+    if (updateStepForm.action === 'delete-step') {
+      const { stepNumber } = updateStepForm
+      const stepIndex = updateGoalForm.steps.findIndex(step => step.stepNumber === stepNumber)
+      updateGoalForm.steps.splice(stepIndex, 1)
+      // Redirect back to the Update Goal page with named anchor taking the user to the edit and remove steps section
+      return res.redirect(`/plan/${prisonNumber}/goals/${goalReference}/update#edit-and-remove-steps`)
     }
 
     return res.redirect(`/plan/${prisonNumber}/goals/${goalReference}/update/review`)
--- a/server/views/pages/goal/update/index.njk
+++ b/server/views/pages/goal/update/index.njk
@@ -69,13 +69,20 @@
 
           <hr class="govuk-section-break govuk-section-break--visible govuk-section-break--m">
 
-          <h2 class="govuk-heading-m">Edit and remove steps</h2>
+          <h2 class="govuk-heading-m" id="edit-and-remove-steps" tabindex="-1">Edit and remove steps</h2>
 
             {% for step in form.steps %}
               <input type="hidden" name="steps[{{ (loop.index-1) }}][reference]" value="{{ step.reference }}" data-qa="step-{{ (loop.index-1) }}-reference" />
               <input type="hidden" name="steps[{{ (loop.index-1) }}][stepNumber]" value="{{ step.stepNumber }}" />
 
-              <h3 class="govuk-heading-m">Step {{ step.stepNumber }}</h3>
+              {% call govukFieldset({
+                classes: 'moj-add-another__item',
+                legend: {
+                  text: 'Step ' + step.stepNumber,
+                  classes: 'moj-add-another__title govuk-fieldset__legend--m',
+                  isPageHeading: false
+                }
+              }) %}
 
               {{ govukTextarea({
                 name: 'steps[' + (loop.index-1) + '][title]',
@@ -155,10 +162,25 @@
                 errorMessage: errors | findError('steps[' + (loop.index-1) + '][targetDateRange]')
               }) }}
 
+              {% if loop.index > 1 %}
+                {{ govukButton({
+                  id: 'delete-step',
+                  attributes: {
+                    'data-qa': 'goal-update-delete-step-button'
+                  },
+                  name: "action",
+                  value: "delete-step",
+                  text: "Remove",
+                  classes: "govuk-button--secondary moj-add-another__remove-button"
+                }) }}
+              {% endif %}
+
+              {% endcall %}
+
               <hr class="govuk-section-break govuk-section-break--visible govuk-section-break--m">
             {% endfor %}
-
-            <div class="govuk-form-group">
+            
+            <div class="moj-button-action">
               {{ govukButton({
                 id: "add-another-step-button",
                 attributes: {
@@ -167,19 +189,21 @@
                 name: "action",
                 value: "add-another-step",
                 text: "Add another step",
-                classes: "govuk-button--secondary"
+                classes: "govuk-button--secondary govuk-!-margin-bottom-4"
               }) }}
             </div>
 
-            {{ govukButton({
-              id: "submit-button",
-              attributes: {
-                'data-qa': 'goal-update-submit-button'
-              },
-              name: "action",
-              value: "submit-form",
-              text: "Submit"
-            }) }}
+            <div class="moj-button-action">
+              {{ govukButton({
+                id: "submit-button",
+                attributes: {
+                  'data-qa': 'goal-update-submit-button'
+                },
+                name: "action",
+                value: "submit-form",
+                text: "Submit"
+              }) }}
+            </div>
         </form>
 
     </div>


server/routes/updateGoal/updateGoalController.ts
```diff
@@ -41,10 +41,11 @@ export default class UpdateGoalController {
   submitUpdateGoalForm: RequestHandler = async (req, res, next): Promise<void> => {
     const { prisonNumber, goalReference } = req.params
     const updateGoalForm: UpdateGoalForm = { ...req.body }
+    const updateStepForm: UpdateStepForm = { ...req.body }
     req.session.updateGoalForm = updateGoalForm
 
     const errors = validateUpdateGoalForm(updateGoalForm)
-    if (errors.length > 0) {
+    if (errors.length > 0 && !updateStepForm.action.includes('delete-step-')) {
```
chrimesdev, line 48: This seems to be breaking the integration tests for submitting the page with validation issues, however the Cypress tests are passing fine for submitting the page with validation issues.

server/views/pages/goal/update/review.njk
```diff
@@ -77,7 +77,7 @@
         <tbody class="govuk-table__body">
           {% for step in data.steps %}
           <tr class="govuk-table__row">
-            <th scope="row" class="govuk-table__header">Step {{ step.stepNumber }}</th>
+            <th scope="row" class="govuk-table__header">Step {{loop.index}}</th>
```
chrimesdev, line 80: Using `loop.index` here as the `stepNumber` has not been updated in the database at this point, as the user should review the answers before saving the data.

server/routes/updateGoal/updateGoalController.ts
```diff
@@ -60,6 +61,17 @@ export default class UpdateGoalController {
       return res.redirect(`/plan/${prisonNumber}/goals/${goalReference}/update#steps[${nextStepNumber - 1}][title]`)
     }
 
+    // Remove the desired step on the action delete step
+    if (updateStepForm.action.includes('delete-step-')) {
+      // Get the step reference inbetween the 2 characters [ ] from the action value
+      const stepReference = updateStepForm.action.match(/\[(.*?)\]/)[1]
```
chrimesdev, line 67: Couldn't think of a better way to find out which step to update than passing the reference through the button value

server/routes/updateGoal/updateGoalController.ts
chrimesdev: This is to stop the validation of empty form fields when the Remove step button is clicked. A user could click "Add another step" but then decide they don't need that step anymore, but because the Step data is blank, they wouldn't be able to remove it.

integration_tests/e2e/goal/updateGoal.cy.ts
```diff
@@ -114,6 +114,35 @@ context('Update a goal', () => {
...
+
+    // Then
+    Page.verifyOnPage(OverviewPage)
+    // Assert that the expected step was removed from the UpdateGoal request to the API - TODO, there has to be a better way of doing this
+    cy.task<UpdateGoalRequest>('getUpdateGoalRequestBody')
+      .then(updateGoalRequestBody => {
+        const { steps } = updateGoalRequestBody
+        return steps
+      })
+      .should('have.length', 0)
```
nathanrussell-moj-digital, line 143: Does this assertion suggest we are sending the update goal request to the API with zero steps?
That shouldn't be allowed - we must always have at least 1 step

nathanrussell-moj-digital: And if the When is removing the 2nd step (`clickRemoveButtonForSecondStep()`); then the first step should still be there (plus any steps that were after step 2)

server/@types/forms/index.d.ts
```diff
@@ -35,5 +35,6 @@ declare module 'forms' {
       | 'MORE_THAN_TWELVE_MONTHS'
     stepNumber: number
     status: 'NOT_STARTED' | 'ACTIVE' | 'COMPLETE'
+    action?: `delete-step-[${string}]`
```
nathanrussell-moj-digital, line 38: Not sure about this? - not sure about the string interpolation .... 🤔

server/routes/updateGoal/updateGoalController.ts
```diff
@@ -41,10 +41,11 @@ export default class UpdateGoalController {
   submitUpdateGoalForm: RequestHandler = async (req, res, next): Promise<void> => {
     const { prisonNumber, goalReference } = req.params
     const updateGoalForm: UpdateGoalForm = { ...req.body }
+    const updateStepForm: UpdateStepForm = { ...req.body }
```
nathanrussell-moj-digital, line 44: Not sure I follow this (or, it feels kinda weird) - how can 2 different variables of 2 different types be instantiated from the same data? (`req.body`)
```
    const updateGoalForm: UpdateGoalForm = { ...req.body }
    const updateStepForm: UpdateStepForm = { ...req.body }
```

server/@types/forms/index.d.ts
nathanrussell-moj-digital: I'm not sure this is the form we want to use (by adding an action field)

I think the action "delete-step" belongs on the action field of `UpdateGoalForm`
That does then beg the question about how we tell it which step we want to remove .... let me give that some thought .... 🤔

server/routes/updateGoal/updateGoalController.ts
```diff
@@ -41,10 +41,11 @@ export default class UpdateGoalController {
...
     req.session.updateGoalForm = updateGoalForm
```
nathanrussell-moj-digital, line 46: You want to remove line 44; then here (before the validation), you want to do something like:
```
if (updateGoalForm.action === 'delete-step') {
    // work out which step to remove .... not sure how to do that yet, needs thought... 🤔 
   // remove the step from the array updateGoalForm.steps
   // Redirect back to the Update Goal page
    return res.redirect(`/plan/${prisonNumber}/goals/${goalReference}/update`)
}
```

nathanrussell-moj-digital: If we check for deleting steps before the validation check, then you can do away with your change on line 48, and we wont have the problem where you cant delete a step you've accidentlally added with no/blank data

server/views/pages/goal/update/index.njk
```diff
@@ -154,11 +161,26 @@
...
-
+              
+              {% if loop.length > 1 %}
+                {{ govukButton({
+                  id: "delete-step-" + (loop.index-1),
+                  attributes: {
+                    'data-qa': 'step-' + (loop.index-1) + '-remove-button'
+                  },
+                  name: "action",
+                  value: 'delete-step-[' + step.reference + ']',
```
nathanrussell-moj-digital, line 172: `step.reference` wont have a value in the case of a new step being added, but not saved to the database (eg: user adds a step then immediately wants to delete it)
The `step.stepNumber` would be better (which will have a value, even if not persisted to the database) - though I'm not convinced that including it dynamically in the button action is the way to go (though I've not thought of a better way yet)

server/views/pages/goal/update/review.njk
chrimesdev: If a goal has 3 steps and I remove step 2,
The `step.stepNumber` in the database for step 3 is still `3`, until I save the form on the review screen
Using the `step.stepNumber` as a table header on the review screen here, will display 
Step 1 - Value
Step 3 - Value

As the `step.stepNumber` hasn't been updated yet

nathanrussell-moj-digital: Ah, I see. In that case we need to re-jig / re-sequence the numbers as part of removing a step in the controller method - so that they array of steps are always sequential 👍

server/views/pages/goal/update/index.njk
```diff
@@ -69,13 +69,20 @@
...
 
             {% for step in form.steps %}
               <input type="hidden" name="steps[{{ (loop.index-1) }}][reference]" value="{{ step.reference }}" data-qa="step-{{ (loop.index-1) }}-reference" />
               <input type="hidden" name="steps[{{ (loop.index-1) }}][stepNumber]" value="{{ step.stepNumber }}" />
 
-              <h3 class="govuk-heading-m">Step {{ step.stepNumber }}</h3>
+              {% call govukFieldset({
+                classes: 'moj-add-another__item',
+                legend: {
+                  text: 'Step ' + (loop.index),
```
nathanrussell-moj-digital, line 81: `stepNumber` rather than the loop index I think 🤔

server/views/pages/goal/update/index.njk
nathanrussell-moj-digital: Like I said, I'm not 100% on referencing the step in some way in the button action; but I can't think of a better way of doing it (without javascript!)
So, lets go with your approach, but use the `stepNumber` because `stepReferrence` wont always be set/present
```
value: 'delete-step-[' + step.stepNumber + ']',
```

server/@types/forms/index.d.ts
nathanrussell-moj-digital: Yes; remove the `action` from the `UpdateStepForm` and add it to line 25 instead (ie. add to the `action` property of `UpdateGoalForm`
```
action?: 'add-another-step' | 'submit-form' | 'delete-step-[0]'
```
Hardcode it here as a zero - I dont think it actually matters at runtime, and it serves as an example that an index is passed in and it is zero indexed

server/routes/updateGoal/updateGoalController.ts
nathanrussell-moj-digital: Based on what I've just suggested about form action on `UpdateGoalForm` I think this part of the code would be
```
    // Remove the desired step on the action delete step
    if (updateStepForm.action.startsWith('delete-step-')) {
      // Get the step index inbetween the 2 characters [ ] from the action value
      const stepIndex = updateStepForm.action.match(/\[(.*?)\]/)[1]

      // Remove the desired step from the array
      updateGoalForm.steps.splice(stepIndex, 1)

      // Re-sequence the step array so that all the step's stepNumber fields are sequential starting from 1
      ....
      ....

      // Redirect back to the Update Goal page with named anchor taking the user to the edit and remove steps section
      return res.redirect(`/plan/${prisonNumber}/goals/${goalReference}/update#edit-and-remove-steps`)
    }
```

server/routes/updateGoal/updateGoalController.ts
```diff
@@ -43,6 +43,22 @@ export default class UpdateGoalController {
     const updateGoalForm: UpdateGoalForm = { ...req.body }
     req.session.updateGoalForm = updateGoalForm
 
+    // Remove the desired step on the action delete step
+    if (updateGoalForm.action && updateGoalForm.action.startsWith('delete-step-')) {
```
chrimesdev, line 47: Check that there is a updateGoalForm.action when using `startsWith` due to test error

```
UpdateGoalController › submitUpdateGoalForm › should redirect to update goal form given validation fails

    TypeError: Cannot read properties of undefined (reading 'startsWith')
```

nathanrussell-moj-digital (approved):
Fantastic; this looks great. Definitely worth going through the various permutations of implementation options to end up with this; looks nice and neat (even with the es-lint error, which probably is not worth worrying too much about )

chrimesdev:
> Fantastic; this looks great. Definitely worth going through the various permutations of implementation options to end up with this; looks nice and neat (even with the es-lint error, which probably is not worth worrying too much about )

I will get around to refactoring that, but I can live with it for now!

PR #162, opened 2023-09-12
RR-143 implement the DPS connect Footer component

chrimesdev:
## Description

Implement the DPS connect Footer component and CSS/JS assets.

### To do

- [ ] `frontendComponentApiClient` tests

## Tickets

https://dsdmoj.atlassian.net/browse/RR-143
https://dsdmoj.atlassian.net/browse/RR-295
https://dsdmoj.atlassian.net/browse/RR-296
https://dsdmoj.atlassian.net/browse/RR-297

## Screenshots

Screenshot 2023-09-12 at 10 37 46

### DIFF
--- a/server/@types/frontendComponentApiClient/index.d.ts
+++ b/server/@types/frontendComponentApiClient/index.d.ts
@@ -1,0 +1,7 @@
+declare module 'frontendComponentApiClient' {
+  export interface Component {
+    html: string
+    css: string[]
+    javascript: string[]
+  }
+}
--- a/server/app.ts
+++ b/server/app.ts
@@ -16,10 +16,12 @@ import setUpWebSecurity from './middleware/setUpWebSecurity'
 import setUpWebSession from './middleware/setUpWebSession'
 
 import routes from './routes'
+import { DataAccess } from './data'
 import type { Services } from './services'
 import authorisationMiddleware from './middleware/authorisationMiddleware'
+import setUpFrontendComponents from './middleware/fetchFrontendComponentMiddleware'
 
-export default function createApp(services: Services): express.Application {
+export default function createApp(services: Services, dataAccess: DataAccess): express.Application {
   const app = express()
 
   app.set('json spaces', 2)
@@ -37,6 +39,7 @@ export default function createApp(services: Services): express.Application {
   app.use(authorisationMiddleware())
   app.use(setUpCsrf())
   app.use(setUpCurrentUser(services))
+  app.use(setUpFrontendComponents(dataAccess))
 
   app.use(routes(services))
 
--- a/server/middleware/fetchFrontendComponentMiddleware.ts
+++ b/server/middleware/fetchFrontendComponentMiddleware.ts
@@ -1,0 +1,28 @@
+import { Router } from 'express'
+import { DataAccess } from '../data'
+import logger from '../../logger'
+import config from '../config'
+
+export default function setUpFrontendComponents({ frontendComponentApiClient }: DataAccess): Router {
+  const router = Router({ mergeParams: true })
+
+  router.get('*', async (req, res, next) => {
+    try {
+      // Frontend components API is only used when feature toggle is provided
+      if (config.featureToggles.frontendComponentsApiToggleEnabled) {
+        const { user } = res.locals
+        const [footer] = await Promise.all([frontendComponentApiClient.getComponent('footer', user.token)])
+        res.locals.feComponents = {
+          footer: footer.html,
+          cssIncludes: [...footer.css],
+          jsIncludes: [...footer.javascript],
+        }
+      }
+    } catch (error) {
+      logger.error(error, 'Failed to retrieve front end components')
+    }
+    next()
+  })
+
+  return router
+}
--- a/server/config.ts
+++ b/server/config.ts
@@ -100,14 +100,14 @@ export default {
       },
       agent: new AgentConfig(Number(get('CIAG_INDUCTION_API_TIMEOUT_RESPONSE', 10000))),
     },
-  },
-  frontendComponents: {
-    url: get('FRONTEND_COMPONENT_API_URL', 'http://localhost:8083', requiredInProduction),
-    timeout: {
-      response: Number(get('FRONTEND_COMPONENT_API_TIMEOUT_RESPONSE', 50000)),
-      deadline: Number(get('FRONTEND_COMPONENT_API_TIMEOUT_DEADLINE', 50000)),
+    frontendComponents: {
+      url: get('FRONTEND_COMPONENT_API_URL', 'http://localhost:8083', requiredInProduction),
+      timeout: {
+        response: Number(get('FRONTEND_COMPONENT_API_TIMEOUT_RESPONSE', 50000)),
+        deadline: Number(get('FRONTEND_COMPONENT_API_TIMEOUT_DEADLINE', 50000)),
+      },
+      agent: new AgentConfig(Number(get('FRONTEND_COMPONENT_API_TIMEOUT_RESPONSE', 5000))),
     },
-    agent: new AgentConfig(Number(get('FRONTEND_COMPONENT_API_TIMEOUT_RESPONSE', 5000))),
   },
   domain: get('INGRESS_URL', 'http://localhost:3000', requiredInProduction),
   dpsHomeUrl: get('DPS_URL', 'http://localhost:3000/', requiredInProduction),
--- a/server/data/frontendComponentApiClient.ts
+++ b/server/data/frontendComponentApiClient.ts
@@ -1,0 +1,16 @@
+import type { Component } from 'frontendComponentApiClient'
+import RestClient from './restClient'
+import config from '../config'
+
+export default class FrontendComponentApiClient {
+  private static restClient(token: string): RestClient {
+    return new RestClient('Frontend Component API', config.apis.frontendComponents, token)
+  }
+
+  async getComponent(component: 'header' | 'footer', token: string) {
+    return FrontendComponentApiClient.restClient(token).get({
+      path: `/${component}`,
+      headers: { 'x-user-token': token },
+    }) as Promise<Component>
+  }
+}
--- a/server/data/index.ts
+++ b/server/data/index.ts
@@ -17,6 +17,7 @@ import PrisonerSearchClient from './prisonerSearchClient'
 import EducationAndWorkPlanClient from './educationAndWorkPlanClient'
 import CuriousClient from './curiousClient'
 import CiagInductionClient from './ciagInductionClient'
+import FrontendComponentApiClient from './frontendComponentApiClient'
 
 type RestClientBuilder<T> = (token: string) => T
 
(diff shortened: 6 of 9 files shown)


server/app.ts
```diff
@@ -37,6 +39,7 @@ export default function createApp(services: Services): express.Application {
   app.use(authorisationMiddleware())
   app.use(setUpCsrf())
   app.use(setUpCurrentUser(services))
+  app.use(setUpFrontendComponents(dataAccess))
```
nathanrussell-moj-digital, line 42: Hmm, I'm not sure how comfortable I am with this (injecting all of the data access clients)
I think this is why when I subtasked it, one of the subtasks was to create a service
I think I'd rather have a front end components service that that has the relevant data access client class injected into it (ie. the same pattern as what we already have)

Could you have a look at creating a service class please, that has one method in it (`getFooter`) that calls the client class and method
Then when we do the header we'll add another method to the service called `getHeader` (but dont do that in this PR)

server/data/frontendComponentApiClient.test.ts
```diff
@@ -0,0 +1,36 @@
...
+  afterEach(() => {
+    nock.cleanAll()
+  })
+
+  describe('getComponent', () => {
+    it('should return data from api', async () => {
+      // Given
+      const userToken = 'a-user-token'
+
+      const expectedResponse = { data: 'data' }
```
nathanrussell-moj-digital, line 24: Could you create a test data builder in `testsupport` please that represents a real response please? Whilst the test itself is nice and easy to read, I can't grok what the response data "looks" like from this test, so would be hard for someone else to maintain the code and make future changes (eg:. from reading this test I might incorrectly assume the response from the frontend components API was an object with a single field `data`)

Something like:
```
const aValidFrontEndComponentsFooterResponse = (): FrontEndComponentResponse {
    html: '<p>some html</p>',
    css: ['https://frontend-componenents-dev/footerStyles.css'],
    js: ['https://frontend-componenents-dev/footerScripts.js'],
}

```

server/@types/frontendComponentApiClient/index.d.ts
```diff
@@ -0,0 +1,7 @@
+declare module 'frontendComponentApiClient' {
+  export interface Component {
+    html: string
+    css: string[]
+    javascript: string[]
+  }
+}
```
nathanrussell-moj-digital, line 7: To follow convention with what we've already done, could you create both an api and apiclient folder please? Something like:

frontendComponentApi/index.d.ts
```
export interface Component {
  html: string
  css: string[]
  javascript: string[]
}
```

and then frontendComponentApiClient.index.d.ts
```
declare module 'frontendComponentApiClient' {
  import Component from '../frontendComponentApi'

  export type FrontendComponent = Component
}
```

I know it might seem like writing code for the sake of it, and it might not be adding much; but it would be consistent with everything else we already have, plus would allow for us to import their swagger spec (if they ever produce one) without too much hassle

server/middleware/fetchFrontendComponentMiddleware.ts
```diff
@@ -0,0 +1,28 @@
...
+    try {
+      // Frontend components API is only used when feature toggle is provided
+      if (config.featureToggles.frontendComponentsApiToggleEnabled) {
+        const { user } = res.locals
+        const [footer] = await Promise.all([frontendComponentApiClient.getComponent('footer', user.token)])
+        res.locals.feComponents = {
+          footer: footer.html,
+          cssIncludes: [...footer.css],
+          jsIncludes: [...footer.javascript],
+        }
```
nathanrussell-moj-digital, line 19: I wonder if this object type could be defined as a type in our viewModels module? 🤔 
Something like
```
declare module `viewModels` {
...
  export interface FrontendComponentsPageAdditions {
    footerHtml: string,
    cssIncludes: Array<string>,
    jsIncludes: Array<string>,
  }
...

```

nathanrussell-moj-digital: Then what you've written here becomes:
```
        res.locals.feComponents = {
          footerHtml: footer.html,
          cssIncludes: [...footer.css],
          jsIncludes: [...footer.javascript],
        } as FrontendComponentsPageAdditions
```

Is there much gain from this? Well, arguably not; but we are using typescript so I like to use types where we can 😁 . Plus it allows us to refactor this out into a mapper function or similar (I'm not proposing you do that here, but it does support us doing that in the future if we want)

server/middleware/fetchFrontendComponentMiddleware.ts
```diff
@@ -0,0 +1,28 @@
+import { Router } from 'express'
+import { DataAccess } from '../data'
+import logger from '../../logger'
+import config from '../config'
+
+export default function setUpFrontendComponents({ frontendComponentApiClient }: DataAccess): Router {
```
nathanrussell-moj-digital, line 6: I'm not sure about the structure of this function 🤔 
We dont have a unit test for it, and I'm not sure how easy it would be to write one because it news up the `Router` internally
I wonder if there's a better/different way of structuring this .... 🤔

server/middleware/fetchFrontendComponentMiddleware.ts
```diff
@@ -0,0 +1,28 @@
...
+  const router = Router({ mergeParams: true })
```
nathanrussell-moj-digital, line 7: What does `mergeParams` do? Why are we setting it?

nathanrussell-moj-digital (commented):
Maybe not for this PR, but as part of this story (create a new subtask for it) we will need to
* enable the feature toggle in `feature.env` for our cypress integration tests
* create a wiremock stub for the cypress tests so that every cypress test gets the footer from the API
* create a wiremock stub in `wiremock/mappings` to support running it locally with wiremockj

nathanrussell-moj-digital:
Also, does this PR handle the eventuality that we the feature toggle is enabled, but the API call fails? In that case we need to display the approved Connect DPS fallback, which I believe is a simple grey box.
In the ticket there are 3 ACs - I'm not sure this PR is meeting the last one yet?

server/data/frontendComponentApiClient.test.ts
chrimesdev: Added a `aValidFrontEndComponentFooter` response and used it within the test

server/middleware/fetchFrontendComponentMiddleware.ts
chrimesdev: Used types here following the change for adding `export interface FrontendComponentsPageAdditions`

server/middleware/fetchFrontendComponentMiddleware.ts
chrimesdev: Currently rewriting this logic with a service

server/middleware/fetchFrontendComponentMiddleware.ts
chrimesdev: Removed as part of rewrite

server/app.ts
chrimesdev: Removed as part of rewrite

chrimesdev:
> Maybe not for this PR, but as part of this story (create a new subtask for it) we will need to
> 
> * enable the feature toggle in `feature.env` for our cypress integration tests
> * create a wiremock stub for the cypress tests so that every cypress test gets the footer from the API
> * create a wiremock stub in `wiremock/mappings` to support running it locally with wiremockj

Created ticket for this https://dsdmoj.atlassian.net/browse/RR-316

chrimesdev:
> Also, does this PR handle the eventuality that we the feature toggle is enabled, but the API call fails? In that case we need to display the approved Connect DPS fallback, which I believe is a simple grey box. In the ticket there are 3 ACs - I'm not sure this PR is meeting the last one yet?

Added this as part of the to do items for tomorrow!

chrimesdev:
> Also, does this PR handle the eventuality that we the feature toggle is enabled, but the API call fails? In that case we need to display the approved Connect DPS fallback, which I believe is a simple grey box. In the ticket there are 3 ACs - I'm not sure this PR is meeting the last one yet?

Tested this and the footer falls back to the standard footer we currently have, the same one that DPS provide the templates for is the just the basic Nunjucks Footer!

chrimesdev:
Added the third footer scenario fallback

- Feature toggle disabled - show the footer we always have done
- Feature toggle enabled and API call works - show the data from the API as the footer
- Feature toggle enabled and API call fails - show the empty grey footer

server/middleware/fetchFrontendComponentMiddleware.ts
```diff
@@ -0,0 +1,21 @@
...
+    try {
+      const footer = await frontendComponentService.getComponents('footer', res.locals.user.token)
+      res.locals.feComponents = {
+        footerHtml: footer.html,
+        cssIncludes: [...footer.css],
+        jsIncludes: [...footer.javascript],
+      } as FrontendComponentsPageAdditions
+      next()
+    } catch (error) {
+      logger.error(error, 'Failed to retrieve front end components')
```
nathanrussell-moj-digital, line 17: I'm not sure this is ever triggered as I _think_ you're catching the error in the service (which is where I think you should be catching it)
Its probably OK to merge at the moment (to allow you to move forward), but I think I'll look at this once you've merged 👍 👀

server/services/frontendComponentService.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export default class FrontendComponentService {
+  constructor(private readonly frontendApiClient: FrontendComponentApiClient) {}
+
+  async getComponents(component: 'header' | 'footer', userToken: string): Promise<FrontendComponent> {
+    try {
+      const frontendComponentResponse = await this.frontendApiClient.getComponents(component, userToken)
+      return frontendComponentResponse
+    } catch (error) {
+      logger.error(`Error retrieving DPS frontend components: ${error}`)
```
nathanrussell-moj-digital, line 13: Here's the catch block I was referring to in my previous comment - I think this is correct
(though I think there might be a problem with the type itself potentially not having a `problemRetrievingData` field? -  but I'll look into it after you've merged )

server/testsupport/frontendComponentTestDataBuilder.ts
```diff
@@ -0,0 +1,9 @@
+import type { FrontendComponentsPageAdditions } from 'viewModels'
+
+export default function aValidFrontEndComponentFooter(): FrontendComponentsPageAdditions {
+  return {
```
nathanrussell-moj-digital, line 4: Nice 👍

nathanrussell-moj-digital (approved):
Nice 👍 
LGTM - as mentioned in a couple of comments I _think_ some of the error handling might need a little looking at / tidying, but I'm happy to do that after you've merged this 👍

chrimesdev:
> Nice 👍 LGTM - as mentioned in a couple of comments I _think_ some of the error handling might need a little looking at / tidying, but I'm happy to do that after you've merged this 👍

I've created the ticket https://dsdmoj.atlassian.net/browse/RR-318

PR #165, opened 2023-09-13
RR-153 - Initial refactoring of "create goal" form types

nathanrussell-moj-digital:
This PR does some initial refactoring to group all forms associated with creating a goal into a single object.

The story that this is part of is about supporting creating multiple goals as part of the Create Goal process.
Currently we only support creating a single Goal at a time, and that single Goal (inc all of its steps) is shown on the Review Your Goals (check your answers) page.
The business objective of the story is to allow creating many Goals as part of the process, and showing each Goal on the check your answers page. In order to do this we need to refactor the data structures etc of how we store the Goal in session before we send it to the API

Currently we have a number of form types (one for each page in the multi-page 'wizard' process), and each of these form objects is held in the session as we move from page to page. Currently we have the following in session:
```
  interface SessionData {
    ....
    createGoalForm: CreateGoalForm
    addStepForm: AddStepForm
    addStepForms: Array<AddStepForm>
    addNoteForm: AddNoteForm
    ....
  }
```
On the check your answers page we collate the data from these form objects to render the Goal you are about to create, then map the data into a DTO and call the API to create the goal.
All of this is based on the idea that we only support a single goal at the moment.

In order to support multiple goals, at a high level we need to:

1. Group these form objects into a single object / container of some kind
2. Hold an array of the grouped objects in session, rather than each individual form object

This PR addresses point 1 only from the above list.

I very much welcome naming suggestions, but I've created the concept of a "composite" form object called `NewGoal`. The fields in `NewGoal` are the 4 create goal form types that were previously directly on the session. An instance of `NewGoal` is then held on the session instead. It's just a container object.

The rest of the refactoring in this PR simply changes all the references where we originally referenced a form type from the session, to instead reference the relevant property of the `NewForm` object on the session.

My plan is that once this PR is approved and merged, we can then look at refactoring such that the session holds an array of `NewGoal` rather than a single one.

And once that is in place we can add the "Add another Goal" button and create a new instance of `NewForm` in the array, and take the user back to the first screen in the process.

Thats the plan anyway ...

### DIFF
--- a/server/@types/forms/index.d.ts
+++ b/server/@types/forms/index.d.ts
@@ -37,3 +37,23 @@ declare module 'forms' {
     status: 'NOT_STARTED' | 'ACTIVE' | 'COMPLETE'
   }
 }
+
+/**
+ * Module declaring "composite forms", specifically types that contain other forms.
+ * The composite form is not a form object in its own right, in that there are no HTML views, forms or express request
+ * handlers that bind request form fields to an object of this type. It is a convenience object to allow collating
+ * several form objects into a single object held in the session; typically used in multi-page / wizard style screens.
+ */
+declare module 'compositeForms' {
+  import type { AddNoteForm, AddStepForm, CreateGoalForm } from 'forms'
+
+  /**
+   * A composite form representing the individual form objects for creating a new Goal
+   */
+  export interface NewGoal {
+    createGoalForm: CreateGoalForm
+    addStepForm: AddStepForm
+    addStepForms: Array<AddStepForm>
+    addNoteForm: AddNoteForm
+  }
+}
--- a/server/routes/createGoal/createGoalController.test.ts
+++ b/server/routes/createGoal/createGoalController.test.ts
@@ -70,7 +70,7 @@ describe('createGoalController', () => {
         'targetDate-year': '2024',
         action: 'submit-form',
       }
-      req.session.newGoalForm = {
+      req.session.newGoal = {
         addStepForms: [],
       } as NewGoal
 
@@ -86,7 +86,7 @@ describe('createGoalController', () => {
       // Then
       expect(res.redirect).toHaveBeenCalledWith('/plan/A1234GC/goals/add-note')
       expect(req.flash).not.toHaveBeenCalled()
-      expect(req.session.newGoalForm.addStepForms).toHaveLength(1)
+      expect(req.session.newGoal.addStepForms).toHaveLength(1)
     })
 
     it('should redirect to add step form given action is add-another-step and validation passes', async () => {
@@ -100,7 +100,7 @@ describe('createGoalController', () => {
         'targetDate-year': '2024',
         action: 'add-another-step',
       }
-      req.session.newGoalForm = {
+      req.session.newGoal = {
         addStepForms: [],
       } as NewGoal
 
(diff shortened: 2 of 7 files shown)


server/@types/forms/index.d.ts
```diff
@@ -37,3 +37,23 @@ declare module 'forms' {
     status: 'NOT_STARTED' | 'ACTIVE' | 'COMPLETE'
   }
 }
+
+/**
+ * Module declaring "composite forms", specifically types that contain other forms.
```
mattwills-moj-digital, line 42: Tidy idea 👌

server/@types/forms/index.d.ts
```diff
@@ -37,3 +37,23 @@ declare module 'forms' {
...
+ * handlers that bind request form fields to an object of this type. It is a convenience object to allow collating
+ * several form objects into a single object held in the session; typically used in multi-page / wizard style screens.
+ */
+declare module 'compositeForms' {
+  import type { AddNoteForm, AddStepForm, CreateGoalForm } from 'forms'
+
+  /**
+   * A composite form representing the individual form objects for creating a new Goal
+   */
+  export interface NewGoal {
```
mattwills-moj-digital, line 53: Whilst reading the description, I wondered if `NewGoalForms` would be better, but now I'm not so sure given it's already within the `compositeForms` module. Also, I guess that could become confusing when you introduce an array of these (if that's your intention).

server/routes/createGoal/createGoalController.test.ts
```diff
@@ -69,7 +70,9 @@ describe('createGoalController', () => {
         'targetDate-year': '2024',
         action: 'submit-form',
       }
-      req.session.addStepForms = []
+      req.session.newGoalForm = {
```
mattwills-moj-digital, line 73: `req.session.newGoalForm` or `req.session.newGoal`? 🤔 

I guess it depends on how we define a 'form' object?

mattwills-moj-digital (dismissed):

server/@types/forms/index.d.ts
nathanrussell-moj-digital: Yeah; naming is hard! and I'm defo open to a better name than `NewGoal` !

Even though its in the `compositeForms` module, it felt "wrong" to call it a "Form" because its not a form in its own right. IE. there are no html forms or request handlers that bind to the properties of this object. Its just a convenient container that contains other form objects.

And yes, naming it plural (`NewGoalForms`) would be really confusing when we start using it in an array! 😁 
Currently its a single instance on the session:
```
interface SessionData {
    newGoalForm: NewGoal
```
(hmmm - maybe the property name on the session is wrong, and it should just be `newGoal` ?

If we called the type `NewGoalForm` it would kinda work (but I dont like it because its not a form! 🤣 )
```
interface SessionData {
    newGoalForm: NewGoalForm
```

If we called the type `NewGoalForms` it would start looking wrong ....
```
interface SessionData {
    newGoalForm: NewGoalForms
```

And then once it becomes an array ....
```
interface SessionData {
    newGoalForms: Array<NewGoalForms>
```

Naming is hard! 🤣

server/routes/createGoal/createGoalController.test.ts
nathanrussell-moj-digital: I think you're right - `req.session.newGoal` .... I'll change it 👍

mattwills-moj-digital (approved):

PR #166, opened 2023-09-13
RR-144 implement the DPS header with fallback and config for environment names

chrimesdev:
## Description

- Added the fallbacks for the 3 scenarios and the relevant markup/styles to display the correct header
- Implemented the `ENVIRONMENT_NAME` variable needed for the Header fallback https://github.com/ministryofjustice/hmpps-micro-frontend-components/blob/main/readme/incorporating.md#fallbacks-services-with-prison-only-users


server/app.ts
```diff
@@ -34,6 +35,11 @@ export default function createApp(services: Services): express.Application {
   app.use(setUpWebSession())
   app.use(setUpWebRequestParsing())
   app.use(setUpStaticResources())
+
+  if (config.featureToggles.frontendComponentsApiToggleEnabled) {
+    setUpEnvironmentName(app)
```
chrimesdev, line 40: This is for the Header fallback https://github.com/ministryofjustice/hmpps-micro-frontend-components/blob/main/readme/incorporating.md#fallbacks-services-with-prison-only-users

server/views/partials/header.njk
```diff
@@ -1,4 +1,4 @@
-<header class="hmpps-header" role="banner">
+<header class="app-header" role="banner">
```
chrimesdev, line 1: As discussed, there are conflicting styles with the DPS fallback header and our original header. `app-` is the usual convention used when its a local application component/style.

server/app.ts
chrimesdev: This needs to become before the `setupNunjucks` otherwise I would have grouped it with the other `if` statement block

feature.env
```diff
@@ -10,7 +10,6 @@ CIAG_INDUCTION_API_URL=http://localhost:9091
 DPS_URL=https://digital-dev.prison.service.justice.gov.uk
 
 STUB_PRISONER_LIST_PAGE_ENABLED=true
-FRONTEND_COMPONENTS_API_FEATURE_TOGGLE_ENABLED=false
```
chrimesdev, line 13: Removed this for the time being as we're not actually using it yet.

nathanrussell-moj-digital: No, I think we want this back in.
I accept its not used yet .... but I think if we finish the footer story before getting too far ahead ourselves with header, then it will be a bit more cohesive.
Specifically I'm thinking we need to finish RR-143 by doing RR-316 (cypress tests and wiremock stubs) first.

chrimesdev: Will look at RR-316 now

chrimesdev: Back in as part of RR-136 👀

server/app.ts
nathanrussell-moj-digital: I reckon we can do with without the if block, as its only setting a couple of locals - its not actually functional per-se, so probably doesn't matter if we do it even if the feature toggle is disabled?

Also, I'm not sure a function in the middleware folder is adding much (especially as I don't think its strictly speaking a middleware function?

Could we replace what you have here:
```
  if (config.featureToggles.frontendComponentsApiToggleEnabled) {
    setUpEnvironmentName(app)
  }
```

with just:
```
  app.locals.environmentName = config.environmentName
  app.locals.environmentNameColour = config.environmentName === 'PRE-PRODUCTION' ? 'govuk-tag--green' : ''
```  

And then delete the middleware function ?
What do you think?

server/middleware/setUpEnvironmentName.ts
```diff
@@ -0,0 +1,8 @@
+/* eslint-disable no-param-reassign */
+import express from 'express'
+import config from '../config'
+
+export default (app: express.Express) => {
+  app.locals.environmentName = config.environmentName
+  app.locals.environmentNameColour = config.environmentName === 'PRE-PRODUCTION' ? 'govuk-tag--green' : ''
```
nathanrussell-moj-digital, line 7: Is this based on their instructions?
I'm not sure this works because it'll only use the green tag style if `config.environmentName === 'PRE-PRODUCTION'` .... but in `values-preprod.yml` it is set to `PREPROD`
So I dont think it'll ever get the green class ?

chrimesdev: Good spot - it is from the instructions specifically telling you to copy this file https://github.com/ministryofjustice/hmpps-micro-frontend-components/blob/main/_fallbacks/setUpEnvironmentName.ts.txt

I'll take a look and maybe open a PR/issue against their repo!

server/app.ts
chrimesdev: Yep that makes sense, will just have to remember this if they make any further changes to the fallback!

chrimesdev:
Tests are fixed after adding the correct Header markup in the wiremocks @nathanrussell-moj-digital also addressed the other comments

integration_tests/e2e/login.cy.ts
```diff
@@ -65,6 +65,6 @@ context('SignIn', () => {
     cy.task('stubAuthUser', 'bobby brown')
     cy.signIn()
 
-    indexPage.headerUserName().contains('B. Brown')
+    indexPage.headerUserName().contains('J. Smith')
```
nathanrussell-moj-digital, line 68: Hmmm, we might want to look at this .... how does this work? If we sign in as Bobby Brown (line 65), how/why does the header say "J. Smith" .... hmmmm ....
I'll checkout your branch and take a look ....

nathanrussell-moj-digital: Ah, I see, its the wiremock stub .... we might be able to do something there , let me take a look ....

integration_tests/mockApis/frontendComponentApi.ts
```diff
@@ -11,7 +11,54 @@ const stubGetFooterComponent = (): SuperAgentRequest =>
...
+        html: `
+          <footer class="connect-dps-common-footer" role="contentinfo">
```
nathanrussell-moj-digital, line 15: Multi-lilne text in backticks FTW 👍 😁

integration_tests/mockApis/frontendComponentApi.ts
```diff
@@ -30,7 +77,7 @@ const stubGetFooterComponent500error = (): SuperAgentRequest =>
     },
   })
 
-const stubGetHeaderComponent = (): SuperAgentRequest =>
+const stubGetHeaderComponent = (name = 'J. Smith'): SuperAgentRequest =>
```
nathanrussell-moj-digital, line 80: I've given the stub a default name parameter that will be used in the stubbed markup on line 123 below 👇 
And then tests can override it if they need to (as we are doing in the login.cy.ts test above ☝️ )

nathanrussell-moj-digital (dismissed):
LGTM 👍

nathanrussell-moj-digital (approved):
LGTM 👍

PR #167, opened 2023-09-13
RR-303 make the view note link black on print

chrimesdev:
The “View note” link should be black so that it doesn’t use colour ink, usually we would hide links all together as they're not clickable on paper. But I think in this instance it needs the View note as a heading for context.

Screenshot 2023-09-13 at 18 09 25

### DIFF
--- a/assets/scss/print.scss
+++ b/assets/scss/print.scss
@@ -5,7 +5,8 @@
 
 .govuk-hint,
 .govuk-tag,
-.govuk-caption-l {
+.govuk-caption-l,
+.govuk-details__summary {
   color: #0b0c0c
 }
 
--- a/assets/scss/print.scss
+++ b/assets/scss/print.scss
@@ -1,3 +1,6 @@
+@import "govuk/base";
+@import 'govuk/core/all';
+
 @page {
   size: auto;
   margin: 8mm 10mm;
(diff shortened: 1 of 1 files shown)


assets/scss/print.scss
```diff
@@ -5,7 +5,8 @@
 
 .govuk-hint,
 .govuk-tag,
-.govuk-caption-l {
+.govuk-caption-l,
+.govuk-details__summary {
   color: #0b0c0c
```
nathanrussell-moj-digital, line 10: LGTM; out of interest, could you use the variable`$govuk-text-colour` instead of the hex reference?
https://design-system.service.gov.uk/styles/colour/

nathanrussell-moj-digital (dismissed):
1 very minor comment/question; but otherwise LGTM 👍

mattwills-moj-digital (approved):

assets/scss/print.scss
chrimesdev: We didn't previously have access to the colour variables in the `print.scss`, I've included the relevant imports we need from `govuk` (not all components styles just the base/core) to allow us to use variables here.

chrimesdev:
This will need re-approval (if happy) as I pushed to the branch and its dismissed reviews.

nathanrussell-moj-digital (approved):
LGTM 👍

PR #170, opened 2023-09-14
RR-143 - Improve error handling; patterns consistent with our other services

nathanrussell-moj-digital:
This PR adds error handling and a mapper to the frontendComponentsService, consistent with our other services

### DIFF
--- a/server/@types/frontendComponentApi/index.d.ts
+++ b/server/@types/frontendComponentApi/index.d.ts
@@ -1,5 +1,9 @@
-export default interface Component {
-  html: string
-  css: string[]
-  javascript: string[]
+export interface components {
+  schemas: {
+    Component: {
+      html: string
+      css: string[]
+      javascript: string[]
+    }
+  }
 }
--- a/server/@types/viewModels/index.d.ts
+++ b/server/@types/viewModels/index.d.ts
@@ -242,8 +242,10 @@ declare module 'viewModels' {
    * DPS frontend components
    */
   export interface FrontendComponentsPageAdditions {
+    headerHtml: string
     footerHtml: string
     cssIncludes: Array<string>
     jsIncludes: Array<string>
+    problemRetrievingData: boolean
   }
 }
--- a/server/middleware/fetchFrontendComponentMiddleware.ts
+++ b/server/middleware/fetchFrontendComponentMiddleware.ts
@@ -1,21 +1,9 @@
 import { RequestHandler } from 'express'
-import type { FrontendComponentsPageAdditions } from 'viewModels'
-import logger from '../../logger'
 import { Services } from '../services'
 
 export default function getFrontendComponents({ frontendComponentService }: Services): RequestHandler {
   return async (req, res, next) => {
-    try {
-      const footer = await frontendComponentService.getComponents('footer', res.locals.user.token)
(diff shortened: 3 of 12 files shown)


server/@types/frontendComponentApi/index.d.ts
```diff
@@ -1,5 +1,9 @@
-export default interface Component {
-  html: string
-  css: string[]
-  javascript: string[]
+export interface components {
```
nathanrussell-moj-digital, line 1: Not strictly related to error handling; and what was there previously worked. I've changed it however so "structurally" it looks like a generated module/type (ie. consistent with the other modules and types that we generate from swagger specs)
The reason I made this change is a) for consistency, and also so that as and when we do import the frontend components API swagger spec (if they ever make one available), the scope of change will be minimal 👍

server/@types/viewModels/index.d.ts
```diff
@@ -242,8 +242,10 @@ declare module 'viewModels' {
    * DPS frontend components
    */
   export interface FrontendComponentsPageAdditions {
+    headerHtml: string
```
nathanrussell-moj-digital, line 245: I know this story is not related to the header (we have a different story for that), but because of the changes I made to the service and mapper (see below) it made sense to add it

Also added the problemRetrievingData property, consistent with our other types like this

server/middleware/fetchFrontendComponentMiddleware.ts
```diff
@@ -1,21 +1,9 @@
...
-        cssIncludes: [...footer.css],
-        jsIncludes: [...footer.javascript],
-      } as FrontendComponentsPageAdditions
-      next()
-    } catch (error) {
-      logger.error(error, 'Failed to retrieve front end components')
-      next()
-    }
+    res.locals.feComponents = await frontendComponentService.getComponents(res.locals.user.token)
+    next()
```
nathanrussell-moj-digital, line 7: The middleware is now much much simpler 😁 as it does not need to worry about error handling (as the service does that) and it does not need to worry about mapping (again, the service does that)

server/services/frontendComponentService.ts
```diff
@@ -1,17 +1,26 @@
...
+import type { FrontendComponentsPageAdditions } from 'viewModels'
 import { FrontendComponentApiClient } from '../data'
 import logger from '../../logger'
+import toFrontendComponentsPageAdditions from '../data/mappers/frontendComponentMapper'
 
 export default class FrontendComponentService {
   constructor(private readonly frontendApiClient: FrontendComponentApiClient) {}
 
-  async getComponents(component: 'header' | 'footer', userToken: string): Promise<FrontendComponent> {
+  async getComponents(userToken: string): Promise<FrontendComponentsPageAdditions> {
```
nathanrussell-moj-digital, line 9: This is the main refactoring in this PR in respect of design change ....
The service is now consistent with our other services in that it gets data from an API, maps it into a view model, and returns it. If there is an error it returns the declared type, but with the field problemRetrievingData set to true. This is what all our services do 👍 

The specifics of this service method are that it actually calls the frontend components API twice - once to get the header, once to get the footer; then maps both responses into the returned FrontendComponentsPageAdditions object

chrimesdev (approved):

PR #173, opened 2023-09-15
RR-315 - Refactor LDD & Health Needs panel based on new design

nathanrussell-moj-digital:
This PR restructures the LDD & Health needs panel on the Support Needs page as per the new design (note - this PR only address the LDD & Health needs panel - the next PR will do the same for Neurodiversity).

### The current originally implemented design
Screenshot 2023-09-15 at 08 20 04

### The new design from the Jira
Screenshot 2023-09-12 at 15 11 31

### The new design as implemented
Screenshot 2023-09-15 at 08 21 12

### DIFF
--- a/server/@types/viewModels/index.d.ts
+++ b/server/@types/viewModels/index.d.ts
@@ -18,8 +18,10 @@ declare module 'viewModels' {
   export interface HealthAndSupportNeeds {
     prisonId: string
     prisonName: string
-    languageSupportNeeded?: string
-    lddAndHealthNeeds: Array<string>
+    rapidAssessmentDate?: Date
+    inDepthAssessmentDate?: Date
+    primaryLddAndHealthNeeds?: string
+    additionalLddAndHealthNeeds: Array<string>
   }
 
   export interface Neurodiversity {
--- a/server/filters/fallbackMessageFilter.test.ts
+++ b/server/filters/fallbackMessageFilter.test.ts
@@ -1,0 +1,28 @@
+import fallbackMessageFilter from './fallbackMessageFilter'
+
+describe('fallbackMessageFilter', () => {
+  it('should return value given value', () => {
+    // Given
+    const value = 'some value'
+    const fallBackValue = 'some alternative fallback value'
+
+    // When
+    const actual = fallbackMessageFilter(value, fallBackValue)
+
+    // Then
+    expect(actual).toEqual(value)
+  })
+
+  Array.of(undefined, null, '', '    ').forEach(value => {
+    it(`should return fallback value given '${value}'`, () => {
+      // Given
+      const fallBackValue = 'some alternative fallback value'
+
+      // When
+      const actual = fallbackMessageFilter(value, fallBackValue)
+
+      // Then
+      expect(actual).toEqual(fallBackValue)
+    })
+  })
+})
--- a/server/views/pages/overview/partials/supportNeedsTabContents.njk
+++ b/server/views/pages/overview/partials/supportNeedsTabContents.njk
@@ -5,40 +5,58 @@
 
       <div class="govuk-summary-card" id="health-and-support-needs-summary-card">
         <div class="govuk-summary-card__title-wrapper">
-          <h2 class="govuk-summary-card__title">Health and support needs</h2>
+          <h2 class="govuk-summary-card__title">LDD and health needs</h2>
         </div>
         <div class="govuk-summary-card__content">
-          <dl class="govuk-summary-list">
+          <p class="govuk-hint">Information from Curious.</p>
 
-            <div class="govuk-summary-list__row">
-              <dt class="govuk-summary-list__key">
-                Language support needed
-              </dt>
-              <dd class="govuk-summary-list__value">
-                <ul class="govuk-list">
-                  {% for healthAndSupportNeed in supportNeeds.healthAndSupportNeeds %}
-                    <li>{{ healthAndSupportNeed.languageSupportNeeded }}</li>
-                  {% endfor %}
-                </ul>
-              </dd>
-            </div>
+          {% for healthAndSupportNeed in supportNeeds.healthAndSupportNeeds %}
 
-            <div class="govuk-summary-list__row">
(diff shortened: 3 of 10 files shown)


server/@types/viewModels/index.d.ts
```diff
@@ -18,8 +18,10 @@ declare module 'viewModels' {
   export interface HealthAndSupportNeeds {
     prisonId: string
     prisonName: string
-    languageSupportNeeded?: string
-    lddAndHealthNeeds: Array<string>
+    rapidAssessmentDate?: Date
```
nathanrussell-moj-digital, line 21: The view model type has changed because, well, the view has changed 😁

server/filters/fallbackMessageFilter.test.ts
```diff
@@ -0,0 +1,28 @@
+import fallbackMessageFilter from './fallbackMessageFilter'
```
nathanrussell-moj-digital, line 1: I wrote a "fallback message filter", whose job is to render some fallback message if the main value is null/blank etc.
You can see how it's used further down in the nunjucks template, but it means we don't need to put if statements or logic in the template 👍 , and we can test the filter once in isolation here 👍

server/views/pages/overview/partials/supportNeedsTabContents.njk
```diff
@@ -5,40 +5,58 @@
...
-              </dd>
-            </div>
+            <dl class="govuk-summary-list">
+              <h3>{{ healthAndSupportNeed.prisonName }}</h3>
+              <div class="govuk-summary-list__row">
+                <dt class="govuk-summary-list__key">
+                  Rapid assessment
+                </dt>
+                <dd class="govuk-summary-list__value">
+                  {{ healthAndSupportNeed.rapidAssessmentDate | formatDate('D MMMM YYYY') | fallbackMessage('Not recorded in Curious') }}
```
nathanrussell-moj-digital, line 22: This is how we use the new filter. Nice, no if blocks or logic 👍 
(apart from 1 - see below)

server/views/pages/overview/partials/supportNeedsTabContents.njk
```diff
@@ -5,40 +5,58 @@
...
+                <dd class="govuk-summary-list__value">
+                  {{ healthAndSupportNeed.primaryLddAndHealthNeeds | fallbackMessage('Not recorded in Curious') }}
+                </dd>
+              </div>
+              <div class="govuk-summary-list__row">
+                <dt class="govuk-summary-list__key">
+                  Additional LDD and health needs
+                </dt>
+                <dd class="govuk-summary-list__value">
+                  {% if healthAndSupportNeed.additionalLddAndHealthNeeds.length > 0 %}
```
nathanrussell-moj-digital, line 46: Unfortunately the filter I wrote only works for single string values. The property here `additionalLddAndHealthNeeds` is an array of string, and we need to test if it is empty of not.
For one or two use cases like this I think an if block is probably OK 🤷‍♂️

server/views/pages/overview/partials/supportNeedsTabContents.test.ts
```diff
@@ -27,17 +27,41 @@ describe('Support Needs tab view', () => {
     const $ = cheerio.load(compiledTemplate.render(viewContext))
 
     // Then
```
nathanrussell-moj-digital, line 29: This test has had to change quite a lot, because the markup in the nunjucks template has changed a lot.
I'm also found this syntax which I think reads quite well:
```
    expect(
      healthAndSupportNeedsCard //
        .find(`.govuk-summary-list__key:contains('Rapid assessment')`)
        .next()
        .text()
        .trim(),
    ).toEqual('18 February 2022')
```
IE. assert the value of the Radid Assessment dt/dd is 18 February 2022

server/views/pages/overview/partials/supportNeedsTabContents.njk
chrimesdev: Smart!

server/views/pages/overview/partials/supportNeedsTabContents.njk
```diff
@@ -5,40 +5,58 @@
...
-                  <ul class="govuk-list">
-                    {% for lddHealthNeed in healthAndSupportNeed.lddAndHealthNeeds %}
-                      <li>{{ lddHealthNeed }}</li>
-                    {% endfor %}
-                  </ul>
-                {% endfor %}
-              </dd>
-            </div>
+            <dl class="govuk-summary-list">
+              <h3>{{ healthAndSupportNeed.prisonName }}</h3>
```
chrimesdev, line 16: Was wondering if you could stick `<h3>` in a description list, but apparently you can!

chrimesdev, line 16: Minor but..

```
Error: Element [h3](https://html.spec.whatwg.org/multipage/#the-h1,-h2,-h3,-h4,-h5,-and-h6-elements) not allowed as child of element [dl](https://html.spec.whatwg.org/multipage/#the-dl-element) in this context. (Suppressing further errors from this subtree.)
```

mattwills-moj-digital (approved):
GS

server/views/pages/overview/partials/supportNeedsTabContents.njk
chrimesdev: Infact we could just move the `<h3>` above the `<dl>` and should be good?

nathanrussell-moj-digital: Ah, OK, so strictly speaking my markup is not quite right?
Let me see what I can do ...

nathanrussell-moj-digital: @chrimesdev - I've moved the h3 out of the dl (and also added some visually hidden text to aid accessibility)
Have a look and see what you think 🤞

chrimesdev:
LGTM!

chrimesdev (approved):

PR #178, opened 2023-09-18
RR-152 - Controller request handling for add another goal

nathanrussell-moj-digital:


server/@types/forms/index.d.ts
```diff
@@ -15,6 +15,10 @@ declare module 'forms' {
     note?: string
   }
 
+  export interface ReviewGoalForm {
```
nathanrussell-moj-digital, line 18: Basic form type with just an action field - because the "Review Your Goals" page submits the form (POST) to go to the next screen - I've now added an action so we can tell if the user has clicked to Add Another Goal, or submit the form to save the action plan (current behaviour)

server/routes/createGoal/createGoalController.test.ts
```diff
@@ -353,6 +353,10 @@ describe('createGoalController', () => {
         },
       ] as Array<NewGoal>
 
+      req.body = {
```
nathanrussell-moj-digital, line 356: Existing unit tests for the `submitReviewGoal` controller method now need the `action: 'submit-form'` so that the contoller method follows the correct path / behaviour

server/routes/createGoal/createGoalController.test.ts
```diff
@@ -420,6 +429,44 @@ describe('createGoalController', () => {
         expectedPrisonId,
       )
       expect(next).toHaveBeenCalledWith(expectedError)
+      expect(req.session.newGoals).not.toBeUndefined()
+    })
+
+    it('should redirect to create goal form given action is add-another-goal', async () => {
```
nathanrussell-moj-digital, line 435: New test to test the `action: 'add-another-goal'` flow

server/routes/createGoal/createGoalController.ts
```diff
@@ -118,6 +118,14 @@ export default class CreateGoalController {
     const { prisonerSummary } = req.session
     const { prisonId } = prisonerSummary
 
+    const reviewGoalForm = { ...req.body }
```
nathanrussell-moj-digital, line 121: This is the `submitReviewGoal` controller method
It now looks at the request body (`ReviewGoalForm`) and the `action` field within it - if it is `add-another-goal` then we reset the `NewGoal` session variable (singular, not the plural array), and redirect back to the start of the journey (ie. the 1st Create Goal screen)

chrimesdev (approved):

PR #180, opened 2023-09-19
RR-328 Cypress test add scenario for creating 2 goals as part of journey

chrimesdev:
## Description

Cypress test add scenario for creating 2 goals as part of journey

https://dsdmoj.atlassian.net/browse/RR-328

### DIFF
--- a/integration_tests/e2e/goal/review.cy.ts
+++ b/integration_tests/e2e/goal/review.cy.ts
@@ -3,6 +3,7 @@ import AddStepPage from '../../pages/goal/AddStepPage'
 import AddNotePage from '../../pages/goal/AddNotePage'
 import ReviewPage from '../../pages/goal/ReviewPage'
 import OverviewPage from '../../pages/overview/OverviewPage'
+import CreateGoalPage from '../../pages/goal/CreateGoalPage'
 
 context('Review goal(s)', () => {
   beforeEach(() => {
@@ -101,38 +102,46 @@ context('Review goal(s)', () => {
     cy.visit(`/plan/${prisonNumber}/view/overview`)
     const overviewPage = Page.verifyOnPage(OverviewPage)
 
-    const createGoalPage = overviewPage.clickAddGoalButton()
+    let createGoalPage = overviewPage.clickAddGoalButton()
     createGoalPage //
       .setGoalTitle('Learn French')
       .submitPage()
 
-    const addStepPage = Page.verifyOnPage(AddStepPage)
+    let addStepPage = Page.verifyOnPage(AddStepPage)
     addStepPage //
       .setStepTitle('Book French course')
       .setStepTargetDateRange('ZERO_TO_THREE_MONTHS')
       .submitPage()
 
-    const addNotePage = Page.verifyOnPage(AddNotePage)
+    let addNotePage = Page.verifyOnPage(AddNotePage)
     addNotePage.setNote("Pay close attention to Chris' behaviour during classes")
     addNotePage.submitPage()
 
-    // When
     const reviewPage = Page.verifyOnPage(ReviewPage)
+    cy.get('.govuk-summary-card').should('have.length', 1)
+    cy.get('.govuk-summary-card').eq(0).contains('Learn French')
+
+    // When
     reviewPage.addAnotherGoal()
 
+    createGoalPage = Page.verifyOnPage(CreateGoalPage)
     createGoalPage //
       .setGoalTitle('Learn Spanish')
       .submitPage()
 
+    addStepPage = Page.verifyOnPage(AddStepPage)
     addStepPage //
       .setStepTitle('Book Spanish course')
       .setStepTargetDateRange('ZERO_TO_THREE_MONTHS')
       .submitPage()
 
+    addNotePage = Page.verifyOnPage(AddNotePage)
     addNotePage.submitPage()
 
     // Then
     Page.verifyOnPage(ReviewPage)
+    cy.get('.govuk-summary-card').should('have.length', 2)
+    cy.get('.govuk-summary-card').eq(1).contains('Learn Spanish')
   })
 
   it('should move to Overview page', () => {
--- a/integration_tests/e2e/goal/review.cy.ts
+++ b/integration_tests/e2e/goal/review.cy.ts
@@ -94,6 +94,47 @@ context('Review goal(s)', () => {
     Page.verifyOnPage(AddNotePage)
   })
 
+  it('should be able to create another goal from the Review screen', () => {
+    // Given
+    const prisonNumber = 'G6115VJ'
+    cy.signIn()
+    cy.visit(`/plan/${prisonNumber}/view/overview`)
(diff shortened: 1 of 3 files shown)


server/routes/createGoal/reviewView.ts
```diff
@@ -4,7 +4,7 @@ import type { CreateGoalDto } from 'dto'
 export default class ReviewView {
   constructor(
     private readonly prisonerSummary: PrisonerSummary,
-    private readonly createGoalDto: Array<CreateGoalDto>,
```
chrimesdev, line 7: Following feedback from the other PR on using plural for dtos

server/routes/createGoal/reviewView.ts
```diff
@@ -4,7 +4,7 @@ import type { CreateGoalDto } from 'dto'
...
+    private readonly createGoalDtos: Array<CreateGoalDto>,
```
chrimesdev, line 7: Following feedback on other PR

integration_tests/e2e/goal/review.cy.ts
```diff
@@ -94,6 +94,47 @@ context('Review goal(s)', () => {
     Page.verifyOnPage(AddNotePage)
   })
 
+  it('should be able to create another goal from the Review screen', () => {
```
nathanrussell-moj-digital, line 97: I love the test scenario name - perfect 👍

integration_tests/e2e/goal/review.cy.ts
```diff
@@ -94,6 +94,47 @@ context('Review goal(s)', () => {
...
+    // Given
+    const prisonNumber = 'G6115VJ'
+    cy.signIn()
+    cy.visit(`/plan/${prisonNumber}/view/overview`)
+    const overviewPage = Page.verifyOnPage(OverviewPage)
```
nathanrussell-moj-digital, line 102: All this pre-amble in the // Given block is getting a bit long and repetitive - it's not your fault, and we don't have any better way of doing it at the moment, so don't worry; it's just be thinking out loud ....

I reckon what we need (NOT for this PR, we'll do it in a separate ticket) is a single cypress task or method called something like `prepareNewGoalAndArriveOnReviewGoalScreenFromOverviewScreen` (don't shoot me for the method name - I suspect we can do better 😁 ), and then we can simply have a test that does
```
It('should .....', () => {
  // Given
  cy.task('prepareNewGoalAndArriveOnReviewGoalScreenFromOverviewScreen')
  const reviewPage = Page.verifyOnPage(ReviewPage)

  // When
  ....
```

Anyway, don't worry about it for now - I'm just thinking out load ....

integration_tests/e2e/goal/review.cy.ts
```diff
@@ -94,6 +94,47 @@ context('Review goal(s)', () => {
...
+    createGoalPage //
+      .setGoalTitle('Learn Spanish')
+      .submitPage()
+
+    addStepPage //
+      .setStepTitle('Book Spanish course')
+      .setStepTargetDateRange('ZERO_TO_THREE_MONTHS')
+      .submitPage()
+
+    addNotePage.submitPage()
```
nathanrussell-moj-digital, line 132: This test is good in that it exercises the navigation of clicking the add another goal button and going back round the journey again, but I think there's a few things we can do to improve it, specifically around what we are asserting.

I think it would be good to assert that once we are on the ReviewGoal page for the first time (ie. at the end of the // Given section and before the // When section) we assert that we are currently displaying 1 goal. Maybe we could also assert the title of the goal (perhaps?), but at a minimum I think we should assert we have added and are displaying 1 goal.

Once we've then added the 2nd goal and have arrived back on the ReviewGoal page (ie. your // Then block) I think we need to assert we are now displaying 2 goals (along with their titles if you chose to do that above)

I also think that the page variables you are using in the // When block (createGoalPage, addStepPage, addNotePage etc) are the object references from the // Given block and possibly might not reference the "2nd display" of that page (if that makes sense). I think you'd be better off re-assigning them, and using a `let` rather than a `const`. Something like this:

```
    let createGoalPage = overviewPage.clickAddGoalButton()

    let addStepPage = createGoalPage //
      .setGoalTitle('Learn French')
      .submitPage()

    let addNotePage = addStepPage //
      .setStepTitle('Book French course')
      .setStepTargetDateRange('ZERO_TO_THREE_MONTHS')
      .submitPage()

    let reviewPage = addNotePage // 
      .setNote("Pay close attention to Chris' behaviour during classes")
      .submitPage()

    // assert here that there is 1 goal on the review page, optionally with its goal title

    // When
    reviewPage.addAnotherGoal()

    createGoalPage = Page.verifyOnPage(CreateGoalPage)
    addStepPage = createGoalPage //
      .setGoalTitle('Learn Spanish')
      .submitPage()

    addNotePage = addStepPage //
      .setStepTitle('Book Spanish course')
      .setStepTargetDateRange('ZERO_TO_THREE_MONTHS')
      .submitPage()

    reviewPage = addNotePage.submitPage()

    // Then
    // assert here that there are 2 goals on the review page, optionally with their goal titles  
})
```

integration_tests/e2e/goal/review.cy.ts
mattwills-moj-digital: Would `prepareReviewGoalScreen`, which creates one goal by default, suffice?

chrimesdev: Aye I was thinking it was getting quite long winded, I think the approach you suggested is a good one and as the service grows I imagine we'll need more of these resuable tasks!

integration_tests/e2e/goal/review.cy.ts
```diff
@@ -94,6 +95,55 @@ context('Review goal(s)', () => {
...
+      .setStepTitle('Book French course')
+      .setStepTargetDateRange('ZERO_TO_THREE_MONTHS')
+      .submitPage()
+
+    let addNotePage = Page.verifyOnPage(AddNotePage)
+    addNotePage.setNote("Pay close attention to Chris' behaviour during classes")
+    addNotePage.submitPage()
+
+    const reviewPage = Page.verifyOnPage(ReviewPage)
+    cy.get('.govuk-summary-card').should('have.length', 1)
```
nathanrussell-moj-digital, line 121: This will do for now (we need to get this merged), but our general pattern is that the test itself knows very little about the underlying page driver technology, and that apart from the early `cy.task` type calls, we dont use `cy` in the test class itself.
The problem is that it couples the test too tightly to the page driver - what if we were to change from cypress to selenium or a.n.other web/page driver technology.
This is why we have the `Page` classes, that abstract the page driver implementation away from the test itself.
So really the `cy.get('.govuk-summary-card').should('have.length', 1)` should be in a suitably named method in `ReviewPage` - something like `hasNumberOfGoalsDisplayed(expected: number)`, which in turn calls the `cy` elements

nathanrussell-moj-digital (approved):
Looks good, We can improve the assertions over time as described in my comment, but lets get this merged 👍

PR #181, opened 2023-09-20
RR-143 - Changed frontend components URL to enable testing of fallbacks

nathanrussell-moj-digital:


chrimesdev (approved):

mattwills-moj-digital (approved):
🤷‍♂️ 🤯

nathanrussell-moj-digital:
> 🤷‍♂️ 🤯

Lol - its just changing the URL of the front end components API to a non-existent URL, so that when we try to get the header and footer it 404's, which in turn forces us to display the fallback header and footer

mattwills-moj-digital:
> > 🤷‍♂️ 🤯
> 
> Lol - its just changing the URL of the front end components API to a non-existent URL, so that when we try to get the header and footer it 404's, which in turn forces us to display the fallback header and footer

I guessed as much really 😃 😉

PR #183, opened 2023-09-20
RR-143 - Updated local wiremock stubs with header and footer assets

nathanrussell-moj-digital:
This PR updates the local wiremock stubs with the assets for the DPS header and footer, to better facilitate running the UI locally with stubbed API calls (the DPS header and footer are now served from API calls)

We already had the basic wiremock stubs for the header and footer markup, but they were missing the stylesheets and fonts; which is what this PR adds.

I've also updated the readme, specifically with some pointers as to how you can introduce delays or other error conditions to simulate network or API latency (useful for demo's)


docker-compose-test.yml
```diff
@@ -14,7 +14,7 @@ services:
     - hmpps_int
     container_name: wiremock
     restart: always
-    command: "--global-response-templating --verbose"
+    command: "--global-response-templating --enable-stub-cors" # add command `--verbose` to output verbose request matching
```
nathanrussell-moj-digital, line 17: Because of the way some of the css and fonts are served via wiremock (in relation to how the page itself is served) there is a CORS issue, so needed to enable CORS headers on wiremock stubs
Also removed `--verbose` as it was getting very very noisy!

wiremock/__files/assets/stylesheets/footer.css
```diff
@@ -0,0 +1,181 @@
+.govuk-width-container {
```
nathanrussell-moj-digital, line 1: Copy and paste verbatim from the css that is served by the frontend components API

wiremock/mappings/062-dps-frontend-assets-mapping.json
```diff
@@ -0,0 +1,14 @@
+{
```
nathanrussell-moj-digital, line 1: We need this mapping (over and above the actual files in `__files`) to add the cross origin header

chrimesdev (approved):

PR #184, opened 2023-09-21
RR-339 hide the new DPS footer on print

chrimesdev:
## Description

Hide the new DPS footer on print

We can't pass in CSS classes like we usually do (`govuk-!-display-none-print`) so we need to use the Sass mixin `@include govuk-visually-hidden();` in the print stylesheet instead.

Before:

Screenshot 2023-09-21 at 09 54 20

After:

Hidden (not much to screenshot!)

### DIFF
--- a/assets/scss/print.scss
+++ b/assets/scss/print.scss
@@ -30,3 +30,7 @@
   break-inside: avoid;
   margin-top: 8.5mm;
 }
+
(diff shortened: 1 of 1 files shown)


nathanrussell-moj-digital (approved):
LGTM 👍

PR #185, opened 2023-09-21
RR-331 trim whitespace from prisoner names and use a shared breadcrumb

chrimesdev:
## Description

Trim the whitespace from the Prisoners first and last name.

There was also an opportunity to reduce duplication by creating a shared Breadcrumb partial for a number of pages, so I have made the change so that don't repeat ourselves across a number of pages.

https://dsdmoj.atlassian.net/browse/RR-331

### DIFF
--- a/server/routes/routerRequestHandlers.test.ts
+++ b/server/routes/routerRequestHandlers.test.ts
@@ -462,7 +462,7 @@ describe('routerRequestHandlers', () => {
       req.params.prisonNumber = prisonNumber
       const prisoner = {
         firstName: ' jimmy ',
-        lastName: ' lightfingers',
+        lastName: ' LIGHTFINGERS',
       } as Prisoner
       prisonerSearchService.getPrisonerByPrisonNumber.mockResolvedValue(prisoner)
 
--- a/server/routes/routerRequestHandlers.test.ts
+++ b/server/routes/routerRequestHandlers.test.ts
@@ -451,6 +451,34 @@ describe('routerRequestHandlers', () => {
       expect(next).toHaveBeenCalled()
     })
 
+    it('should return prisoners first name and last name without whitespace and a capital letter at the start', async () => {
+      // Given
+      const username = 'a-dps-user'
+      req.user.username = username
+
+      req.session.prisonerSummary = undefined
+
+      const prisonNumber = 'A1234GC'
+      req.params.prisonNumber = prisonNumber
+      const prisoner = {
+        firstName: ' jimmy ',
+        lastName: ' lightfingers',
+      } as Prisoner
+      prisonerSearchService.getPrisonerByPrisonNumber.mockResolvedValue(prisoner)
+
+      const expectedPrisonerFirstName = 'Jimmy'
+      const expectedPrisonerLastName = 'Lightfingers'
+
+      // When
+      await requestHandler(req as undefined as Request, res as undefined as Response, next as undefined as NextFunction)
+
+      // Then
+      expect(prisonerSearchService.getPrisonerByPrisonNumber).toHaveBeenCalledWith(prisonNumber, username)
+      expect(req.session.prisonerSummary.firstName).toEqual(expectedPrisonerFirstName)
+      expect(req.session.prisonerSummary.lastName).toEqual(expectedPrisonerLastName)
+      expect(next).toHaveBeenCalled()
+    })
+
     it('should not retrieve prisoner given prisoner already in session', async () => {
       // Given
       const username = 'a-dps-user'
--- a/server/routes/routerRequestHandlers.ts
+++ b/server/routes/routerRequestHandlers.ts
@@ -112,6 +112,12 @@ const checkNewGoalsFormExistsInSession = async (req: Request, res: Response, nex
  *  Middleware function that returns a Request handler function to look up the prisoner from prisoner-search, map to a PrisonerSummary, and store in the session
  */
 const retrievePrisonerSummaryIfNotInSession = (prisonerSearchService: PrisonerSearchService): RequestHandler => {
+  // Trim whitespace from a name string and capitalize the first letter and lowercase the rest of the string
+  const capitalize = (name: string): string => {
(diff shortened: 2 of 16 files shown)


server/views/pages/functionalSkills/index.njk
```diff
@@ -33,7 +17,7 @@
     <div class="govuk-grid-row">
       <div class="govuk-grid-column-two-thirds">
         <span class="govuk-caption-l">Learning and work progress</span>
-        <h1 class="govuk-heading-l">{{ prisonerSummary.firstName | capitalize }} {{ prisonerSummary.lastName | capitalize }}'s functional skills</h1>
+        <h1 class="govuk-heading-l">{{ prisonerSummary.firstName | trim | capitalize }} {{ prisonerSummary.lastName | trim | capitalize }}'s functional skills</h1>
```
nathanrussell-moj-digital, line 20: Hmm, this is not quite what I had in mind. The problem with this approach is that we're repeating the trim operation all over the place; plus we're making it a view responsibility rather than a "data" type concern. To me it feels like a problem of the data, rather than the view?

I think in the Jira I'd suggested doing something with `PrisonerSummary` - seeing as that is an interface I suspect it would be in `routerRequestHandler#retrievePrisonerSummaryIfNotInSession`

My thinking was if the data is correct in `PrisonerSummary` in the first place we wouldn't need to do all this trimming etc on all the views

What do you think?

server/routes/routerRequestHandlers.test.ts
```diff
@@ -451,6 +451,34 @@ describe('routerRequestHandlers', () => {
       expect(next).toHaveBeenCalled()
     })
 
+    it('should return prisoners first name and last name without whitespace and a capital letter at the start', async () => {
```
nathanrussell-moj-digital, line 454: Oh yeah; now we're talking! This is perfect. Loving the test for it 👍

chrimesdev: Testing... I'm getting there 😆

nathanrussell-moj-digital (approved):
LGTM 👍

server/routes/routerRequestHandlers.ts
```diff
@@ -123,8 +123,8 @@ const retrievePrisonerSummaryIfNotInSession = (prisonerSearchService: PrisonerSe
           prisonNumber: prisoner.prisonerNumber,
           prisonId: prisoner.prisonId,
           releaseDate: prisoner.releaseDate,
-          firstName: prisoner.firstName,
-          lastName: prisoner.lastName,
+          firstName: `${prisoner.firstName.trim().charAt(0).toUpperCase()}${prisoner.firstName.trim().slice(1)}`,
```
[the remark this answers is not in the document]
nathanrussell-moj-digital: Hang on - thats not quite right (nearly, but not quite! 😁 )
You need to lower case the rest of the string
As it happens all names we get from NOMIS are UPPERCASE
So, with this change, all names displayed will be uppercase on screen!

nathanrussell-moj-digital (changes requested):
Sorry, I know I approved it, but I've just spotted a bug re name casing

server/routes/routerRequestHandlers.ts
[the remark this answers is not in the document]
chrimesdev: When I tested this in a browser I think I still had an old name from a previous change in session! One to watch in future.

server/routes/routerRequestHandlers.ts
```diff
@@ -123,8 +123,14 @@ const retrievePrisonerSummaryIfNotInSession = (prisonerSearchService: PrisonerSe
...
-          firstName: `${prisoner.firstName.trim().charAt(0).toUpperCase()}${prisoner.firstName.trim().slice(1)}`,
-          lastName: `${prisoner.lastName.trim().charAt(0).toUpperCase()}${prisoner.lastName.trim().slice(1)}`,
+          firstName: `${prisoner.firstName.trim().charAt(0).toUpperCase()}${prisoner.firstName
+            .trim()
+            .slice(1)
+            .toLowerCase()}`,
+          lastName: `${prisoner.lastName.trim().charAt(0).toUpperCase()}${prisoner.lastName
+            .trim()
+            .slice(1)
+            .toLowerCase()}`,
```
chrimesdev, line 133: Took a similar approach to https://github.com/mozilla/nunjucks/blob/master/nunjucks/src/filters.js#L50-L54 here but I think this one is a little easier to read?

nathanrussell-moj-digital: What about a small private function?
```
const capitalize = (name: string): string => {
  const trimmedLowercaseName = name.trim().toLowerCase()
  return trimmedLowercaseName.charAt(0).toUpperCase() + trimmedLowercaseName.slice(1)
```

chrimesdev: Makes sense!

chrimesdev: Added

server/routes/routerRequestHandlers.ts
```diff
@@ -112,6 +112,12 @@ const checkNewGoalsFormExistsInSession = async (req: Request, res: Response, nex
  *  Middleware function that returns a Request handler function to look up the prisoner from prisoner-search, map to a PrisonerSummary, and store in the session
  */
 const retrievePrisonerSummaryIfNotInSession = (prisonerSearchService: PrisonerSearchService): RequestHandler => {
+  // Trim whitespace from a name string and capitalize the first letter and lowercase the rest of the string
+  const capitalize = (name: string): string => {
```
nathanrussell-moj-digital, line 116: Love it ❤️ 👍

nathanrussell-moj-digital (approved):
Sorry its been a long drag, but it's important to get some of this stuff right
Thanks for your patience with me, and thanks for making the changes 👍

chrimesdev:
> Sorry its been a long drag, but it's important to get some of this stuff right Thanks for your patience with me, and thanks for making the changes 👍

Thanks for the feedback - all made sense!

PR #187, opened 2023-09-22
RR-152 - Submit each goal to the API

nathanrussell-moj-digital:
This PR changes the `submitReviewGoal` controller function so that it submits each Goal to the `createGoal` service/API
Previously it only submitted the first in the array, and I'd forgotten we need to do this as well! 🙄


mattwills-moj-digital (approved):
LGTM

PR #190, opened 2023-09-25
fix incorrect page link in the breadcrumb partial

chrimesdev:
## Description 

fix incorrect page link in the breadcrumb partial that was introduced in https://github.com/ministryofjustice/hmpps-education-and-work-plan-ui/pull/185/commits/e783b72ea8ab967823e92a21ebf05fe05e4ba229

This was easier than reverting the changes introduced due to various file changes in the commit.

Makes the shared partial breadcrumb use the correct link for the Overview page, but the "View all functional skills" page should link back to Education and training tab as its a sub section of that page.

### DIFF
--- a/server/views/pages/functionalSkills/index.njk
+++ b/server/views/pages/functionalSkills/index.njk
@@ -5,7 +5,23 @@
 {% set pageId = "functional-skills" %}
 
 {% block beforeContent %}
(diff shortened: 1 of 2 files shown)


nathanrussell-moj-digital (approved):
LGTM

PR #199, opened 2023-09-27
RR-367 update create goal mapper to map new field

chrimesdev:
## Description 

Add `targetCompletionDate` to `createGoalMapper` and add `targetCompletionDate` values to the relevant tests and tests builders.

### DIFF
--- a/integration_tests/mockApis/educationAndWorkPlanApi.ts
+++ b/integration_tests/mockApis/educationAndWorkPlanApi.ts
@@ -122,6 +122,6 @@ export interface UpdateGoalRequest {
       sequenceNumber: number
     },
   ]
-  targetCompletionDate?: Date
+  targetCompletionDate?: string
   notes?: string
 }
--- a/integration_tests/mockApis/educationAndWorkPlanApi.ts
+++ b/integration_tests/mockApis/educationAndWorkPlanApi.ts
@@ -122,6 +122,6 @@ export interface UpdateGoalRequest {
       sequenceNumber: number
     },
   ]
-  targetCompletionDate?: string
+  targetCompletionDate?: Date
   notes?: string
 }
--- a/server/@types/dto/index.d.ts
+++ b/server/@types/dto/index.d.ts
@@ -3,7 +3,7 @@ declare module 'dto' {
     prisonNumber: string
     title: string
     steps: Array<AddStepDto>
-    targetCompletionDate?: string
+    targetCompletionDate?: Date
     note?: string
     prisonId: string
   }
@@ -18,7 +18,7 @@ declare module 'dto' {
     title: string
     status: string
     steps: Array<UpdateStepDto>
-    targetCompletionDate?: string
+    targetCompletionDate?: Date
     notes?: string
     prisonId: string
   }
--- a/server/@types/dto/index.d.ts
+++ b/server/@types/dto/index.d.ts
@@ -18,7 +18,7 @@ declare module 'dto' {
     title: string
     status: string
     steps: Array<UpdateStepDto>
-    targetCompletionDate?: Date
+    targetCompletionDate?: string
     notes?: string
     prisonId: string
   }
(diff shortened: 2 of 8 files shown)


server/@types/dto/index.d.ts
```diff
@@ -3,6 +3,7 @@ declare module 'dto' {
     prisonNumber: string
     title: string
     steps: Array<AddStepDto>
+    targetCompletionDate?: string
```
nathanrussell-moj-digital, line 6: Re: my comment in your other PR I'm pretty sure this, and the same field in the form class, should be `Date` rather than string 🤔

server/testsupport/createGoalDtoTestDataBuilder.ts
```diff
@@ -9,6 +9,7 @@ const aValidCreateGoalDtoWithOneStep = (title = 'Learn Spanish'): CreateGoalDto
     prisonNumber: 'A1234BC',
     title,
     steps: [addStepDto],
+    targetCompletionDate: '2024-01-01',
```
nathanrussell-moj-digital, line 12: The values in your test data builders here will need to be dates rather than strings
Have a look where we've used `moment` elsewhere, but I think you want something like
```
moment('2024-01-01').toDate
```
Something like that anyway

integration_tests/mockApis/educationAndWorkPlanApi.ts
```diff
@@ -122,6 +122,6 @@ export interface UpdateGoalRequest {
       sequenceNumber: number
     },
   ]
-  targetCompletionDate?: string
+  targetCompletionDate?: Date
```
nathanrussell-moj-digital, line 125: We need to be careful we stick to the scope of the story. This story is specifically about the Create journey only, so I think I'd like us to revert the changes to "update" related classes , back to being a string.
This field in particular is wrong, as its a mock of the request API body that we send to the Update Goal API. The REST API field will always be a string (because its serialized)

(we will come back and make these changes to the Update related classes in a subsequent story, just not _this_ file, and not in _this_ PR)

server/@types/dto/index.d.ts
```diff
@@ -3,7 +3,7 @@ declare module 'dto' {
     prisonNumber: string
     title: string
     steps: Array<AddStepDto>
-    targetCompletionDate?: string
+    targetCompletionDate?: Date
```
nathanrussell-moj-digital, line 6: This is OK 👍

server/@types/dto/index.d.ts
```diff
@@ -18,7 +18,7 @@ declare module 'dto' {
     title: string
     status: string
     steps: Array<UpdateStepDto>
-    targetCompletionDate?: string
+    targetCompletionDate?: Date
```
nathanrussell-moj-digital, line 21: But this needs to be back as a string if we can (unless you found a compilation or test error that made you change this?)

server/data/mappers/createGoalMapper.test.ts
```diff
@@ -20,6 +20,7 @@ describe('createGoalMapper', () => {
       title: createGoalDto.title,
       category: 'WORK',
       steps: [expectedAddStepRequest1, expectedAddStepRequest2],
+      targetCompletionDate: createGoalDto.targetCompletionDate.toISOString(),
```
nathanrussell-moj-digital, line 23: Rather than put `createGoalDto.targetCompletionDate.toISOString()` in the expected here, we often find its more readable to put the expected value as a literal
```
targetCompletionDate: '2024-01-01'
```
This has a couple of advantages. It makes the test more readable and easier to reason about, and makes it easier to understand what the implementation is actually doing. 👍 

You might also find it highlights a bug with the implementation. You _might_ find (I'm not sure) that the call to `toISOString()` also returns a time component, and we defo don't want that

server/data/mappers/createGoalMapper.ts
```diff
@@ -7,6 +7,7 @@ const toCreateGoalRequest = (createGoalDto: CreateGoalDto): CreateGoalRequest =>
     title: createGoalDto.title,
     category: 'WORK',
     steps: toAddStepRequests(createGoalDto),
+    targetCompletionDate: createGoalDto.targetCompletionDate.toISOString().split('T')[0],
```
chrimesdev, line 10: This strips out the time from the `Date`

nathanrussell-moj-digital (approved):
LGTM 👍

PR #200, opened 2023-09-27
RR-369 code to work out future dates for 3 6 12 month radio button options add to create goal view

chrimesdev:
## Description 

Code to work out future dates for 3 6 12 month radio button options add to create goal view

### DIFF
--- a/server/routes/createGoal/createGoalController.ts
+++ b/server/routes/createGoal/createGoalController.ts
@@ -10,6 +10,7 @@ import { toCreateGoalDto } from './mappers/createGoalFormToCreateGoalDtoMapper'
 import validateCreateGoalForm from './createGoalFormValidator'
 import validateAddStepForm from './addStepFormValidator'
 import ReviewView from './reviewView'
+import futureGoalTargetDateCalculator from '../futureGoalTargetDateCalculator'
 
 export default class CreateGoalController {
   constructor(private readonly educationAndWorkPlanService: EducationAndWorkPlanService) {}
@@ -25,19 +26,11 @@ export default class CreateGoalController {
       } as NewGoal
     }
 
+    const today = moment().toDate()
     const futureGoalTargetDates = [
-      {
-        text: moment(moment().add(3, 'months').toDate()).format('YYYY-MM-DD'),
-        value: `in 3 months (${moment(moment().add(3, 'months').toDate()).format('D MMMM YYYY')})`,
-      },
-      {
-        text: moment(moment().add(6, 'months').toDate()).format('YYYY-MM-DD'),
-        value: `in 6 months (${moment(moment().add(6, 'months').toDate()).format('D MMMM YYYY')})`,
-      },
-      {
-        text: moment(moment().add(12, 'months').toDate()).format('YYYY-MM-DD'),
-        value: `in 12 months (${moment(moment().add(12, 'months').toDate()).format('D MMMM YYYY')})`,
-      },
+      futureGoalTargetDateCalculator(today, 3),
+      futureGoalTargetDateCalculator(today, 6),
+      futureGoalTargetDateCalculator(today, 12),
     ]
 
     const view = new CreateGoalView(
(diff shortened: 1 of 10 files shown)


server/routes/createGoal/createGoalController.ts
```diff
@@ -24,7 +25,18 @@ export default class CreateGoalController {
       } as NewGoal
     }
 
-    const view = new CreateGoalView(prisonerSummary, req.session.newGoal.createGoalForm, req.flash('errors'))
+    const currentDatePlus3Months = moment().add(3, 'months').toDate()
```
chrimesdev, line 28: What format do these needs to be in when passed to the UI? The UI will use the `dateFilter` Nunjucks filter to render 2 dates

1. The human readable date for the Radio button label (27 January 2022)
2. The `value` of the Radio button, used to be passed into the data (presumably 2022-01-27?)

chrimesdev:
@nathanrussell-moj-digital this one should be ready for review now

server/routes/createGoal/createGoalController.ts
```diff
@@ -24,7 +25,27 @@ export default class CreateGoalController {
       } as NewGoal
     }
 
-    const view = new CreateGoalView(prisonerSummary, req.session.newGoal.createGoalForm, req.flash('errors'))
+    const futureGoalTargetDates = [
+      {
```
nathanrussell-moj-digital, line 29: I really like the idea of `futureGoalTargetDates` being an array of `{ text: string; value: string }` and t being a field on the view model object. Nice 👍 

I wonder if we could factor some of this out into a separate module function, and then that can be unit tested separately.
I was thinking something like this here in the controller:
```
const futureGoalTargetDates = [
    someFunctionCall(moment().toDate(), 3),
    someFunctionCall(moment().toDate(), 6),
    someFunctionCall(moment().toDate(), 12),
]
```
I've deliberately called it `someFunctionCall` because I'm not sure what we want to call it yet (I'll let you have a think about that 😁 )

And then the implementation of `someFunctionCall` might be something like:
```
export default someFunctionCall = (startDate: Date, additionalMonths: number): { text: string, value: string } => {
    const futureDate = moment(startDate).add(additionalMonths, 'months').format('D-MMMM-YYYY`)
    return {
        text: futureDate,
        value: `in ${additionalMonths} months (${futureDate})`,
    }
}
```
Something like that .... however ....

It also needs to cater for the stupid rule where if today is the 29/30/31st, and the future month does not have 29/30/31 days, then it needs to be "rolled back" to the last day of the month.
This is where your unit tests will come into play 👍 

Have a think and see what you think

chrimesdev: Thanks @nathanrussell-moj-digital will take a look at that.

As for the final point around days in a month, `moment` is smart enough to handle this itself..

> Special considerations for months and years
> If the day of the month on the original date is greater than the number of days in the final month, the day of the month will change to the last day in the final month.

https://momentjscom.readthedocs.io/en/latest/moment/03-manipulating/01-add/

But we can definitely add tests on our end to make sure this is true

nathanrussell-moj-digital: Oh, fantastic news! `moment` is so smart 😁 Saves us writing what would otherwise be some pretty f'ugly code! 👍 

Re: unit tests, defo include one that would trigger a 29th Feb leap year. 2024 is the next leap year, so perhaps a test where the start date is 30/11/2023, add 3 months, expect output to be 29/02/2024

Something like that anyway 👍

chrimesdev: Added the shared function and tests. The tests do work because a) I spelt February wrong in the tests despite being born in Feburary b) Miscalculated 6 months from February - the tests caught both!

Also the test `should return the last day of month, if the day of the month on the original date is greater than the number of days in the final month` covers both the scenarios we discussed above; the leap year but also the if the future month does not have the same amount of days it rolls back. Hopefully 1 test for both is good enough, rather than duplicating it.

nathanrussell-moj-digital (approved):
This looks really good 👍 
Nice work 👍

PR #202, opened 2023-09-28
RR-368 disable the  rule in Axe to allow us to use conditional radios

chrimesdev:
## Description

Now we plan to use conditional radio buttons from `govuk-frontend`, we were getting an accessibility error which were preventing the tests from passing.

The only time we use `aria` attributes is when we're using `govuk-frontend` components, so we're safe to disable this rule, as these components have been tested by the GDS team.


chrimesdev:
This can go independently of others @nathanrussell-moj-digital, its pre-empting a Cypress-Axe error you get when using conditional Radios in the view.

nathanrussell-moj-digital (approved):

PR #203, opened 2023-09-28
RR-368 add UI components to the create goal target date fields

chrimesdev:
## Description

Add UI components to the create goal target date fields


server/views/pages/goal/create/index.njk
```diff
@@ -23,37 +23,84 @@
...
+              }
             },
             hint: {
-              html: hintTextHtml
-            },
-            errorMessage: errors | findError('title')
+              text: "For example, 27 3 2007"
+            }
           }) }}
+        {% endset -%}
```
nathanrussell-moj-digital, line 77: Whats the `-%}` syntax here? Is the `-` a typo? or is it some magic secret sauce I've not see before? 😁

server/views/pages/goal/create/index.njk
```diff
@@ -23,37 +23,84 @@
...
-            text: "Continue to add steps"
-          }) }}
+        {% set targetCompletionDateItems = (targetCompletionDateItems.push({
+          "value": "another-date",
+          "text": "set another date",
+          "conditional": {
+            "html": anotherDateHtml
+          }
+          }), targetCompletionDateItems)
+        %}
```
nathanrussell-moj-digital, line 86: Ohh, this is messier than I'd anticipated, but I can see what it's doing and why ...

I wonder if we can do away with the `targetCompletionDateItems` array altogether, and simply push to the `futureGoalTargetDates` array?

Would something like this work?
```
{{ futureGoalTargetDates.push({
          "value": "another-date",
          "text": "set another date",
          "conditional": {
            "html": anotherDateHtml
          }
          })
}}
```

server/views/pages/goal/create/index.njk
chrimesdev: It just removes whitespace in HTML. Because we're just setting something in the template file, we don't want the whitespace to appear in the rendered HTML DOM. Nothing fancy!

server/@types/forms/index.d.ts
```diff
@@ -3,6 +3,9 @@ declare module 'forms' {
     prisonNumber: string
     title?: string
     targetCompletionDate?: Date
+    'another-date-day'?: string
```
nathanrussell-moj-digital, line 6: Notwithstanding what I've commented on below re: the nunjucks view and manipulating arrays etc - could we change these field names please to `targetCompletionDate-xxx` ?
I think if you are only looking at this form class and see a field `targetCompletionDate` and `another-date-xxx` they are not obviously related. Someone might be forgiven for thinking they are completely different fields / have different purposes, which is not the case here

chrimesdev: Yep that makes sense - updated to be `targetCompletionDate-xxx`

nathanrussell-moj-digital (approved):
LGTM 👍 
(though I still find the array manipulation stuff in the nunjucks view horrible! 🤣 )

Lets get this merged 👍

PR #207, opened 2023-09-29
use the correct heading level for goals overview list

chrimesdev:
## Description

The list of Goals heading should be a level 3 heading as it sits under "Goals in progress" which is a level 2 heading.


nathanrussell-moj-digital (approved):
LGTM 👍

PR #212, opened 2023-10-02
RR-372 add Cypress test for user entering a custom target completion date

chrimesdev:
## Description

Add Cypress test for user entering a custom target completion date


nathanrussell-moj-digital (approved):
LGTM 👍

PR #215, opened 2023-10-03
RR-338 - Show error bars on `targetCompletionDate` on validation error

nathanrussell-moj-digital:
This PR shows the error bars on the Update Goal `targetCompletionDate` when the field is in error (forgot to do it in my last PR) and adds some cypress tests to prove

Screenshot 2023-10-03 at 09 18 03

### DIFF
--- a/integration_tests/e2e/goal/addNote.cy.ts
+++ b/integration_tests/e2e/goal/addNote.cy.ts
@@ -44,7 +44,7 @@ context('Add a note', () => {
     const createGoalPage = overviewPage.clickAddGoalButton()
     createGoalPage //
       .setGoalTitle('Learn French')
-      .setGoalTargetDate()
+      .setTargetCompletionDate0to3Months()
       .submitPage()
 
     const addStepPage = Page.verifyOnPage(AddStepPage)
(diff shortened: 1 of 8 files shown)


integration_tests/e2e/goal/addNote.cy.ts
```diff
@@ -44,7 +44,7 @@ context('Add a note', () => {
     const createGoalPage = overviewPage.clickAddGoalButton()
     createGoalPage //
       .setGoalTitle('Learn French')
-      .setGoalTargetDate()
+      .setTargetCompletionDate0to3Months()
```
nathanrussell-moj-digital, line 47: Not strictly related to Update Goal, but I renamed this page method on `CreateGoalPage` to better reflect what it does (the thinking being we might want to add other methods one day that select the 3-6 month option, or the 6-12 month option etc

integration_tests/e2e/goal/updateGoal.cy.ts
```diff
@@ -48,6 +48,7 @@ context('Update a goal', () => {
     // When
     updateGoalPage //
       .clearGoalTitle()
+      .setTargetCompletionDateFromValuePreviouslySaveInGoal()
```
nathanrussell-moj-digital, line 51: Quite a long method name, but it does what it says on the tin?
Basically its a method that sets the target completion date radio button on the Update Goal page from the value that was previously set for the Goal

chrimesdev (approved):

PR #217, opened 2023-10-03
RR-338 - Added target completion date to the Update Goal Review screen

nathanrussell-moj-digital:
This PR adds the `targetCompletionDate` field to the Update Goal "review your answers" page.
We don't have any clear direction in the ticket as to where the field should be positioned / what the text should be (there are no links to the prototype or other screenshots), so I've followed my nose a little and based it on the review your answers page for Create Goal. If its wrong we can always change it 🤷 

Screenshot 2023-10-03 at 10 50 36

### DIFF
--- a/server/routes/updateGoal/reviewUpdateGoalView.ts
+++ b/server/routes/updateGoal/reviewUpdateGoalView.ts
@@ -1,16 +1,16 @@
 import type { PrisonerSummary } from 'viewModels'
-import type { UpdateGoalForm } from 'forms'
+import type { UpdateGoalDto } from 'dto'
 
 export default class ReviewUpdateGoalView {
-  constructor(private readonly prisonerSummary: PrisonerSummary, private readonly updateGoalForm: UpdateGoalForm) {}
+  constructor(private readonly prisonerSummary: PrisonerSummary, private readonly updateGoalDto: UpdateGoalDto) {}
 
   get renderArgs(): {
     prisonerSummary: PrisonerSummary
-    data: UpdateGoalForm
+    data: UpdateGoalDto
   } {
     return {
       prisonerSummary: this.prisonerSummary,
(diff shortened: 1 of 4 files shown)


server/routes/updateGoal/reviewUpdateGoalView.ts
```diff
@@ -1,16 +1,16 @@
 import type { PrisonerSummary } from 'viewModels'
-import type { UpdateGoalForm } from 'forms'
+import type { UpdateGoalDto } from 'dto'
 
 export default class ReviewUpdateGoalView {
-  constructor(private readonly prisonerSummary: PrisonerSummary, private readonly updateGoalForm: UpdateGoalForm) {}
+  constructor(private readonly prisonerSummary: PrisonerSummary, private readonly updateGoalDto: UpdateGoalDto) {}
```
nathanrussell-moj-digital, line 5: I've changed this to pass the DTO into the view object, rather than the form
The reason for this is that the DTO contains a single field for the `targetCompletionDate` of type `Date`, whereas the form contains 4 optional fields (the date, or the dd, mm and yyyy fields - all of which are string)
The form to DTO mapper takes care of the logic re: which fields from the form to use to construct the `targetCompletionDate` `Date` instance; so I felt this was easier / safer to use, rather than having some scary logic in the nunjucks view to work out the correct date to render.

server/views/pages/goal/update/review.njk
```diff
@@ -42,7 +50,7 @@
               text: "Note"
             },
             value: {
-              text: data.note
+              text: data.notes
```
nathanrussell-moj-digital, line 53: A couple of the fields in the DTO have slightly different names to the form object - hence this change to `notes` and the one below to `sequenceNumber`

chrimesdev (approved):

PR #219, opened 2023-10-03
RR-386 use the correct page title format

chrimesdev:
## Description

Use the correct page title format

https://dsdmoj.atlassian.net/browse/RR-386

### DIFF
--- a/server/views/pages/error.njk
+++ b/server/views/pages/error.njk
@@ -1,7 +1,7 @@
 {% extends "../partials/layout.njk" %}
 
 {% set pageId = 'service-unavailable' %}
-{% set pageTitle = applicationName + " - Service unavailable" %}
+{% set pageTitle =  "Sorry, there is a problem with this service" %}
 
 {% block content %}
 
--- a/server/views/pages/404.njk
+++ b/server/views/pages/404.njk
@@ -1,7 +1,7 @@
 {% extends "../partials/layout.njk" %}
 
 {% set pageId = 'page-not-found' %}
-{% set pageTitle = applicationName + " - Page not found" %}
+{% set pageTitle = "Page not found" %}
 
 {% block content %}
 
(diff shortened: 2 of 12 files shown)


nathanrussell-moj-digital (commented):
Can we talk about this @chrimesdev ?
What you've done looks fine, but I don't understand the Jira story (RR-386), specifically what we are being asked to do in relation to `h1` elements which I dont think you are changing here?

server/views/pages/error.njk
```diff
@@ -1,7 +1,7 @@
 {% extends "../partials/layout.njk" %}
 
 {% set pageId = 'service-unavailable' %}
-{% set pageTitle = applicationName + " - Service unavailable" %}
+{% set pageTitle =  "Sorry, there is a problem with this service" %}
```
nathanrussell-moj-digital, line 4: Not that I have a problem with them, but where did we get all these page `<title>` element values from? Are they in some content document somewhere? because they don't appear to be listed or detailed in the Jira ticket at all?

chrimesdev:
> Can we talk about this @chrimesdev ? What you've done looks fine, but I don't understand the Jira story (RR-386), specifically what we are being asked to do in relation to `h1` elements which I dont think you are changing here?

The ask is that the page `<title>` should follow the format of `Page name - Digital Prison Services` with the Page name being the page `<h1>`. 

However because Google analytics tracks page titles, anywhere that the prisoner name appears this should be replaced with "prisoner/prisoner's".

Current pages `<h1>` are all fine and don't need changing.

server/views/pages/error.njk
chrimesdev: There were a number of key pages listed in the JIRA ticket, however pages like this must have been forgotten. So this example is just ensuring the page name `<h1>` is correct in the `<title>`

chrimesdev:
Looking at the JIRA ticket, just seems theres been some confusion with `<h1>` and `<title>` elements. I'd switched the wording in the JIRA ticket the use the correct element naming.

nathanrussell-moj-digital (approved):
Sorry, I forgot to approve this one yesterday after we spoke about it!

All looks good; nice work and thanks for the explanation and update to the Jira ticket to clarify 👍 Thank

PR #222, opened 2023-10-04
RR-389 add hidden text to the goal overview update links

chrimesdev:
## Description 

Add hidden text to the goal overview update links, so that a screen reader has context as to what goal number they are updating.

https://dsdmoj.atlassian.net/browse/RR-389


nathanrussell-moj-digital (approved):
LGTM 👍

PR #224, opened 2023-10-04
RR-338 - Fix targetCompletionDate validation on Update Goal screen

nathanrussell-moj-digital:
This PR fixes the bug that James found whilst testing the Update Goal screen in respect of the target completion date validation.

There were 2 problems:

* The date validator itself needed a little beefing up, in that it allowed some date values which where clearly invalid!!! (ie. apparently `][-08-2024` is a valid date! 🤣 
* We were not holding the original value of the target completion date (ie. from the goal's original state before any updates) - so when we submitted an invalid date, the value on screen became `Invalid date` 😢

### DIFF
--- a/server/@types/forms/index.d.ts
+++ b/server/@types/forms/index.d.ts
@@ -34,6 +34,7 @@ declare module 'forms' {
     note?: string
     steps: Array<UpdateStepForm>
     action?: 'add-another-step' | 'submit-form' | 'delete-step-[0]'
+    originalTargetCompletionDate: string
   }
 
   export interface UpdateStepForm {
--- a/server/routes/updateGoal/updateGoalController.ts
+++ b/server/routes/updateGoal/updateGoalController.ts
@@ -36,7 +36,7 @@ export default class UpdateGoalController {
     req.session.updateGoalForm = undefined
 
     const goalCreatedDate = moment(updateGoalForm.createdAt)
-    const goalTargetCompletionDate = moment(updateGoalForm.targetCompletionDate)
+    const goalTargetCompletionDate = moment(updateGoalForm.originalTargetCompletionDate)
     const goalTargetCompletionDateOption = {
       value: goalTargetCompletionDate.format('YYYY-MM-DD'),
       text: `by ${goalTargetCompletionDate.format('D MMMM YYYY')} (goal created on ${goalCreatedDate.format(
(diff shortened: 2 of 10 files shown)


server/@types/forms/index.d.ts
```diff
@@ -34,6 +34,7 @@ declare module 'forms' {
     note?: string
     steps: Array<UpdateStepForm>
     action?: 'add-another-step' | 'submit-form' | 'delete-step-[0]'
+    originalTargetCompletionDate: string
```
nathanrussell-moj-digital, line 37: We now hold the original target completion date (from the original un-edited Goal) in the form, so that we know what it was (in the case of the user entering an invalid date)

server/routes/updateGoal/updateGoalController.ts
```diff
@@ -36,7 +36,7 @@ export default class UpdateGoalController {
     req.session.updateGoalForm = undefined
 
     const goalCreatedDate = moment(updateGoalForm.createdAt)
-    const goalTargetCompletionDate = moment(updateGoalForm.targetCompletionDate)
+    const goalTargetCompletionDate = moment(updateGoalForm.originalTargetCompletionDate)
```
nathanrussell-moj-digital, line 39: Here we now use the `originalTargetCompletionDate` from the form (which never gets mutated) to create the radio button option, rather than `targetCompletionDate` (which will be `Invalid date` if the user has tried to submit an invalid date)

server/validators/goalTargetCompletionDateValidator.test.ts
```diff
@@ -11,52 +11,27 @@ describe('goalTargetDateValidator', () => {
...
-    // Given
-    const day = '26'
-    const month = '40'
-    const year = '2024'
-    // When
-    const errors = validateTargetDate('another-date', day, month, year)
-
-    // Then
-    expect(errors).toStrictEqual(['Enter a valid date'])
+  Array.of(
```
nathanrussell-moj-digital, line 14: I've removed a load of individual tests from this test class, and replaced them with an array
IE. they are all testing the same outcome; and are just a series of examples that should drive the same behaviour/outcome
All the tests I've removed are in this array, plus a few extras that I found based on the testing I did on the deployed app in order to reproduce the bug

server/validators/goalTargetCompletionDateValidator.ts
```diff
@@ -29,3 +35,7 @@ export default function goalTargetCompletionDateValidator(
 
   return errors
 }
+
+const isNotNumeric = (value: string): boolean => {
```
nathanrussell-moj-digital, line 39: This is how we catch some of the weird cases (such as `][` apparently being a valid day component !!!!)

chrimesdev (approved):

PR #227, opened 2023-10-05
RR-353 update the curious down messaging on support needs tabs

chrimesdev:
## Description

Update the curious down messaging on support needs tabs

https://dsdmoj.atlassian.net/browse/RR-353

Screenshot 2023-10-05 at 11 06 28


chrimesdev:
May need to update a few tickets content at once as the Cypress test uses a shared test for the Curious error message title..

integration_tests/pages/overview/OverviewPage.ts
```diff
@@ -134,7 +134,7 @@ export default class OverviewPage extends Page {
   }
 
   hasCuriousUnavailableMessageDisplayed() {
-    cy.get('h2').contains('Sorry, the Curious service is currently unavailable')
```
nathanrussell-moj-digital, line 137: Good point about this breaking other cypress tests ... 
I think this is a case of a fragile test (ie. the test is too tightly coupled to a specific implementation - in this case words on screen)

I wonder if there's a different way of doing it (though it would mean a little refactoring elsewhere) ...

What about if this test didn't check the words in a h2 (which interestingly could be any h2 on the page as it currently stands), and instead checked for the presence of an element with a specific data-qa attribute

And then you go through all the pages that show an error like this (ie. it will be all the test that break! 😁 ), and add the data-qa attribute to a suitable element

Does that make sense? Is that do-able ?

chrimesdev: Sounds sensible! Will take a look

chrimesdev: @nathanrussell-moj-digital how do this look now?

nathanrussell-moj-digital (approved):
Looks great; and I like the updated cypress assertion so that it checks for an element rather then a very specific phrase of text. Nice work 👍

PR #235, opened 2023-10-10
RR-338 - Fix date validation so that years have to be entered as 4 digit years

nathanrussell-moj-digital:
This PR fixes a problem with our date validation where it previously allowed a 2 digit year, but then got very confused when turning it into a date object, which in turn caused at error at the API (which wants a 4 digit year)

Date validation now checks for a 4 digit year:

Screenshot 2023-10-10 at 08 56 07


mattwills-moj-digital (approved):
LGTM

PR #241, opened 2023-10-12
RR-402 use the correct colour on the recorded on date text

chrimesdev:
## Description 

Use the correct colour on the recorded on date text in the Neurodiversity section

https://dsdmoj.atlassian.net/browse/RR-402


nathanrussell-moj-digital (approved):
LGTM

PR #248, opened 2023-10-13
RR-364 update the styling of the qualifications sidebar

chrimesdev:
## Description

Update the styling of the qualifications sidebar

(Don't ask me why its green)

https://dsdmoj.atlassian.net/browse/RR-364

Screenshot 2023-10-13 at 15 45 09

### DIFF
--- a/assets/scss/components/_summary-card.scss
+++ b/assets/scss/components/_summary-card.scss
@@ -1,6 +1,14 @@
-.app-summary-card__title-wrapper--white {
+.app-summary-card__title-wrapper {
   background-color: #fff;
   border-bottom: 1px solid transparent;
-  border-top: 4px solid #000;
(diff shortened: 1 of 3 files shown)


nathanrussell-moj-digital (approved):
I'm really anti this (as you probably know 🤣, and I think you share my concerns too )
Not in respect of what you've done here - the "how" is perfect 👍 
I'm anti the "what" and the "why" ; but I think we've both said our pieces and its fallen on deaf ears 😢 

I think all we can do is go with it and merge it; and hope that the accessibility testers (which starts today) raise questions "why is this green but nothing else is? What does the colour "mean"? What is it trying to tell the user through the use of colour? How would a blind or colour blind person get the same meaning from it?"

🤷‍♂️

PR #249, opened 2023-10-13
RR-400 - introduced PagedPrisonerSearchSummary 

nathanrussell-moj-digital:
This PR introduces the class `PagedPrisonerSearchSummary` which has methods to sort, filter and pagination a list of `PrisonerSearchSummary`

The idea is that `PagedPrisonerSearchSummary` is essentially a "view model" class and it will be used by the (yet to be developed) Prisoner List controller and nunjucks view. 
Normally we do "view models" as types (interfaces) in the `@types` folder, but this needs to be more than just an interface because it needs to have some behaviours. Specifically it needs to do sorting, filtering and pagination.

My thinking is that the controller (yet to be written) will call a service (yet to be written) that returns a collection of `PrisonerSearchSummary` instances. (`PrisonerSearchSummary` already exists).
The controller will then new up a `PagedPrisonerSearchSummary` passing in the array of `PrisonerSearchSummary` instances returned from the service.
Once the controller has the `PagedPrisonerSearchSummary` instance it can call the various filter, sort and paginate methods, based on "requests" (form submissions) from the nunjucks view.
Thats the basic idea anyway

### DIFF
--- a/server/routes/prisonerList/pagedPrisonerSearchSummary.test.ts
+++ b/server/routes/prisonerList/pagedPrisonerSearchSummary.test.ts
@@ -66,7 +66,7 @@ describe('pagedPrisonerSearchSummary', () => {
       expect(pagedPrisonerSearchSummaries.pageSize).toEqual(5)
       expect(pagedPrisonerSearchSummaries.totalResults).toEqual(3)
       expect(pagedPrisonerSearchSummaries.totalPages).toEqual(1)
-      expect(pagedPrisonerSearchSummaries.currentPage).toEqual(1)
+      expect(pagedPrisonerSearchSummaries.currentPageNumber).toEqual(1)
       expect(pagedPrisonerSearchSummaries.resultIndexFrom).toEqual(1)
       expect(pagedPrisonerSearchSummaries.resultIndexTo).toEqual(3)
     })
@@ -83,7 +83,7 @@ describe('pagedPrisonerSearchSummary', () => {
       expect(pagedPrisonerSearchSummaries.pageSize).toEqual(2)
       expect(pagedPrisonerSearchSummaries.totalResults).toEqual(3)
       expect(pagedPrisonerSearchSummaries.totalPages).toEqual(2)
-      expect(pagedPrisonerSearchSummaries.currentPage).toEqual(1)
+      expect(pagedPrisonerSearchSummaries.currentPageNumber).toEqual(1)
       expect(pagedPrisonerSearchSummaries.resultIndexFrom).toEqual(1)
       expect(pagedPrisonerSearchSummaries.resultIndexTo).toEqual(2)
     })
@@ -100,13 +100,13 @@ describe('pagedPrisonerSearchSummary', () => {
       expect(pagedPrisonerSearchSummaries.pageSize).toEqual(3)
       expect(pagedPrisonerSearchSummaries.totalResults).toEqual(3)
       expect(pagedPrisonerSearchSummaries.totalPages).toEqual(1)
-      expect(pagedPrisonerSearchSummaries.currentPage).toEqual(1)
+      expect(pagedPrisonerSearchSummaries.currentPageNumber).toEqual(1)
       expect(pagedPrisonerSearchSummaries.resultIndexFrom).toEqual(1)
       expect(pagedPrisonerSearchSummaries.resultIndexTo).toEqual(3)
     })
   })
 
-  describe('setCurrentPage', () => {
+  describe('setCurrentPageNumber', () => {
     it('should set current page given page number less than 1', () => {
       // Given
       const pageSize = 2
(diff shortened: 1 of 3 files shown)


server/routes/prisonerList/pagedPrisonerSearchSummary.test.ts
```diff
@@ -0,0 +1,570 @@
+import aValidPrisonerSearchSummary from '../../testsupport/prisonerSearchSummaryTestDataBuilder'
```
nathanrussell-moj-digital, line 1: Strap yourself in, this is a big test class (570 lines!)
However, I think its a good test class with all the possible scenarios and permutations covered 👍 

On a serious note, it might be worth reviewing the implementation class first (next file down 👇 ) as hopefully that will give you a clear understanding of what the class is and what it does, which in turn will make reviewing the test easier 🤞

server/testsupport/prisonerSearchSummaryTestDataBuilder.ts
```diff
@@ -0,0 +1,32 @@
+import moment from 'moment'
+import type { PrisonerSearchSummary } from 'viewModels'
+
```
nathanrussell-moj-digital, line 3: This is quite a neat pattern for our test data builders, and makes them a bit more like what we do in kotlin 👍 
What we do is pass an optional map of "options" into the test data builder function, where each "option" is a named field. All of the options are optional. If a given option value is not provided then it gets a default value (though to be fair the default values themselves are not as clear and obvious as they are in our beloved kotlin 🤷‍♂️ )

Anyway you can do this:
```
const defaultPrisoner = aValidPrisonerSearchSummary() // A1234BC, BXI, Jimmy Lightfingers etc)
const jimmyInMoorland = aValidPrisonerSearchSummary({ prisonId: 'MDI' }) // A1234BC, MDI, Jimmy Lightfingers etc)
const aDifferentJimmyInADifferentPrisoner = aValidPrisonerSearchSummary({ prisonNumber: 'T4385GH', prisonId: 'MDI' }) // T4385GH, MDI, Jimmy Lightfingers etc)
const anEntirelyDifferentPrisoner = aValidPrisonerSearchSummary({ prisonNumber: 'Y8310JK', prisonId: 'RDG', lastName: 'Mahone', firstName: 'James' }) // Y8310JK, RDG, James Mahone etc)
```

You only need to specify the options that you want to change from the defaults, and because they are named arguments you can specify them in any order .... just like a kotlin function 👍 😁

server/routes/prisonerList/pagedPrisonerSearchSummary.ts
```diff
@@ -0,0 +1,219 @@
...
+export default class PagedPrisonerSearchSummary {
+  private prisonerSearchSummaries: PrisonerSearchSummary[]
+
+  currentPageNumber: number
+
+  totalPages: number
+
+  pageSize: number
+
+  totalResults: number
```
nathanrussell-moj-digital, line 24: I'm not entirely happy with the level/access control of the encapsulation of these fields; but I guess this is more to do with the language, and that I cant help but compare it with what we'd do in java / kotlin.

By that I mean this (and the other properties here) are simple properties. They can be read publicly which is perfect:
```
if (pagedPrisonerSearchSummary.totalResults === 0) { 
  ....
```
but .... they can also be set publicly 😢 
```
pagedPrisonerSearchSummary.totalResults = 100
```

This annoys (the purist in) me!, and I did spend some time trying to get round it. The main problem is that javascript has no support for this, and typescript has limited support in the area of it's `readonly` modifier. Our fields here are not readonly - they are set internally at various stages as you apply filters, sorts, etc
The only way I could get it to work was to make the field private, then expose a getter method bot no setter method. It worked but it was kinda ugly, both in respect of the length of this class itself, plus the calling/consuming code (because it now has to call a getter method rather then simply reading the property)

In the end I decided to leave it as it is, accept that its not ideal (IMHO), and move on with life

mattwills-moj-digital: I think you may have been in danger of worrying unnecessarily! 😉

nathanrussell-moj-digital: You are 💯 right! 👍 😁 
I think the problem was that because it ended up being a class with behaviour, it kinda felt like I was doing it OO style in java or kotlin 😁 So you naturally start thinking like you would in java etc 😁 
But you are right, I'm worrying over nothing 😁

server/routes/prisonerList/pagedPrisonerSearchSummary.ts
```diff
@@ -0,0 +1,219 @@
...
+   * Attempting to set the current page number to a number greater than the total number of pages will set the current
+   * page number to the last page number.
+   *
+   * This method also updates the `resultIndexFrom` and `resultIndexTo` fields to allow view concerns to be able to render
+   * content such as "Showing records x thru y"
+   */
+  setCurrentPageNumber = (pageNumber: number): PagedPrisonerSearchSummary => {
+    this.currentPageNumber = Math.min(this.totalPages, Math.max(1, pageNumber))
+    this.totalResults = this.prisonerSearchSummaries.length
+    this.resultIndexFrom = Math.min(1 + (this.currentPageNumber - 1) * this.pageSize, this.totalResults)
```
mattwills-moj-digital, line 53: Minor - I know that the addition (+ 1) happens after the multiplication because it takes precedence, but what about surrounding it with an extra set of parenthesis? I think it helps improve readability (but like I said, this is a minor comment!):
```
1 + ((this.currentPageNumber - 1) * this.pageSize)
```

mattwills-moj-digital (approved):
LGTM. I can see you've put a lot of time into this. Nice work 🙌

chrimesdev:
LGTM

server/routes/prisonerList/pagedPrisonerSearchSummary.ts
nathanrussell-moj-digital: Interesting you should suggest this - I did exactly this when I first wrote it, and the linter told me off and said the parenthesis were unnecessary and made me remove them! 🤣 🤷‍♂️

PR #254, opened 2023-10-18
RR-395 - Fleshed out `PrisonerListService` and associated classes

nathanrussell-moj-digital:
This PR fleshes out the `PrisonerListService` class by introducing the method `getPrisonerSearchSummariesForPrisonId`

It also does a little tidying in a closely related class

### DIFF
--- a/server/data/mappers/prisonerSearchSummaryMapper.ts
+++ b/server/data/mappers/prisonerSearchSummaryMapper.ts
@@ -1,0 +1,15 @@
+import type { PrisonerSearchSummary } from 'viewModels'
+import type { Prisoner } from 'prisonRegisterApiClient'
+import toPrisonerSummary from './prisonerSummaryMapper'
+
+export default function toPrisonerSearchSummary(
+  prisoner: Prisoner,
+  hasCiagInduction: boolean,
+  hasActionPlan: boolean,
+): PrisonerSearchSummary {
+  return {
+    ...toPrisonerSummary(prisoner),
+    hasCiagInduction,
+    hasActionPlan,
+  }
+}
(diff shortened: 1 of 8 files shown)


server/data/mappers/prisonerSearchSummaryMapper.ts
```diff
@@ -0,0 +1,15 @@
+import type { PrisonerSearchSummary } from 'viewModels'
```
nathanrussell-moj-digital, line 1: I introduced the `PrisonerSearchSummary` view model type the other day. This is the first PR that actually starts using it, hence this mapper.
A `PrisonerSearchSummary` is basically just a `PrisonerSummary` (built from a `Prisoner`) with 2 additional fields - `hasCiagInduction` and `hasActionPlan`, where these 2 new fields will be used on the Prisoner List screen to determine the status (Needs Plan)

server/services/prisonerSearchService.ts
```diff
@@ -1,14 +1,16 @@
-import type { Prisoner } from 'prisonRegisterApiClient'
+import type { PrisonerSummary } from 'viewModels'
 import { HmppsAuthClient, PrisonerSearchClient } from '../data'
+import toPrisonerSummary from '../data/mappers/prisonerSummaryMapper'
 
 export default class PrisonerSearchService {
   constructor(
     private readonly hmppsAuthClient: HmppsAuthClient,
     private readonly prisonerSearchClient: PrisonerSearchClient,
   ) {}
```
nathanrussell-moj-digital, line 10: This change is not directly related to `PrisonerListService`, but is to do with `PrisonerSearchService`
All of our other services deal with a REST API via it's client class which returns the type from the API definition, and then maps it into a view model or similar. The mapping is in the service method, and the service method does not expose the type from the API definition.

`PrisonerSearchService` did not do that - it exposed and returned the type from the REST API definition, and expected it's consumers to map it into something like a view model

This change (and the associated changes to its test class and its consuming class `routerRequestHandler`) correct that and change this service so that it adopts the patterns we use for all our other services

server/routes/routerRequestHandlers.test.ts
```diff
@@ -397,10 +396,8 @@ describe('routerRequestHandlers', () => {
       const prisonNumber = 'A1234GC'
       const prisonId = 'MDI'
       req.params.prisonNumber = prisonNumber
```
nathanrussell-moj-digital, line 398: See my comments below on `PrisonerSearchService` to understand why this test and its implementation have changed

chrimesdev (approved):
Really good work

PR #256, opened 2023-10-18
RR-406 - Remove stub prisoner list page

nathanrussell-moj-digital:
This PR removes the stub prisoner list page.

This is now OK to do as we have our own prisoner list page, controlled by its own feature toggle.

The only current use case of our stub prisoner list page is for our cypress integration tests, because without it there is nothing on route `/` which means cypress cannot login and drive the browser. The cypress tests made no interaction on the stub page itself (ie. there were no tests that clicked any given prisoners etc)

The feature toggle is removed and all uses of it; and the cypress tests are updated a little (because going to our new Prisoner List page requires that the user is logged in, whereas the stub prisoner list page did not)


chrimesdev:
LGTM

chrimesdev (approved):

PR #262, opened 2023-10-20
RR-445 - Correctly render prisoner name on Create Goal screen

nathanrussell-moj-digital:
Bugfix to treat the prisoner's name as safe markup (trusted), which means special characters are correctly rendered.

This is how the page now renders, with the apostrophe in `'Incentine`'s name being correctly rendered (see jira ticket for screenshot of how it renders without this fix)
Screenshot 2023-10-20 at 09 06 32

### DIFF
--- a/server/views/pages/goal/create/index.njk
+++ b/server/views/pages/goal/create/index.njk
@@ -27,7 +27,7 @@
         <input type="hidden" name="prisonNumber" value="{{ form.prisonNumber }}" />
 
         {% set titleLabelHtml %}
-          Describe the goal you want to create for {{ prisonerSummary.firstName }} {{ prisonerSummary.lastName }}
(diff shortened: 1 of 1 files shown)


server/views/pages/goal/create/index.njk
```diff
@@ -27,7 +27,7 @@
         <input type="hidden" name="prisonNumber" value="{{ form.prisonNumber }}" />
```
nathanrussell-moj-digital, line 28: This is the only place that needed this fix. I've been through all the nunjucks templates looking for where we reference `{{ prisonerSummary.firstName }}` and `{{ prisonerSummary.lastName }}`
This is the only place where we use it as part of setting a nunjucks variable, and is the only screen that has this problem.
All other screens where we reference the prisoner name do so within HTML tags, so I guess the rendering/escaping treatment is different based on whether you use it directly in an element/markup, or in a nunjucks variable.

mattwills-moj-digital (approved):
Safe innit

PR #263, opened 2023-10-20
RR-439 - Map `inductionQuestionSet` from `hopingToGetWork` from CIAG data

nathanrussell-moj-digital:
This PR introduces a new field `inductionQuestionSet` to the view models `WorkAndInterests` and `OtherQualifications`, and maps it from the `hopingToGetWork` answer in the CIAG data.

This is an enabler PR where the new field `inductionQuestionSet` is required for us to render different content on the "Work and interests" tab and "Education and training" tab based on whether the prisoner did the "short" or "long" route/flow in their CIAG Induction.

I've tried to steer clear from using terminology like "route" or "flow", and indeed "CIAG" Induction. Instead I've seen it as the type of question set that was presented at the Induction. It was either the short or long question set. The problem with "flow" is that it implies a specific route or flow through the system, and we've already seen that its not as clear cut as 2 different flows - the short "flow" actually has other flows within it based on how you answer other questions. Perhaps "question set" also suffers with the same problem? but I think its better than "flow"

Anyway, the idea is that the 2 tabs in question use either the  `WorkAndInterests` or `OtherQualifications` view model object, which is what we populate the nunjucks view from. By adding a new field `inductionQuestionSet` we can introduce the necessary logic in the view to render different things (this will be in subsequent PRs)

### DIFF
--- a/server/@types/viewModels/index.d.ts
+++ b/server/@types/viewModels/index.d.ts
@@ -116,6 +116,7 @@ declare module 'viewModels' {
    */
   export interface WorkAndInterests {
     problemRetrievingData: boolean
+    inductionQuestionSet?: 'SHORT_QUESTION_SET' | 'LONG_QUESTION_SET'
     data?: WorkAndInterestsData
   }
 
@@ -222,6 +223,7 @@ declare module 'viewModels' {
    */
   export interface OtherQualifications {
     problemRetrievingData: boolean
+    inductionQuestionSet?: 'SHORT_QUESTION_SET' | 'LONG_QUESTION_SET'
     highestEducationLevel?:
       | 'PRIMARY_SCHOOL'
       | 'SECONDARY_SCHOOL_LEFT_BEFORE_TAKING_EXAMS'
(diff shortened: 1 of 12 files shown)


server/@types/viewModels/index.d.ts
```diff
@@ -116,6 +116,7 @@ declare module 'viewModels' {
    */
   export interface WorkAndInterests {
     problemRetrievingData: boolean
+    inductionQuestionSet?: 'SHORT_QUESTION_SET' | 'LONG_QUESTION_SET'
```
nathanrussell-moj-digital, line 119: I wish we could model and export enums in these type interfaces, but it seems we cant 😢

server/services/ciagInductionService.test.ts
```diff
@@ -1,20 +1,13 @@
...
 import CiagInductionClient from '../data/ciagInductionClient'
 import CiagInductionService from './ciagInductionService'
 
 describe('ciagInductionService', () => {
-  const hmppsAuthClient = {
-    getSystemClientToken: jest.fn(),
-  }
   const ciagClient = {
     getCiagInduction: jest.fn(),
   }
```
nathanrussell-moj-digital, line 9: I have removed `hmppsAuthClient` from the service class implementation, as it wasnt being used

PR #265, opened 2023-10-23
RR-439 - Refactor OverviewPage class into a separate page driver class per Overview tab

nathanrussell-moj-digital:
This PR refactors the `OverviewPage` (god) class into a separate page driver class per Overview tab.

This is necessary because the current `OverviewPage` class contains functionality to drive and assert all 4 tabs on the Overview page. It has become a bit of a god class, and this will only get worse once 2 of the tabs (Education and training, and Work interests) have difference content based on short or long Induction question set, and also as and when the Timeline comes along.

It turned out to be a PITA to do this refactoring - I wanted to use OO techniques , with the concept of an abstract overview page class with common functionality (like changing and asserting tab). But this turned into a world of pain, so I've just gone with a separate page class per tab, and some of them have duplicated methods/functionality. Its the best I could do, and is better than it currently is in respect of breaking up the god class.


mattwills-moj-digital (approved):
AToAB

chrimesdev (approved):

PR #267, opened 2023-10-25
RR-450 make the boxes full width of the page on print

chrimesdev:
## Description

On print only, make the boxes full width of the page

https://dsdmoj.atlassian.net/browse/RR-450

This also should fix https://dsdmoj.atlassian.net/browse/RR-449

### DIFF
--- a/assets/scss/print.scss
+++ b/assets/scss/print.scss
@@ -31,6 +31,10 @@
   margin-top: 8.5mm;
 }
 
+.app-u-print-full-width {
+  width: 100%;
+}
+
 .connect-dps-common-footer {
   @include govuk-visually-hidden();
 }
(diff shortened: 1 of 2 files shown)


assets/scss/print.scss
```diff
@@ -31,6 +31,18 @@
   margin-top: 8.5mm;
 }
 
+.app-u-print-full-width {
+  width: 100%;
```
nathanrussell-moj-digital, line 35: Does this need a `!important` or similar?
Just wondering which will take precedence when used with other width based classes?
```
<div class="govuk-grid-column-two-thirds app-u-print-full-width">
```

chrimesdev: Usually with a utility class you probably would use `!important` but in this case in print, the `print.css` file is taking precedence and `app-u-print-full-width` has higher specificity than `govuk-grid-column-two-thirds`

nathanrussell-moj-digital: I cant claim to fully understand how that works, but 🤷‍♂️ , if you are happy then I'm happy 😁

nathanrussell-moj-digital (approved):

PR #273, opened 2023-10-26
RR-451 use N/A for empty state in release date field

chrimesdev:
## Description 

When there is no data on a release date in the header, the header should read “Release date: N/A”

https://dsdmoj.atlassian.net/browse/RR-451


nathanrussell-moj-digital (approved):
LGTM 👍

PR #281, opened 2023-10-30
RR-439 - Correctly set the last updated by/at fields on the Work & Interest tab

nathanrussell-moj-digital:
This PR correctly sets the last updated by/at fields on the Work & Interest tab, based on whether the main 'top level' induction, or the future Work Interests were the last thing to be updated.

### DIFF
--- a/server/data/mappers/workAndInterestMapper.ts
+++ b/server/data/mappers/workAndInterestMapper.ts
@@ -92,6 +92,11 @@ const toLongQuestionSetWorkInterests = (ciagInduction: CiagInduction): WorkInter
 }
 
 const toShortQuestionSetWorkInterests = (ciagInduction: CiagInduction): WorkInterests => {
+  const mostRecentlyUpdatedSection: 'MAIN_INDUCTION' | 'WORK_INTERESTS' =
+    ciagInduction.modifiedDateTime >= ciagInduction.inPrisonInterests.modifiedDateTime
+      ? 'MAIN_INDUCTION'
+      : 'WORK_INTERESTS'
+
   return {
     hopingToWorkOnRelease: ciagInduction.hopingToGetWork,
     longQuestionSetAnswers: undefined,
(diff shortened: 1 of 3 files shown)


chrimesdev:
This approach makes sense to me 👍🏻

server/data/mappers/workAndInterestMapper.ts
```diff
@@ -65,6 +65,11 @@ const toWorkExperience = (ciagInduction: CiagInduction): WorkExperience => {
 }
 
 const toLongQuestionSetWorkInterests = (ciagInduction: CiagInduction): WorkInterests => {
+  const mostRecentlyUpdatedSection: 'MAIN_INDUCTION' | 'WORK_INTERESTS' =
+    ciagInduction.modifiedDateTime >= ciagInduction.workExperience.workInterests.modifiedDateTime
```
mattwills-moj-digital, line 69: This is ok, but personally I would have found an if statement quicker to grasp. It's a shame typescript doesn't have when statements!

mattwills-moj-digital:
I also think this approach is perfectly reasonable.

A related question...what happens if say the `workAndInterests` are updated, but the main induction isn't? Is the `modifiedDateTime` of the main induction updated?

nathanrussell-moj-digital:
> A related question...what happens if say the workInterests are updated, but the main induction isn't? Is the modifiedDateTime of the main induction updated?

You'd probably be closer to me than this as thats a (CIAG) API question! 😁 
But my understanding is that if you updated workInterests, then the CIAG API updated by/at fields for the workInterests object are updated, but the main induction ones (top level) are not.
That's certainly the basis of my logic to be fair, and is what I've experienced from the (limited) testing I've done

mattwills-moj-digital:
> You'd probably be closer to me than this as thats a (CIAG) API question! 😁 But my understanding is that if you updated workInterests, then the CIAG API updated by/at fields for the workInterests object are updated, but the main induction ones (top level) are not. That's certainly the basis of my logic to be fair, and is what I've experienced from the (limited) testing I've done

I might be closer to the API, but we both know that the API behaviour here is driven by the UI! 😂

mattwills-moj-digital (approved):
LGTM

nathanrussell-moj-digital:
> I might be closer to the API, but we both know that the API behaviour here is driven by the UI! 😂

Sad but true! 🤷‍♂️

PR #284, opened 2023-10-30
RR-441 duplicate the updated at/by in education and training 

chrimesdev:
## Description

Move the `Updated at` & `Updated by` values out of the parent and duplicate them into the `EducationAndTrainingLongQuestionSet` and `EducationAndTrainingShortQuestionSet`. 

Then separate out the `inPrisonInterestsEducation` to allow for individual `updatedAt` and `updatedBy` for the "Training and education interests in prison" section

The data displayed remains the same as before, until we do the work for the mapping/data sources.


nathanrussell-moj-digital (approved):
Yes; perfect; this gives us the correct level of separation and will allow us to do the correct mapping (in the next PR 👍 )
Nice work 👍

PR #288, opened 2023-11-01
RR-460 - Corrected mapping and rendering of work interest job types and specific job roles

nathanrussell-moj-digital:
## Description
This PR corrects the mapping from the CIAG Induction to the view model for the "Work interests" section of the "Work and interests" tab (long question set)

### Background
Previously "Type of work" and "Specific job role of interest" were mapped and treated separately which resulted in a weird mismatch between the 2 fields on our screen, and some apparent "empty state" requirements (which is what the story was written up to address)
It turns out the issue was less about empty states (though that is addressed and covered in the "Specific job role of interest" field), and more about a lack of understanding about how the CIAG API returned this data and how 2 fields relate to each other.

In the CIAG API response, we can get data back that looks like this:
```
...
      workInterests: {
        modifiedBy: options?.workInterestModifiedBy || 'ANOTHER_DPS_USER_GEN',
        modifiedDateTime: options?.workInterestModifiedByDateTime || '2023-08-22T11:12:31.943Z',
        workInterests: ['RETAIL', 'CONSTRUCTION', 'OTHER'],
        particularJobInterests: [
          {
            workInterest: 'CONSTRUCTION',
            role: 'General labourer',
          },
          {
            workInterest: 'OTHER',
            role: 'Being a stunt double for Tom Cruise, even though he does all his own stunts',
          },
        ],
      },
...
```

Previously we mapped and rendered this incorrectly, in that we mapped `particularJobInterests.workInterest` to our "Type of work" field, and mapped `particularJobInterests.role` to our "Specific job role of interest" field. We ignored and did not map from `workInterests`
This means that (in the above example) we would not render and would have lost `RETAIL`. Further, if the CIAG and prisoner had not entered any specific job roles (its an optional field on the CIAG Induction, and would have meant the array `particularJobInterests` was empty) then we incorrectly rendered both "Type of work" and "Specific job role of interest" fields on our page as empty (hence the apparent need for "empty state" handling!)

### What we should have done
It turns out that the way the CIAG Induction API fields `workInterests` and `particularJobInterests` work is that they work together, and `particularJobInterests` is a subset of `workInterests`.

`workInterests` is a simple array of enums representing the checkboxes on the CIAG Induction which ask about the types of job that a prisoner is interesting in doing after release. That is the set of job interests, and is what we should map into our "Type of work" field. Worth noting also that this field (`workInterests` - ie. the checkbox of job types) is a mandatory field and can never be null or empty. There is no "empty state" requirement for this field - it's not a valid scenario.

`particularJobInterests` is an array that contains the specific job roles that the CIAG / prisoner entered - one for each selected checkbox of job types. The difference here however (and the source of our misunderstanding) is that the job role field for each job type is not mandatory. They do not need to enter any on this screen.

### Solution
What we need to do (and is what this PR does) is map, and effectively flatten, both fields into a single object array in our view model, rather than treating them as two field mappings.
We need to iterate through the `workInterests` array, and for each one, if there is a corresponding element in the `particularJobInterests` array map that specific job role, otherwise `undefined`

I've changed our view model so that we no longer have separate arrays for `jobTypes` and `specificJobRoles`, and instead have an array `jobs` which contains an object of `jobType` and `specificJobRole` (singular).
Therefore, using the example data from the CIAG API above, our resultant mapping becomes:
```
...
jobs: [
  {
    jobType: 'RETAIL',
    specificJobRole: undefined,
  },
  {
    jobType: 'CONSTRUCTION',
    specificJobRole: 'General labourer',
  },
  {
    jobType: 'OTHER',
    specificJobRole: 'Being a stunt double for Tom Cruise, even though he does all his own stunts',
  },
]
...
```

In the nunjucks view we handle `undefined` in the array of specific job roles by rendering as `None` (as per the ticket requirements). This is the "empty state" case.

### Screenshot
This approach also has the added benefit that in our rendering the list of "Type of work" and "Specific job role of interest" are sequenced correctly. The screenshot below shows 3 job roles, but only one of them had a specific job role (`Animal care and farming` was set to `Vet`. The other 2 were left blank in the CIAG Induction)

Screenshot 2023-11-01 at 08 39 01

### DIFF
--- a/server/routes/overview/workAndInterestsView.ts
+++ b/server/routes/overview/workAndInterestsView.ts
@@ -7,16 +7,11 @@ export default class WorkAndInterestsView {
     tab: string
     prisonerSummary: PrisonerSummary
     workAndInterests: WorkAndInterests
-    hasAtLeastOneSpecificJobRole: boolean
   } {
     return {
       tab: 'work-and-interests',
       prisonerSummary: this.prisonerSummary,
       workAndInterests: this.workAndInterests,
-      hasAtLeastOneSpecificJobRole:
-        this.workAndInterests.data?.workInterests.longQuestionSetAnswers?.jobs.filter(
-          jobInterest => jobInterest.specificJobRole,
-        ).length > 0,
     }
   }
 }
--- a/server/routes/overview/workAndInterestsView.ts
+++ b/server/routes/overview/workAndInterestsView.ts
@@ -7,11 +7,16 @@ export default class WorkAndInterestsView {
     tab: string
     prisonerSummary: PrisonerSummary
     workAndInterests: WorkAndInterests
+    hasAtLeastOneSpecificJobRole: boolean
   } {
     return {
       tab: 'work-and-interests',
       prisonerSummary: this.prisonerSummary,
       workAndInterests: this.workAndInterests,
+      hasAtLeastOneSpecificJobRole:
+        this.workAndInterests.data?.workInterests.longQuestionSetAnswers?.jobs.filter(
+          jobInterest => jobInterest.specificJobRole,
+        ).length > 0,
     }
   }
 }
--- a/server/views/pages/overview/partials/workAndInterestsTab/_inductionLongQuestionSet.njk
+++ b/server/views/pages/overview/partials/workAndInterestsTab/_inductionLongQuestionSet.njk
@@ -147,15 +147,11 @@ questions are different.
             Specific job role of interest
           </dt>
           <dd class="govuk-summary-list__value">
-            {% if hasAtLeastOneSpecificJobRole %}
-              <ul class="govuk-list">
-                {% for job in workAndInterests.data.workInterests.longQuestionSetAnswers.jobs %}
-                  <li>{{ job.specificJobRole or 'None' }}</li>
-                {% endfor %}
-              </ul>
-            {% else %}
-              None
-            {% endif %}
+            <ul class="govuk-list">
+              {% for job in workAndInterests.data.workInterests.longQuestionSetAnswers.jobs %}
+                <li>{{ job.specificJobRole or 'None' }}</li>
+              {% endfor %}
+            </ul>
           </dd>
           <dd class="govuk-summary-list__actions govuk-!-display-none-print">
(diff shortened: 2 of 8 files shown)


server/routes/overview/workAndInterestsView.ts
```diff
@@ -7,11 +7,16 @@ export default class WorkAndInterestsView {
     tab: string
     prisonerSummary: PrisonerSummary
     workAndInterests: WorkAndInterests
+    hasAtLeastOneSpecificJobRole: boolean
   } {
     return {
       tab: 'work-and-interests',
       prisonerSummary: this.prisonerSummary,
       workAndInterests: this.workAndInterests,
+      hasAtLeastOneSpecificJobRole:
```
chrimesdev, line 16: Good way of checking for this!

chrimesdev (approved):
Looks good, nice work and thorough description!

server/views/pages/overview/partials/workAndInterestsTab/_inductionLongQuestionSet.njk
```diff
@@ -145,11 +147,15 @@ questions are different.
             Specific job role of interest
           </dt>
           <dd class="govuk-summary-list__value">
-            <ul class="govuk-list">
-              {% for specificJobRole in workAndInterests.data.workInterests.longQuestionSetAnswers.specificJobRoles %}
-                <li>{{ specificJobRole }}</li>
-              {% endfor %}
-            </ul>
+            {% if hasAtLeastOneSpecificJobRole %}
```
nathanrussell-moj-digital, line 150: I'm not sure I need this outer if check actually .... I think the array will always have at least 1 element in it, and the nunjucks ternary `{{ job.specificJobRole or 'None' }}` will address it ... 🤔 

Just going to make that change locally and see what happens ....

nathanrussell-moj-digital: Yup, didn't need it 😁 👍 
Just pushing a commit that removes this outer if statement and simplifies the view object slightly 👍

server/routes/overview/workAndInterestsView.ts
nathanrussell-moj-digital: Even though I've just nuked it as we don't need it anymore! 🤣

chrimesdev (approved):

PR #292, opened 2023-11-02
RR-459 - Sort a prisoner's work interests based on job type

nathanrussell-moj-digital:
This PR implements a comparator function to sort a prisoners post-release work interests alphabetically by job type with Other always at the end.
There was a little related refactoring that needed doing; plus testing locally identified a related problem with how and where we map/render "Other" for these future work interests.

Screenshot 2023-11-02 at 08 32 30

### DIFF
--- a/server/@types/viewModels/index.d.ts
+++ b/server/@types/viewModels/index.d.ts
@@ -173,30 +173,32 @@ declare module 'viewModels' {
     updatedAt: Date
   }
 
+  export interface WorkInterestJob {
+    jobType:
+      | 'OUTDOOR'
+      | 'CONSTRUCTION'
+      | 'DRIVING'
+      | 'BEAUTY'
+      | 'HOSPITALITY'
+      | 'TECHNICAL'
+      | 'MANUFACTURING'
+      | 'OFFICE'
+      | 'RETAIL'
+      | 'SPORTS'
+      | 'WAREHOUSING'
+      | 'WASTE_MANAGEMENT'
+      | 'EDUCATION_TRAINING'
+      | 'CLEANING_AND_MAINTENANCE'
+      | 'OTHER'
+    specificJobRole?: string
+  }
+
   export interface WorkInterestsLongQuestionSet {
     constraintsOnAbilityToWork: Array<
       'CARING_RESPONSIBILITIES' | 'LIMITED_BY_OFFENSE' | 'HEALTH_ISSUES' | 'NO_RIGHT_TO_WORK' | 'OTHER' | 'NONE'
     >
     otherConstraintOnAbilityToWork?: string
-    jobs: Array<{
-      jobType:
-        | 'OUTDOOR'
-        | 'CONSTRUCTION'
-        | 'DRIVING'
-        | 'BEAUTY'
-        | 'HOSPITALITY'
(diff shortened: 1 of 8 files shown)


server/@types/viewModels/index.d.ts
```diff
@@ -173,30 +173,33 @@ declare module 'viewModels' {
     updatedAt: Date
   }
 
+  export interface WorkInterestJob {
```
nathanrussell-moj-digital, line 176: Exporting this as a type was more convenient (for the comparator function and its tests) than defining it inline, which we did previously. It means the comparator can refer to its type 👍

server/@types/viewModels/index.d.ts
```diff
@@ -173,30 +173,33 @@ declare module 'viewModels' {
...
+      | 'MANUFACTURING'
+      | 'OFFICE'
+      | 'RETAIL'
+      | 'SPORTS'
+      | 'WAREHOUSING'
+      | 'WASTE_MANAGEMENT'
+      | 'EDUCATION_TRAINING'
+      | 'CLEANING_AND_MAINTENANCE'
+      | 'OTHER'
+    otherJobType?: string
```
nathanrussell-moj-digital, line 193: Also added `otherJobType` which was missing previously

server/data/mappers/jobComparator.ts
```diff
@@ -24,4 +24,28 @@ const jobComparator = (left: Job, right: Job): number => {
...
-export default jobComparator
+/**
+ * Comparator function that compares two `WorkInterestJob` view model instances, returning -1, 0 or 1 depending on whether the
+ * first job's type property is alphabetically before, equal or after the second job's type property.
+ * If the first job's type is 'OTHER' then 1 is returned.
+ *
+ * This comparator function can be used to sort an array of `WorkInterestJob` view model instances, resulting in the array being sorted
+ * alphabetically on the `type` property, except 'OTHER' which will always be at the end of the array.
+ */
+const workInterestJobComparator = (left: WorkInterestJob, right: WorkInterestJob): number => {
```
nathanrussell-moj-digital, line 35: Rather than put this in a separate source file (eg: `workInterestJobComparator.ts`) I opted to put this in `jobComparator.ts` as the 2 functions (`jobComparator` and `workInterestJobComparator` are so closely related. They operate on different types of objects, but essentially they do the same thing, and the object types are both "jobs")

server/views/pages/overview/partials/workAndInterestsTab/_inductionLongQuestionSet.njk
```diff
@@ -130,7 +130,11 @@ questions are different.
           <dd class="govuk-summary-list__value">
             <ul class="govuk-list">
               {% for job in workAndInterests.data.workInterests.longQuestionSetAnswers.jobs %}
-                <li>{{ job.jobType | formatJobType }}</li>
+                {% if job.jobType === 'OTHER' %}
```
nathanrussell-moj-digital, line 133: The "other" logic and prefixing of `Other - ` should have been here, not where we had it previously (that was my bad in an earlier PR)

chrimesdev (approved):

PR #298, opened 2023-11-02
RR-454 associate textarea with warning message

chrimesdev:
## Description

Associate `textarea` of the Note with the warning message.

Had to move away from the Nunjucks macro here as the Nunjuck macro does not allow you to pass an `id` (because we're not using the correct component in the designs)

https://dsdmoj.atlassian.net/browse/RR-454


nathanrussell-moj-digital (approved):
👍

PR #299, opened 2023-11-02
RR-455 display visually hidden table column

chrimesdev:
## Description

Display visually hidden table column to all users as this is likely to be confusing for several user groups, in particular low-vision screen reader users, who will be able to partially see the screen and not understand why the screen reader is reading out elements that aren’t visible.

https://dsdmoj.atlassian.net/browse/RR-455


nathanrussell-moj-digital (approved):
👍

PR #301, opened 2023-11-02
RR-452 improve the Breadcrumb navigation markup

chrimesdev:
## Description 

Improve the Breadcrumb navigation markup, by using a semantic `<nav>` element and `aria-label="Breadcrumb"`

As mentioned in the Accessibility report, this is actually different to how GDS markup the Breadcrumb and needs to be reported.

Interestingly its the same as how the NHS Design System does it!

https://dsdmoj.atlassian.net/browse/RR-452


nathanrussell-moj-digital (approved):
👍

PR #312, opened 2023-11-07
RR-467 - Corrected default sort order; maintain sort order in session

nathanrussell-moj-digital:
This PR corrects the default sort order on the Prisoner List page (should have been reception-date descending, as per the ticket)
Also (as per the ticket), persist the last used sort order in the session, so that when the CIAG goes to an Induction or Action Plan of a prisoner, when they come back to the Prisoner List page their last sort order is maintained.


mattwills-moj-digital (approved):
GS

PR #315, opened 2023-11-09
RR-436 - Cypress tests for sorting and filtering

nathanrussell-moj-digital:
This PR adds cypress tests for sorting and filtering operations on the Prisoner List page.

It also makes a change to the nunjucks markup and view file so that the count of records is always present, even if there is only 1 page to display:

Screenshot 2023-11-09 at 08 15 50

### DIFF
--- a/integration_tests/e2e/prisonerList/prisonerList.cy.ts
+++ b/integration_tests/e2e/prisonerList/prisonerList.cy.ts
@@ -12,7 +12,6 @@ import PrisonerListPage from '../../pages/prisonerList/PrisonerListPage'
  * in order to set specific stubs in the Action Plan API for the specific prisoner clicked on.
  */
 context(`Display the prisoner list screen`, () => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   let prisonerSearchSummaries: Array<PrisonerSearchSummary>
 
   beforeEach(() => {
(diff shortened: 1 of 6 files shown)


integration_tests/e2e/prisonerList/prisonerList.cy.ts
```diff
@@ -36,48 +35,121 @@ context(`Display the prisoner list screen`, () => {
...
+      prisonerListPage.hasNoResultsDisplayed()
+    })
+
+    it('should filter for prisoners named John', () => {
+      // Given
+      cy.signIn()
+      cy.visit('/')
+      const prisonerListPage = Page.verifyOnPage(PrisonerListPage)
+
+      const numberOfPrisonersNamedJohn = prisonerSearchSummaries.filter(
```
nathanrussell-moj-digital, line 70: There will be some John's in the `prisonerSearchSummaries`, but given that it was randomly generated (725 records) we don't know how many of them have John in their name, so we need to work it out

integration_tests/e2e/prisonerList/prisonerList.cy.ts
```diff
@@ -36,48 +35,121 @@ context(`Display the prisoner list screen`, () => {
...
+        .isSortedBy('reception-date', 'descending')
+    })
+
+    it('should sort by location descending', () => {
+      // Given
+      cy.signIn()
+      cy.visit('/')
+      const prisonerListPage = Page.verifyOnPage(PrisonerListPage)
+
+      const expectedFirstRowInTable = prisonerSearchSummaries
```
nathanrussell-moj-digital, line 128: As with the "filtering by John" test above, given that `prisonerSearchSummaries` was randomly generated we need to sort it in the test here to work out what the first row should be

integration_tests/mockData/prisonerSearchSummaryMockDataGenerator.ts
```diff
@@ -128,6 +128,11 @@ const FORENAMES = [
   'JIMMY',
   'STEVE',
   'STEPHEN',
+  'JOHN',
```
nathanrussell-moj-digital, line 131: Added some extra names, including John as a forename and Johnson as a surname

server/routes/prisonerList/prisonerListController.test.ts
```diff
@@ -72,14 +72,7 @@ describe('prisonerListController', () => {
...
         sortOrder: 'descending',
-        items: [
-          {
-            href: '?sort=reception-date,descending&page=1',
-            selected: true,
-            text: '1',
-            type: undefined,
-          },
-        ],
+        items: [],
```
nathanrussell-moj-digital, line 75: I made a change to the View class (👇 ) so that if there is only 1 record in the items array (and therefore it must be page 1 / current page), then return an empty array instead.
This was to prevent a lone `1` being displayed as the pagination controls, and instead there are no pagination controls and just the record count (see screenshot in main description)

server/routes/prisonerList/prisonerListView.ts
```diff
@@ -63,6 +63,11 @@ const buildItemsArray = (config: {
   sortOrder: string
 }): Item[] => {
   const items: Item[] = []
+
```
nathanrussell-moj-digital, line 66: I made this change to the View class so that if there is only 1 record in the items array (and therefore it must be page 1 / current page), then return an empty array instead.
This was to prevent a lone `1` being displayed as the pagination controls, and instead there are no pagination controls and just the record count (see screenshot in main description)

server/views/pages/prisonerList/partials/searchTable.njk
```diff
@@ -85,14 +85,14 @@
       </table>
     </form>
 
-    {% if renderPaginationControls %}
+    <section data-qa="prisoner-list-pagination">
```
nathanrussell-moj-digital, line 88: We now display the pagination all of the time (in a named section so that we can address it in the cypress tests - `mojPagination` macro does not allow you to set an `id` or `attributes` !!)

Anyway, we display it all of the time now so that we always get the result/record count
If there are no `items` (page links) then it neatly displays nothing for them 👍

PR #318, opened 2023-11-09
RR-444 - Service method to determine if a CIAG Induction exists or not

nathanrussell-moj-digital:
RR-444 needs to show different content on the Overview screen depending on whether there is a CIAG induction or not for the prisoner.
In order to do this we need a service method that determines if there is a CIAG Induction or not for the specified prison number.

This PR introduces the service method (but does not wire it in yet)

### DIFF
--- a/server/services/ciagInductionService.ts
+++ b/server/services/ciagInductionService.ts
@@ -13,7 +13,6 @@ export default class CiagInductionService {
       const ciagInduction = await this.getCiagInduction(prisonNumber, token)
       return toWorkAndInterests(ciagInduction)
     } catch (error) {
-      logger.error(`Error retrieving data from CIAG Induction API: ${JSON.stringify(error)}`)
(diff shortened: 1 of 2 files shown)


server/services/ciagInductionService.ts
```diff
@@ -13,7 +13,6 @@ export default class CiagInductionService {
       const ciagInduction = await this.getCiagInduction(prisonNumber, token)
       return toWorkAndInterests(ciagInduction)
     } catch (error) {
-      logger.error(`Error retrieving data from CIAG Induction API: ${JSON.stringify(error)}`)
```
nathanrussell-moj-digital, line 16: I moved the log messages down into the private function; otherwise we would have had 3 occurrences of this same error logging (IE. I would have added it to the new `ciagInductionExists` function otherwise)

mattwills-moj-digital (approved):
LGTM

PR #320, opened 2023-11-10
RR-468 cypress tests for the pagination controls

chrimesdev:
## Description

Cypress tests for the pagination controls

https://dsdmoj.atlassian.net/browse/RR-468

### DIFF
--- a/integration_tests/e2e/prisonerList/prisonerList.cy.ts
+++ b/integration_tests/e2e/prisonerList/prisonerList.cy.ts
@@ -148,6 +148,40 @@ context(`Display the prisoner list screen`, () => {
         .firstRowLocationIs(expectedFirstRowInTable.location)
     })
   })
+
+  describe('pagination', () => {
+    it('should display pagination controls, displaying 1 to 10 with the next link, on the first page of the prisoner list', () => {
+      // Given
+      cy.signIn()
+      cy.visit('/')
+
+      // When
+      const prisonerListPage = Page.verifyOnPage(PrisonerListPage)
+
+      // Then
+      prisonerListPage //
+        .hasPaginationControlsDisplayed()
+        .hasNumberOfPagesDisplayed(10)
+        .hasNextLinkDisplayed()
+    })
+
+    it('should display pagination controls, displaying 2 to 11 with the next and previous links, on the 7th page of the prisoner list', () => {
+      // Given
+      cy.signIn()
+      cy.visit('/')
+
+      // When
+      const prisonerListPage = Page.verifyOnPage(PrisonerListPage)
+
+      // Then
+      prisonerListPage //
+        .hasPaginationControlsDisplayed()
+        .hasNumberOfPagesDisplayed(10)
+        .setPaginationNumber(7)
+        .hasPreviousLinkDisplayed()
+        .hasNextLinkDisplayed()
+    })
+  })
 })
 
 const capitalize = (name: string): string => {
--- a/integration_tests/e2e/prisonerList/prisonerList.cy.ts
+++ b/integration_tests/e2e/prisonerList/prisonerList.cy.ts
@@ -160,8 +160,15 @@ context(`Display the prisoner list screen`, () => {
 
       // Then
       prisonerListPage //
-        .hasPaginationControlsDisplayed()
-        .hasNumberOfPagesDisplayed(10)
+        .paginationCurrentPageIs(1)
+        .hasPaginationLinkForPage(2)
+        .hasPaginationLinkForPage(3)
+        .hasPaginationLinkForPage(4)
+        .hasPaginationLinkForPage(5)
+        .hasPaginationLinkForPage(6)
+        .hasPaginationLinkForPage(8)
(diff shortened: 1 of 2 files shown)


integration_tests/pages/prisonerList/PrisonerListPage.ts
```diff
@@ -76,6 +76,33 @@ export default class PrisonerListPage extends Page {
...
+
+  hasNextLinkDisplayed(): PrisonerListPage {
+    this.paginationControls().find('li:last-of-type a').should('contain', 'Next')
+    return this
+  }
+
+  hasNumberOfPagesDisplayed(expected: number): PrisonerListPage {
+    this.paginationControls()
+      .find('li.moj-pagination__item')
+      .should('have.length', expected + 1)
```
chrimesdev, line 102: Need to improve this - will fix now

integration_tests/pages/prisonerList/PrisonerListPage.ts
```diff
@@ -76,6 +76,35 @@ export default class PrisonerListPage extends Page {
...
+
+  hasNextLinkDisplayed(): PrisonerListPage {
+    this.paginationControls().find('li:last-of-type a').should('contain', 'Next')
+    return this
+  }
+
+  hasNumberOfPagesDisplayed(expected: number): PrisonerListPage {
+    this.paginationControls()
+      .find('li.moj-pagination__item')
+      .not('.moj-pagination__item--next')
```
chrimesdev, line 102: Maybe not the best approach but only way I could think of without having the ability to pass attributes in the Nunjucks macros!

nathanrussell-moj-digital: It's fine; I agree, we don't want to litter the nunjucks markup with too many attributes etc just to support tests.

integration_tests/pages/prisonerList/PrisonerListPage.ts
```diff
@@ -76,6 +76,35 @@ export default class PrisonerListPage extends Page {
...
+    return this
+  }
+
+  setPaginationNumber(value: number): PrisonerListPage {
+    this.paginationControls().find(`li:nth-of-type(${value}) a`).click()
+    return this
+  }
+
+  hasPreviousLinkDisplayed(): PrisonerListPage {
+    this.paginationControls().find('li:first-of-type a').should('contain', 'Previous')
```
nathanrussell-moj-digital, line 90: Could you use your `paginationFirstLink` method here?
```
this.paginationFirstLink().should('contain', 'Previous')
```

Would that work? And then do the same thing for the next link displayed?

integration_tests/e2e/prisonerList/prisonerList.cy.ts
```diff
@@ -148,6 +148,40 @@ context(`Display the prisoner list screen`, () => {
...
+
+      // Then
+      prisonerListPage //
+        .hasPaginationControlsDisplayed()
+        .hasNumberOfPagesDisplayed(10)
+        .setPaginationNumber(7)
+        .hasPreviousLinkDisplayed()
+        .hasNextLinkDisplayed()
+    })
+  })
```
nathanrussell-moj-digital, line 184: I'm not sure this is what I had in mind .... I think I was thinking more something like this (most of it is pseudo code)
```
    it('should display pagination controls, displaying 2 to 11 with the next and previous links, on the 7th page of the prisoner list', () => {
      // Given
      cy.signIn()
      cy.visit('/')
      const prisonerListPage = Page.verifyOnPage(PrisonerListPage)

      // When
      prisonerListPage.gotoPage(7) // I think "gotoPage" is a better name than your current "setPaginationNumber" method, but its basically the same thing

      // Then
      prisonerListPage //
        .doesNotHavePaginationLinkForPage(1)
        .hasPreviousLinkDisplayed()
        .hasPaginationLinkForPage(2)
        .hasPaginationLinkForPage(3)
        .hasPaginationLinkForPage(4)
        .hasPaginationLinkForPage(5)
        .hasPaginationLinkForPage(6)
        .paginationCurrentPageIs(7)
        .hasPaginationLinkForPage(8)
        .hasPaginationLinkForPage(9)
        .hasPaginationLinkForPage(10)
        .hasPaginationLinkForPage(11)
        .hasPaginationLinkForPage(12)
        .hasNextLinkDisplayed()
        .doesNotHavePaginationLinkForPage(13)
    })
```
Something like that perhaps?
There might be better ways to express that we are expecting a specific set of pagination controls,  but you get the idea

chrimesdev: Will revisit!

integration_tests/pages/prisonerList/PrisonerListPage.ts
chrimesdev: Good spot, just realised that `paginationFirstLink` isn't even used currently - maybe I started this trail of thought and got lost!

nathanrussell-moj-digital (approved):
LGTM

PR #321, opened 2023-11-10
RR-471 content change highest level of education

chrimesdev:
## Description

Content change Highest level of education -> Highest level of education before prison

https://dsdmoj.atlassian.net/browse/RR-471


chrimesdev:
Can't work git today 😮‍💨!

PR #325, opened 2023-11-13
RR-458 change phase banner to BETA and add feedback content and link

chrimesdev:
## Description

Change phase banner to BETA and add feedback content and link

https://dsdmoj.atlassian.net/browse/RR-458

### Screenshot

Screenshot 2023-11-13 at 09 09 01

### DIFF
--- a/helm_deploy/values-dev.yaml
+++ b/helm_deploy/values-dev.yaml
@@ -22,7 +22,6 @@ generic-service:
     CIAG_INDUCTION_UI_URL: "http://ciag-induction-dev.hmpps.service.justice.gov.uk"
     DPS_URL: "https://digital-dev.prison.service.justice.gov.uk"
     FRONTEND_COMPONENT_API_URL: "https://frontend-components-dev.hmpps.service.justice.gov.uk"
-    FEEDBACK_URL: "https://eu.surveymonkey.com/r/H9CYM7B"
     ENVIRONMENT_NAME: "DEV"
 
   allowlist:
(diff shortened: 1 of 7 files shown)


server/views/partials/layout.njk
```diff
@@ -92,13 +92,14 @@
...
     {{ govukPhaseBanner({
-      attributes: {
-        role: "complementary"
-      },
       tag: {
-        text: "beta"
+        text: "Beta"
       },
-      classes: "govuk-!-display-none-print"
+      html: 'This is a new service – get help or <a class="govuk-link" href="https://eu.surveymonkey.com/r/H9CYM7B" target="_blank" referrerpolicy="no-referrer" rel="noopener">give your feedback</a> to help us to improve it.',
```
nathanrussell-moj-digital, line 98: We don't want the URL hardcoded and embedded in the nunjucks view like this (I can almost guarantee they'll change it at some point! 🙄 )
The feedback URL needs to be an environment variable in the helm charts (it can be hardcoded there for the time being 👍 ), which in turn sets a property in the `config` object, which in turn is made available to the view layer.

Then the `html` property above becomes:
```
html: 'This is a new service – get help or <a class="govuk-link" href="{{ feebackUrl }}" target="_blank" referrerpolicy="no-referrer" rel="noopener">give your feedback</a> to help us to improve it.',
```

helm_deploy/values-dev.yaml
```diff
@@ -22,6 +22,7 @@ generic-service:
     CIAG_INDUCTION_UI_URL: "http://ciag-induction-dev.hmpps.service.justice.gov.uk"
     DPS_URL: "https://digital-dev.prison.service.justice.gov.uk"
     FRONTEND_COMPONENT_API_URL: "https://frontend-components-dev.hmpps.service.justice.gov.uk"
+    FEEDBACK_URL: "https://eu.surveymonkey.com/r/H9CYM7B"
```
nathanrussell-moj-digital, line 25: This is the right idea , but instead of defining it in each environment helm chart you can do it once in the "top level / shared" helm chart
`helm_deploy/hmpps-education-and-work-plan-ui/values.yaml`

nathanrussell-moj-digital (approved):
LGTM 👍

PR #328, opened 2023-11-14
RR-340 use a character count component on the add/update goal flows

chrimesdev:
## Description 

Use a character count component with a text limit of 512 on the Add & Update a goal flow.

There was also an issue with the spacing on the MoJ Add another component conflicting with the GOVUK error message styling; so there is an admittedly horrible CSS fix in here to resolve it.

Before: 

Screenshot 2023-11-14 at 12 09 54

After:

Screenshot 2023-11-14 at 12 09 48

https://dsdmoj.atlassian.net/browse/RR-340

### DIFF
--- a/assets/scss/components/_add-another.scss
+++ b/assets/scss/components/_add-another.scss
@@ -1,0 +1,5 @@
+// Fix an issue with spacing when there is a GOVUK character count component 
+// with an error message inside a MOJ add another component
+.moj-add-another__item:has(.govuk-character-count):has(.govuk-error-message) .moj-add-another__title {
+  padding-left: 20px;
+}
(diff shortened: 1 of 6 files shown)


assets/scss/components/_add-another.scss
```diff
@@ -0,0 +1,5 @@
+// Fix an issue with spacing when there is a GOVUK character count component 
+// with an error message inside a MOJ add another component
+.moj-add-another__item:has(.govuk-character-count):has(.govuk-error-message) .moj-add-another__title {
```
chrimesdev, line 3: This is horrible, I know, but it was the only way I could think of!

nathanrussell-moj-digital (approved):
LGTM 👍

PR #330, opened 2023-11-14
RR-484 - Add app insights client code

nathanrussell-moj-digital:
This PR adds the Application Insights client code to the rendered page markup that the application serves.
It adds it in the `layout.njk` file (so that all pages get it) and it adds it to the bottom of the page with all the other javascript includes so as to not block rendering of the page.

### DIFF
--- a/assets/js/applicationinsights.js
+++ b/assets/js/applicationinsights.js
@@ -1,0 +1,22 @@
+window.applicationInsights = (function () {
+  let appInsights
+
+  return {
+    init: (connectionString, applicationInsightsRoleName, authenticatedUser) => {
+      if (!appInsights && connectionString) {
+        appInsights = new Microsoft.ApplicationInsights.ApplicationInsights({
+          config: {
+            connectionString,
+            autoTrackPageVisitTime: true,
+          },
+        })
+        appInsights.addTelemetryInitializer(envelope => {
+          envelope.tags['ai.cloud.role'] = applicationInsightsRoleName
+        })
+        appInsights.setAuthenticatedUserContext(authenticatedUser)
+        appInsights.loadAppInsights()
+        appInsights.trackPageView()
+      }
+    },
+  }
+})()
--- a/package.json
+++ b/package.json
@@ -91,6 +91,7 @@
     ]
   },
   "dependencies": {
+    "@microsoft/applicationinsights-web": "^3.0.4",
     "@ministryofjustice/frontend": "^1.8.0",
     "agentkeepalive": "^4.3.0",
     "applicationinsights": "^2.7.0",
--- a/server/middleware/setUpStaticResources.ts
+++ b/server/middleware/setUpStaticResources.ts
@@ -1,4 +1,3 @@
-import fs from 'fs'
 import path from 'path'
 import compression from 'compression'
 import express, { Router } from 'express'
--- a/server/middleware/setUpStaticResources.ts
+++ b/server/middleware/setUpStaticResources.ts
@@ -1,3 +1,4 @@
+import fs from 'fs'
 import path from 'path'
 import compression from 'compression'
 import express, { Router } from 'express'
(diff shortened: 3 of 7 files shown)


assets/js/applicationinsights.js
```diff
@@ -0,0 +1,22 @@
+window.applicationInsights = (function () {
```
nathanrussell-moj-digital, line 1: This file adds an object `applicationInsights` to the `window` object (essentially a global variable) with a single method `init`
This syntax is a _bit like_ a class (except of course that classes don't exist in javascript - I think this is called the "revealing module pattern"). Anyway, think of it a bit like a class, which currently has one method `init`
I've written it this way because I'm fully expecting us to have to extend and build upon this in the future by adding additional methods .... such as `trackEvent`

In its current form, with the call to `trackPageView` on line 18, it will record an app insights `pageView` event for every page render.

server/middleware/setUpStaticResources.ts
```diff
@@ -22,6 +22,7 @@ export default function setUpStaticResources(): Router {
     '/node_modules/@ministryofjustice/frontend/moj/assets',
     '/node_modules/@ministryofjustice/frontend',
     '/node_modules/jquery/dist',
+    '/node_modules/@microsoft/applicationinsights-web/dist/es5',
```
nathanrussell-moj-digital, line 25: This is how we map js files in the NPM module to the static resource path

server/middleware/setUpWebSecurity.ts
```diff
@@ -25,6 +25,7 @@ export default function setUpWebSecurity(): Router {
   const styleSrc = ["'self'", (_req: Request, res: Response) => `'nonce-${res.locals.cspNonce}'`]
   const imgSrc = ["'self'", 'data:']
   const fontSrc = ["'self'"]
+  const connectSrc = ["'self'", '*.applicationinsights.azure.com']
```
nathanrussell-moj-digital, line 28: We need to configure the web security component (`helmet` 🤣 🤐 ) to allow connections outbound to the app insights domain

server/utils/nunjucksSetup.ts
```diff
@@ -31,6 +31,9 @@ export default function nunjucksSetup(app: express.Express, applicationInfo: App
   app.locals.asset_path = '/assets/'
   app.locals.applicationName = 'Digital Prison Services'
 
+  app.locals.applicationInsightsConnectionString = config.applicationInsights.connectionString
```
nathanrussell-moj-digital, line 34: Make the app insights connection string, and the application name, available to `app.locals` so that nunjucks can reference them in the template

server/views/partials/layout.njk
```diff
@@ -100,6 +100,11 @@
   <script src="/assets/govukFrontendInit.js"></script>
   <script src="/assets/moj/all.js"></script>
   <script src="/assets/print.js"></script>
+  <script src="/assets/applicationinsights-web.min.js"></script>
```
nathanrussell-moj-digital, line 103: This line loads the App Insights library code from our assets static resources path (which in turn comes from the node_modules folder)

server/views/partials/layout.njk
```diff
@@ -100,6 +100,11 @@
...
+  <script src="/assets/applicationinsights.js"></script>
```
nathanrussell-moj-digital, line 104: This line loads our custom JS file which makes the object `applicationInsights` available

server/views/partials/layout.njk
```diff
@@ -100,6 +100,11 @@
...
+  <script nonce="{{ cspNonce }}">
+    window.applicationInsights.init('{{ applicationInsightsConnectionString }}', '{{ applicationInsightsRoleName }}', '{{ user.username }}');
```
nathanrussell-moj-digital, line 106: This line calls the `init` method of `applicationInsights` , passing in the connection string, application name (cloud role name), and the authenticated user name

package.json
```diff
@@ -91,6 +91,7 @@
     ]
   },
   "dependencies": {
+    "@microsoft/applicationinsights-web": "^3.0.4",
```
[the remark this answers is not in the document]
chrimesdev: Serving it from a static resource is better overall, its a pattern we followed at the NHS too

Old article but still relevant https://csswizardry.com/2019/05/self-host-your-static-assets/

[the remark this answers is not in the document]
nathanrussell-moj-digital: Good article; thanks 👍

chrimesdev (approved):

PR #332, opened 2023-11-15
RR-470 - configured ClickAnalyticsPlugin to send click events to App Insights

nathanrussell-moj-digital:
This PR adds the MS App Insights `ClickAnalyticsPlugin` to the page so that click events are sent to App Insights (as well as page views)

Click Events are recorded in App Insights as Custom Events. This story is to record clicks to the "Print this page" link. Whilst many page click events are recorded, it is easy to query and filter for just the ones of interest:
Screenshot 2023-11-15 at 14 03 20

### DIFF
--- a/assets/js/applicationinsights.js
+++ b/assets/js/applicationinsights.js
@@ -1,13 +1,26 @@
+/* global Microsoft */
 window.applicationInsights = (function () {
   let appInsights
 
   return {
     init: (connectionString, applicationInsightsRoleName, authenticatedUser) => {
       if (!appInsights && connectionString) {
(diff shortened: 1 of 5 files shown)


assets/js/applicationinsights.js
```diff
@@ -1,13 +1,26 @@
+/* global Microsoft */
```
nathanrussell-moj-digital, line 1: A special eslint comment to declare the name `Microsoft` as a global (which it is by virtue of the App Insights library inclusions on the page)
This simply means our IDEs do not moan about the name `Microsoft` not being declared in lines 8 and 16 👍

chrimesdev (approved):

PR #334, opened 2023-11-15
RR-126 Retrieve basic data set for the Timeline and display on the UI

chrimesdev:
## Description

Retrieve basic data set for the Timeline and display on the UI

### DIFF
--- a/server/@types/educationAndWorkPlanApi/index.d.ts
+++ b/server/@types/educationAndWorkPlanApi/index.d.ts
@@ -216,6 +216,8 @@ export interface components {
        * @enum {string}
        */
       eventType:
+        | 'INDUCTION_CREATED'
+        | 'INDUCTION_UPDATED'
         | 'ACTION_PLAN_CREATED'
         | 'GOAL_CREATED'
         | 'GOAL_UPDATED'
@@ -242,6 +244,12 @@ export interface components {
        * @example 2023-06-19T09:39:44Z
        */
       timestamp: string
+      /**
+       * Format: uuid
+       * @description A correlationId for this and any other TimelineEvents that occurred at the same time (i.e. within same atomic action). For example, this could be an update to a Goal, plus one or more of its child Steps. If required, this allows a client to easily group the events together.
+       * @example 113d1833-0ce1-45ad-ab44-878c9d589358
+       */
+      correlationId: string
       /**
        * @description Contextual information that's relevant to the event in question. For example the title of a Goal that was completed.
        * @example Learn French
--- a/server/@types/viewModels/index.d.ts
+++ b/server/@types/viewModels/index.d.ts
@@ -361,6 +361,36 @@ declare module 'viewModels' {
     level: 'ENTRY_LEVEL' | 'LEVEL_1' | 'LEVEL_2' | 'LEVEL_3' | 'LEVEL_4' | 'LEVEL_5' | 'LEVEL_6' | 'LEVEL_7' | 'LEVEL_8'
   }
 
+  export interface Timeline {
+    reference?: string // TODO - Are these required/optional?
+    prisonNumber?: string // TODO - Are these required/optional?
+    events?: Array<TimelineEntry>
+  }
+
+  export interface TimelineEntry {
+    reference: string
+    sourceReference: string
+    eventType:
+      | 'ACTION_PLAN_CREATED'
+      | 'GOAL_CREATED'
+      | 'GOAL_UPDATED'
+      | 'GOAL_STARTED'
+      | 'GOAL_COMPLETED'
+      | 'GOAL_ARCHIVED'
+      | 'STEP_UPDATED'
+      | 'STEP_NOT_STARTED'
+      | 'STEP_STARTED'
+      | 'STEP_COMPLETED'
+      | 'INDUCTION_UPDATED'
+      | 'INDUCTION_CREATED'
+    prisonId: string
+    actionedBy: string
+    timestamp: string
+    correlationId: string
+    contextualInfo?: string
+    actionedByDisplayName?: string
+  }
+
   /**
    * DPS frontend components
    */
--- a/server/data/educationAndWorkPlanClient.test.ts
+++ b/server/data/educationAndWorkPlanClient.test.ts
@@ -9,6 +9,7 @@ import {
 } from '../testsupport/createGoalsRequestTestDataBuilder'
 import aValidActionPlanSummaryListResponse from '../testsupport/actionPlanSummaryListResponseTestDataBuilder'
 import aValidActionPlanSummaryResponse from '../testsupport/actionPlanSummaryResponseTestDataBuilder'
+import aValidTimelineResponse from '../testsupport/timelineTestDataBuilder'
 
 describe('educationAndWorkPlanClient', () => {
   const educationAndWorkPlanClient = new EducationAndWorkPlanClient()
@@ -205,6 +206,48 @@ describe('educationAndWorkPlanClient', () => {
       // Then
       expect(nock.isDone()).toBe(true)
       expect(actual).toEqual(expectedActionPlanSummaryListResponse)
+    })
+  })
+
+  describe('getTimeline', () => {
+    it('should get Timeline', async () => {
+      // Given
+      const prisonNumber = 'A1234BC'
+      const systemToken = 'a-system-token'
+
+      const expectedTimelineResponse = aValidTimelineResponse()
+
+      educationAndWorkPlanApi.get(`/timelines/${prisonNumber}`).reply(200, expectedTimelineResponse)
+
+      // When
+      const actual = await educationAndWorkPlanClient.getTimeline(prisonNumber, systemToken)
+
+      // Then
+      expect(nock.isDone()).toBe(true)
+      expect(actual).toEqual(expectedTimelineResponse)
+    })
+
+    it('should not get Timeline given API returns error response', async () => {
+      // Given
+      const prisonNumber = 'A1234BC'
+      const systemToken = 'a-system-token'
+
+      const expectedResponseBody = {
+        status: 500,
+        userMessage: 'An unexpected error occurred',
+        developerMessage: 'An unexpected error occurred',
+      }
+      educationAndWorkPlanApi.get(`/timelines/${prisonNumber}`).reply(500, expectedResponseBody)
+
+      // When
+      try {
+        await educationAndWorkPlanClient.getTimeline(prisonNumber, systemToken)
+      } catch (e) {
+        // Then
+        expect(nock.isDone()).toBe(true)
+        expect(e.status).toEqual(500)
+        expect(e.data).toEqual(expectedResponseBody)
+      }
     })
   })
 })
--- a/server/filters/formatTimelineEvent.ts
+++ b/server/filters/formatTimelineEvent.ts
@@ -1,0 +1,19 @@
+export default function formatTimelineEvent(value: string): string {
+  const timelineEventValue = TimelineEventValue[value as keyof typeof TimelineEventValue]
+  return timelineEventValue
+}
+
+enum TimelineEventValue {
+  ACTION_PLAN_CREATED = 'Action plan created',
+  GOAL_CREATED = 'Goal created',
+  GOAL_UPDATED = 'Goal updated',
+  GOAL_STARTED = 'Goal started',
+  GOAL_COMPLETED = 'Goal completed',
+  GOAL_ARCHIVED = 'Goal archived',
+  STEP_UPDATED = 'Step updated',
+  STEP_NOT_STARTED = 'Step not started',
+  STEP_STARTED = 'Step started',
+  STEP_COMPLETED = 'Step completed',
+  INDUCTION_UPDATED = 'Induction updated',
+  INDUCTION_CREATED = 'Induction created',
+}
--- a/server/routes/overview/overviewController.test.ts
+++ b/server/routes/overview/overviewController.test.ts
@@ -1,6 +1,6 @@
 import createError from 'http-errors'
 import moment from 'moment'
-import type { FunctionalSkills, InPrisonEducationRecords, WorkAndInterests } from 'viewModels'
+import type { FunctionalSkills, InPrisonEducationRecords, Timeline, WorkAndInterests } from 'viewModels'
 import { SessionData } from 'express-session'
 import { NextFunction, Request, Response } from 'express'
 import OverviewController from './overviewController'
@@ -16,6 +16,7 @@ import CiagInductionService from '../../services/ciagInductionService'
 import aValidLongQuestionSetWorkAndInterests from '../../testsupport/workAndInterestsTestDataBuilder'
 import aValidPrisonerSummary from '../../testsupport/prisonerSummaryTestDataBuilder'
 import { aValidShortQuestionSetEducationAndTraining } from '../../testsupport/educationAndTrainingTestDataBuilder'
+import aValidTimelineResponse from '../../testsupport/timelineTestDataBuilder'
 
 describe('overviewController', () => {
   const curiousService = {
@@ -25,6 +26,7 @@ describe('overviewController', () => {
   }
   const educationAndWorkPlanService = {
     getActionPlan: jest.fn(),
+    getTimeline: jest.fn(),
   }
   const ciagInductionService = {
     getWorkAndInterests: jest.fn(),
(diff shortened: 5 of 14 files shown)


server/@types/educationAndWorkPlanApi/index.d.ts
```diff
@@ -242,6 +244,12 @@ export interface components {
        * @example 2023-06-19T09:39:44Z
        */
       timestamp: string
+      /**
+       * Format: uuid
+       * @description A correlationId for this and any other TimelineEvents that occurred at the same time (i.e. within same atomic action). For example, this could be an update to a Goal, plus one or more of its child Steps. If required, this allows a client to easily group the events together.
```
chrimesdev, line 249: Re-imported the Swagger API spec for the Timeline

server/@types/viewModels/index.d.ts
```diff
@@ -361,6 +361,36 @@ declare module 'viewModels' {
     level: 'ENTRY_LEVEL' | 'LEVEL_1' | 'LEVEL_2' | 'LEVEL_3' | 'LEVEL_4' | 'LEVEL_5' | 'LEVEL_6' | 'LEVEL_7' | 'LEVEL_8'
   }
 
+  export interface Timeline {
+    reference?: string // TODO - Are these required/optional?
```
chrimesdev, line 365: Need to have a look at this, when attempting to pass the 

```
const timeline = await this.educationAndWorkPlanService.getTimeline(prisonNumber, req.user.token)
```

into the `TimelineView` in the `overviewController` 

`const view = new TimelineView(prisonerSummary, timeline)`

It said that the `reference and `prisonNumber` were missing 🤔

server/routes/overview/overviewController.test.ts
```diff
@@ -16,6 +16,7 @@ import CiagInductionService from '../../services/ciagInductionService'
 import aValidLongQuestionSetWorkAndInterests from '../../testsupport/workAndInterestsTestDataBuilder'
 import aValidPrisonerSummary from '../../testsupport/prisonerSummaryTestDataBuilder'
 import { aValidShortQuestionSetEducationAndTraining } from '../../testsupport/educationAndTrainingTestDataBuilder'
+import aValidTimelineResponse from '../../testsupport/timelineTestDataBuilder'
```
chrimesdev, line 19: Didn't know if `aValidTimelineResponse` was the best approach for this

server/filters/formatTimelineEvent.ts
```diff
@@ -0,0 +1,19 @@
+export default function formatTimelineEvent(value: string): string {
+  const timelineEventValue = TimelineEventValue[value as keyof typeof TimelineEventValue]
+  return timelineEventValue
+}
+
+enum TimelineEventValue {
```
chrimesdev, line 6: These `ENUM` values will need updating according to the content for the Timeline but this converts ENUM values into human readable text in the meantime

server/views/pages/overview/partials/timelineTab/timelineTabContents.njk
```diff
@@ -3,25 +3,26 @@
...
+          <div class="moj-timeline__header">
+            <h2 class="moj-timeline__title">{{ event.eventType | formatTimelineEvent }}</h2>
+            <p class="moj-timeline__byline govuk-!-display-none-print">by {{ event.actionedByDisplayName }}, {{ event.prisonId }}</p>
+          </div>
+
+          <p class="moj-timeline__date">
+            <time datetime="{{ event.timestamp }}">{{ event.timestamp | formatDate('D MMMM YYYY') }}</time>
+          </p>
+
+          {# TODO: Display the description and action
```
chrimesdev, line 17: We need to figure out the logic for when to display a "description" with an "action" (eg. Link, Tag) etc

server/views/pages/overview/partials/timelineTab/timelineTabContents.njk
```diff
@@ -3,25 +3,26 @@
 
     <div class="moj-timeline">
 
-      <!-- Timeline item -->
+      {% for event in timeline.events %}
+        <div class="moj-timeline__item">
+          <div class="moj-timeline__header">
+            <h2 class="moj-timeline__title">{{ event.eventType | formatTimelineEvent }}</h2>
+            <p class="moj-timeline__byline govuk-!-display-none-print">by {{ event.actionedByDisplayName }}, {{ event.prisonId }}</p>
```
chrimesdev, line 10: For now we're using `prisonId` as we don't have a method for retrieving the prison name 

Note: Hidden on Timeline as per requirements of the Print ticket

server/views/pages/overview/partials/timelineTab/timelineTabContents.njk
```diff
@@ -3,25 +3,26 @@
...
+          </div>
+
+          <p class="moj-timeline__date">
+            <time datetime="{{ event.timestamp }}">{{ event.timestamp | formatDate('D MMMM YYYY') }}</time>
```
chrimesdev, line 14: `datetime="{{ event.timestamp }}"` attribute is a pattern the MoJ timeline component uses

server/@types/viewModels/index.d.ts
```diff
@@ -361,6 +361,36 @@ declare module 'viewModels' {
     level: 'ENTRY_LEVEL' | 'LEVEL_1' | 'LEVEL_2' | 'LEVEL_3' | 'LEVEL_4' | 'LEVEL_5' | 'LEVEL_6' | 'LEVEL_7' | 'LEVEL_8'
   }
 
+  export interface Timeline {
```
chrimesdev, line 364: I wonder if `problemRetrievingData` needs adding to the Timeline API so we can handle the error on the UI?

server/views/pages/overview/partials/timelineTab/timelineTabContents.njk
```diff
@@ -3,25 +3,26 @@
 
     <div class="moj-timeline">
```
chrimesdev, line 4: Need to add `problemRetrievingData` error handling but currently not in the API

chrimesdev, line 4: Need to handle empty states

server/routes/overview/overviewController.test.ts
```diff
@@ -25,6 +26,7 @@ describe('overviewController', () => {
   }
```
nathanrussell-moj-digital, line 26: Can we revert the changes to `overviewController.test.ts` and `overviewController.ts` please - I don't think we're quite ready for the controller to call the service yet as we've not fully planned/designed where the filtering/mapping/grouping etc will happen. This really is quite a complex piece of this overall feature, so we need to step back from it and make sure we have designed and understood it.

nathanrussell-moj-digital: (chances are the test and implementation will look like you've done them, but equally they might not, so safest at this stage to take it in small chunks and do each part iteratively. We can always get your test etc from git history as and when we get there, but I dont think this story should be doing it yet)

server/routes/overview/timelineView.ts
```diff
@@ -1,15 +1,17 @@
-import type { PrisonerSummary } from 'viewModels'
+import type { PrisonerSummary, Timeline } from 'viewModels'
```
nathanrussell-moj-digital, line 1: Likewise, please can you revert this until we are ready 👍

server/services/educationAndWorkPlanService.ts
```diff
@@ -31,4 +31,8 @@ export default class EducationAndWorkPlanService {
     const updateGoalRequest = toUpdateGoalRequest(updateGoalDto)
     return this.educationAndWorkPlanClient.updateGoal(prisonNumber, updateGoalRequest, token)
   }
+
+  async getTimeline(prisonNumber: string, token: string): Promise<Timeline> {
```
nathanrussell-moj-digital, line 35: I think this might need some thought ....
I was envisaging that this method (`getTimeline`) would possibly be in its own service class `TimelineService`, where we have a service class designed around business domains/terminology, rather than where it gets the data from.
(the elephant in the room is that `EducationAndWorkPlanService` is clearly the wrong name, and should be split out to something like `GoalService` and `ActionPlanService` .... but that is one for another day and not this PR)

The other thing, not withstanding _where_ this method lives, is that its wrong and might fail at runtime! (or maybe not because there are no "types" in the resultant javascript 🤔 )
Anyway, its wrong because the method signature says it will return a promise of `Timeline`, where `Timeline` is your new view model.
But the `educationAndWorkPlanClient.getTimeline` method returns a promise of `TimelineResponse` , which is the API type.
If you look at all our other service methods like this, this is where we use a mapper. You'll need to write a mapper class that maps from a `TimelineResponse` and `TimelineEventResponse` into a `Timeline` and `TimelineEvent` (even if the fields are a one to one mapping, you still need a mapper)

server/testsupport/timelineTestDataBuilder.ts
```diff
@@ -0,0 +1,54 @@
+import type { Timeline } from 'viewModels'
+
+export default function aValidTimelineResponse(): Timeline {
```
nathanrussell-moj-digital, line 3: I think this might need some thought .... is it returning a `TimelineResponse`? (the API type) or a `Timeline`? (the UI view model)
Personally I think we want both, so I'd be tempted to rename the method to `aValidTimeline` (everything else looks OK I think); and then create a 2nd test builder file called `timelineResponseTestDataBuilder.ts` with a method `aValidTimelineResponse` that returns a `TimelineResponse`

server/@types/viewModels/index.d.ts
nathanrussell-moj-digital: Yes, we will certainly want that field in this view model class 👍 
I think you'll also want to make `reference` optional as well, because if there is a problem retrieving the data you wont have a reference number to populate the field with in the mapper

server/@types/viewModels/index.d.ts
```diff
@@ -361,6 +361,36 @@ declare module 'viewModels' {
...
+      | 'ACTION_PLAN_CREATED'
+      | 'GOAL_CREATED'
+      | 'GOAL_UPDATED'
+      | 'GOAL_STARTED'
+      | 'GOAL_COMPLETED'
+      | 'GOAL_ARCHIVED'
+      | 'STEP_UPDATED'
+      | 'STEP_NOT_STARTED'
+      | 'STEP_STARTED'
+      | 'STEP_COMPLETED'
```
mattwills-moj-digital, line 383: As things stand, the business don't want to see changes to individual steps on the Timeline (it's too detailed), so it might be best to remove these.

server/services/educationAndWorkPlanService.ts
mattwills-moj-digital: The thing to bear in mind here is that this mapper may not be straightforward, depending on where we apply the logic to map and filter between the API model (which is granular and includes updates to steps) and the view model (which presents a simplified view of these events, grouping some of them together).

I mentioned yesterday that it's probably worth the 3 of us chatting about this mapping/filtering, as there are some complexities/nuances here (e.g. concerning the `correlationId` that you've imported as part of the updated swagger spec).

mattwills-moj-digital: Wrt the `EducationAndWorkPlanService`, I agree that's not the best name and is something we should address (in a separate PR). I like the idea of naming services in terms of the business domain (so `TimelineService` would fit), rather than based on services (or "backend systems") which can change. The CIAG API is a prime example of this, though fortunately `CiagInductionService` works because CIAG is the business term for the people carrying out prisoner Inductions, and not just the name of the team who wrote the API! 😄

mattwills-moj-digital: You may then reasonably ask whether this means that the client class should be called `TimelineClient` or `EducationAndWorkPlanClient`? I think there's a key difference here. The client classes abstract away the details of the APIs they're calling. They know all about the API, including it's URL and data model and I think it makes sense to have one client per API (and one set of configuration for each one). 

In contrast, the service classes are more interested in the "business" domain, which in theory doesn't change if the underlying API changes (e.g. we move Inductions from the CIAG API to our API: Inductions are still the same). However, I realise this argument is flawed because our client classes are returning the API model objects to the service classes, without doing any mapping, so we can't re-write an API without it impacting the service classes. I'm not sure if we've got this right, but either way the intention is different. Hope that makes sense!

server/@types/viewModels/index.d.ts
nathanrussell-moj-digital: @chrimesdev - I've just re-read your question here, and seen the subtlety ! 😁 
You wrote:

> I wonder if `problemRetrievingData` needs adding to the Timeline API ....

Ah, I'd missed the phrase "Timeline API" when I first read it ...

OK, so, no, `problemRetrievingData` does not need adding to the Timeline API .... but it does need adding to this view model class. Have a look at the other view models in this source file - you'll find all of the "top level" view models that are built from API calls have this property. And yes, I think we need that property in this model type as well 👍

server/data/educationAndWorkPlanClient.test.ts
```diff
@@ -207,4 +208,46 @@ describe('educationAndWorkPlanClient', () => {
...
     })
   })
+
+  describe('getTimeline', () => {
+    it('should get Timeline', async () => {
+      // Given
+      const prisonNumber = 'A1234BC'
+      const systemToken = 'a-system-token'
+
+      const expectedTimelineResponse = aValidTimeline()
```
nathanrussell-moj-digital, line 218: This should be:
```
const expectedTimelineResponse = aValidTimelineResponse()
```
which means you'll need to write a test data builder for a `TimelineResponse` (keep the test builder you already have for a `Timeline` - we'll use that later. But _this_ test needs a `TimelineResponse`)

server/@types/viewModels/index.d.ts
chrimesdev: Added this 👍🏻

server/routes/overview/mappers/timelineMapper.test.ts
```diff
@@ -0,0 +1,83 @@
+import type { TimelineResponse, TimelineEventResponse } from 'educationAndWorkPlanApiClient'
+import type { Timeline, TimelineEvent } from 'viewModels'
+import { toTimeline, toTimelineEvent } from './timelineMapper'
+
+describe('timelineMapper', () => {
+  it('should map to Timeline', () => {
+    // Given
+    const timeline: TimelineResponse = {
+      problemRetrievingData: false,
```
nathanrussell-moj-digital, line 9: This is not _quite_ right. A `TimelineResponse` object is the API object, straight out of the REST API. It does not have a `problemRetrievingData` field. It's only our view models that have this field, not the REST API types.
Just nuke this line (line 9) and it'll be OK 👍 

(but I'm about to make the same comment on the test data builder as well)

server/testsupport/timelineResponseTestDataBuilder.ts
```diff
@@ -0,0 +1,55 @@
+import type { TimelineResponse } from 'educationAndWorkPlanApiClient'
+
+export default function aValidTimelineResponse(): TimelineResponse {
+  return {
+    problemRetrievingData: false,
```
nathanrussell-moj-digital, line 5: This is not _quite_ right. A `TimelineResponse` object is the API object, straight out of the REST API. It does not have a `problemRetrievingData` field. It's only our view models that have this field, not the REST API types.
Just nuke this line (line 5) and it'll be OK 👍

server/testsupport/timelineResponseTestDataBuilder.ts
```diff
@@ -0,0 +1,55 @@
+import type { TimelineResponse } from 'educationAndWorkPlanApiClient'
```
nathanrussell-moj-digital, line 1: This is the test data builder for the REST API type, and with the exception of my comment below is fine 👍 

Do we also have a test data builder for the view model type `Timeline` ?
We will need one, so you'll need to create a new test data builder file `timelineTestDataBuilder.ts` that builds a valid `Timeline` with a function `aValidTimeline`

Keep this test data builder because it's fine (after deleting `problemRetrievingData`); and add a new file `timelineTestDataBuilder.ts` that builds a valid `Timeline` with a function `aValidTimeline` 👍

server/routes/overview/mappers/timelineMapper.test.ts
chrimesdev: Good spot I've resolved this

server/testsupport/timelineResponseTestDataBuilder.ts
chrimesdev: I've added a new `timelineTestDataBuilder.ts` data is pretty much the same as the response (for now) but with the additional `problemRetrievingData`

nathanrussell-moj-digital (approved):
Looks good; lets get this merged 👍 
Thanks for making the changes, and sorry it's probably been a bit of a drag! 😬

PR #335, opened 2023-11-16
RR-490 - Catch errors when calling APIs for prisoner-list page and redirect to error page

nathanrussell-moj-digital:
This PR handles errors from the 3 APIs that the Prisoner List page calls (`prisoner-search`, `ciag-api`, `plp-api`) and displays the generic error screen
(The data from all 3 APIs is required to be able to display the Prisoner List page, so if any of the 3 fail there is nothing that the user can do at that point. Its pretty much an end game and all we can do is show the error page)

### DIFF
--- a/integration_tests/e2e/overview/preInductionOverview.cy.ts
+++ b/integration_tests/e2e/overview/preInductionOverview.cy.ts
@@ -1,6 +1,6 @@
 import Page from '../../pages/page'
 import OverviewPage from '../../pages/overview/OverviewPage'
-import CreateCiagInductionPage from '../ciagUi/createCiagInductionPage'
+import CreateCiagInductionPage from '../../pages/ciagUi/createCiagInductionPage'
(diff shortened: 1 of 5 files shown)


integration_tests/e2e/overview/preInductionOverview.cy.ts
```diff
@@ -1,6 +1,6 @@
 import Page from '../../pages/page'
 import OverviewPage from '../../pages/overview/OverviewPage'
-import CreateCiagInductionPage from '../ciagUi/createCiagInductionPage'
+import CreateCiagInductionPage from '../../pages/ciagUi/createCiagInductionPage'
```
nathanrussell-moj-digital, line 3: Not strictly related to this PR, but this `Page` class (`CreateCiagInductionPage` was in the wrong directory in the source code structure, so I've moved it)

chrimesdev (approved):

PR #336, opened 2023-11-20
RR-480 - Import swagger spec for prison-register-api

nathanrussell-moj-digital:
Import the swagger spec for `prison-register-api` so that we can call the API to lookup prison names from their IDs (not this PR)


chrimesdev (approved):

mattwills-moj-digital (approved):
Beautiful :)

PR #340, opened 2023-11-21
RR-480 - Added prisonRegisterClient

nathanrussell-moj-digital:
This PR adds a `PrisonRegisterClient` in order to make REST API calls to lookup prison data
Also included are the wiremock stubs for this API call for as and when we need them 👍 

The next PR will be the service class to lookup and cache a prison to return the prison name, which will use the `PrisonRegisterClient` in this PR plus the `PrisonRegisterStore` from my last PR


mattwills-moj-digital (approved):
A fine piece of work

PR #342, opened 2023-11-21
RR-474 display learning and work progress plan updated events on timeline

chrimesdev:
## Description

Display "Learning and work progress plan updated" events on the Timeline UI which is the `INDUCTION_UPDATED` event in the API.

https://dsdmoj.atlassian.net/browse/RR-474

screencapture-localhost-3000-plan-A9412DY-view-timeline-2023-11-21-14_04_42

### DIFF
--- a/server/services/timelineService.ts
+++ b/server/services/timelineService.ts
@@ -9,7 +9,7 @@ export default class TimelineService {
   async getTimeline(prisonNumber: string, token: string): Promise<TimelineResponse> {
     try {
       const timelineResponse = await this.educationAndWorkPlanClient.getTimeline(prisonNumber, token)
-      const timelineResponseFilteredEvents = this.filterTimelineByActionPlanCreatedEvent(timelineResponse)
+      const timelineResponseFilteredEvents = this.filterTimelineByActionPlanCreatedOrUpdatedEvent(timelineResponse)
       timelineResponse.events = timelineResponseFilteredEvents
       return toTimeline(timelineResponse)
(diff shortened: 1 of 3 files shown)


server/services/timelineService.ts
```diff
@@ -18,7 +18,10 @@ export default class TimelineService {
     }
   }
 
-  filterTimelineByActionPlanCreatedEvent = (timelineResponse: TimelineResponse) => {
-    return timelineResponse.events.filter((event: { eventType: string }) => event.eventType === 'ACTION_PLAN_CREATED')
+  filterTimelineByActionPlanCreatedOrUpdatedEvent = (timelineResponse: TimelineResponse) => {
```
chrimesdev, line 21: This could possibly be 2 functions, but felt they were under the same category of "Action plan" events so grouped them together.

nathanrussell-moj-digital: Happy for it to be one function 👍 , it's how I envisaged it in these early commits.
Though I think this service class might change pretty soon once I wire in the prison name lookup! 😁

nathanrussell-moj-digital: One thing I did note is that we dont have a unit test for this class 😢 
Dont worry for now, because as mentioned I'm about to change this class quite a lot, so I'll create the skeleton unit class at that time; but I might ask you to flesh out some of the tests

nathanrussell-moj-digital (approved):
LGTM 👍

PR #343, opened 2023-11-21
RR-480 - Refactor Prison to PrisonResponse; added Prison view model and added to TimelineEvent

nathanrussell-moj-digital:
This PR:

* renames the API response type `Prison` to `PrisonResponse`
* Adds a new view model type `Prison`
* Adds a mapper to map from `PrisonResponse` to `Prison`
* Adds a `prison` field of type `Prison` to `TimelineEvent`
* Updates the `TimelineMapper` (and moves to `data/mappers`)
* Updates test data builders etc


server/data/mappers/timelineMapper.ts
```diff
@@ -15,7 +15,10 @@ const toTimelineEvent = (timelineEventResponse: TimelineEventResponse): Timeline
     reference: timelineEventResponse.reference,
     sourceReference: timelineEventResponse.sourceReference,
     eventType: timelineEventResponse.eventType,
-    prisonName: timelineEventResponse.prisonId, // (JIRA: 480) TODO: Lookup Prison name from the service
+    prison: {
+      prisonId: timelineEventResponse.prisonId,
+      prisonName: undefined, // This mapper cannot look up / resolve the prisonName; that has to be post-processed
```
nathanrussell-moj-digital, line 20: All the change in this PR is essentially related to this line!
We can't lookup the prison name within the mapper. ie. we have service classes that use/depend on mappers - we don't want a mapper that uses/depends on a service - it the wrong way round.
So, the mapper will leave the `prison.prisonName` field undefined, and the `TimelineService` will post-process it to update this field (not this PR)

chrimesdev (approved):
LGTM

PR #345, opened 2023-11-22
RR-480 - Wired PrisonService into timelineService to lookup prison names for each TimelineEvent

nathanrussell-moj-digital:
This PR completes the prison name lookup, populating the prison name into each `TimelineEvent`

Screenshot 2023-11-22 at 10 28 14

### DIFF
--- a/server/data/prisonRegisterClient.ts
+++ b/server/data/prisonRegisterClient.ts
@@ -13,9 +13,10 @@ export default class PrisonRegisterClient {
     })
   }
 
-  async getAllPrisons(token: string): Promise<PrisonResponse> {
-    return PrisonRegisterClient.restClient(token).get({
+  async getAllPrisons(token: string): Promise<Array<PrisonResponse>> {
+    const allPrisonResponses = PrisonRegisterClient.restClient(token).get({
       path: '/prisons',
     })
+    return allPrisonResponses as Promise<Array<PrisonResponse>>
   }
 }
--- a/server/routes/overview/overviewController.ts
+++ b/server/routes/overview/overviewController.ts
@@ -128,7 +128,7 @@ export default class OverviewController {
     const { prisonNumber } = req.params
     const { prisonerSummary } = req.session
 
-    const timeline = await this.timelineService.getTimeline(prisonNumber, req.user.token)
+    const timeline = await this.timelineService.getTimeline(prisonNumber, req.user.token, req.user.username)
     const view = new TimelineView(prisonerSummary, timeline)
     res.render('pages/overview/index', { ...view.renderArgs })
   }
--- a/server/services/prisonService.test.ts
+++ b/server/services/prisonService.test.ts
@@ -1,0 +1,211 @@
+import type { PrisonResponse } from 'prisonRegisterApiClient'
+import toPrison from '../data/mappers/prisonMapper'
+import PrisonService from './prisonService'
+import PrisonRegisterStore from '../data/cache/prisonRegisterStore'
+import PrisonRegisterClient from '../data/prisonRegisterClient'
+import aValidPrisonResponse from '../testsupport/prisonResponseTestDataBuilder'
+import aValidPrison from '../testsupport/prisonTestDataBuilder'
+
+jest.mock('../data/mappers/prisonMapper')
+
+describe('prisonService', () => {
+  const mockedPrisonMapper = toPrison as jest.MockedFunction<typeof toPrison>
+
+  const prisonRegisterStore = {
+    getActivePrisons: jest.fn(),
+    setActivePrisons: jest.fn(),
+  }
+
+  const prisonRegisterClient = {
+    getAllPrisons: jest.fn(),
+  }
+
+  const prisonService = new PrisonService(
+    prisonRegisterStore as unknown as PrisonRegisterStore,
+    prisonRegisterClient as unknown as PrisonRegisterClient,
+  )
+
+  beforeEach(() => {
+    jest.resetAllMocks()
+  })
+
+  const allPrisons: Array<PrisonResponse> = [
+    aValidPrisonResponse({
+      prisonId: 'ACI',
+      prisonName: 'Altcourse (HMP)',
+      active: false,
+    }),
+    aValidPrisonResponse({
+      prisonId: 'ASI',
+      prisonName: 'Ashfield (HMP)',
+      active: true,
+    }),
+    aValidPrisonResponse({
+      prisonId: 'MDI',
+      prisonName: 'Moorland (HMP & YOI)',
+      active: true,
+    }),
+  ]
+
+  const activePrisons: Array<PrisonResponse> = [
+    aValidPrisonResponse({
+      prisonId: 'ASI',
+      prisonName: 'Ashfield (HMP)',
+      active: true,
+    }),
+    aValidPrisonResponse({
+      prisonId: 'MDI',
+      prisonName: 'Moorland (HMP & YOI)',
+      active: true,
+    }),
+  ]
+
+  describe('getPrisonByPrisonId', () => {
+    it('should get prison by ID given prison has been previously cached', async () => {
+      // Given
+      const prisonId = 'MDI'
+      const systemToken = 'a-system-token'
+
+      prisonRegisterStore.getActivePrisons.mockResolvedValue(activePrisons)
+
+      const moorlandPrisonResponse = aValidPrisonResponse({
+        prisonId: 'MDI',
+        prisonName: 'Moorland (HMP & YOI)',
+        active: true,
+      })
+
+      const expectedPrison = aValidPrison({
+        prisonId: 'MDI',
+      })
+      mockedPrisonMapper.mockReturnValue(expectedPrison)
+
+      // When
+      const actual = await prisonService.getPrisonByPrisonId(prisonId, systemToken)
+
+      // Then
+      expect(actual).toEqual(expectedPrison)
+      expect(prisonRegisterStore.getActivePrisons).toHaveBeenCalled()
+      expect(mockedPrisonMapper).toHaveBeenCalledWith(moorlandPrisonResponse)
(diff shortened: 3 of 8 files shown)


server/data/prisonRegisterClient.ts
```diff
@@ -13,9 +13,10 @@ export default class PrisonRegisterClient {
     })
   }
 
-  async getAllPrisons(token: string): Promise<PrisonResponse> {
-    return PrisonRegisterClient.restClient(token).get({
+  async getAllPrisons(token: string): Promise<Array<PrisonResponse>> {
```
nathanrussell-moj-digital, line 16: Whoops! I used the wrong method signature when I first wrote this! It only came out of the woodwork now we are using it! 😁

server/routes/overview/overviewController.ts
```diff
@@ -128,7 +128,7 @@ export default class OverviewController {
     const { prisonNumber } = req.params
     const { prisonerSummary } = req.session
 
-    const timeline = await this.timelineService.getTimeline(prisonNumber, req.user.token)
+    const timeline = await this.timelineService.getTimeline(prisonNumber, req.user.token, req.user.username)
```
nathanrussell-moj-digital, line 131: We now also need to pass the username into the `TimelineService` method (see the change to `TimelineService` to see why)

server/services/timelineService.test.ts
```diff
@@ -0,0 +1,212 @@
...
+    jest.resetAllMocks()
+  })
+
+  const prisonNumber = 'A1234BC'
+  const username = 'a-dps-user'
+  const userToken = 'a-user-token'
+  const systemToken = 'a-system-token'
+
+  describe('getTimeline', () => {
+    describe('lookup prison names', () => {
```
nathanrussell-moj-digital, line 41: This test class didnt exist previously. I've started it by writing tests specifically for the prison name lookup stuff that it does, but I have not wriiten tests for the rest of the functionality in this class. We will need to do that 👍

server/services/timelineService.test.ts
```diff
@@ -0,0 +1,212 @@
...
+              timestamp: '2023-09-01T10:47:38.565Z',
+              correlationId: '246aa049-c5df-459d-8231-bdeab3936d0f',
+              contextualInfo: '',
+              actionedByDisplayName: 'Ralph Gen',
+            },
+          ],
+        }
+        educationAndWorkPlanClient.getTimeline.mockResolvedValue(timelineResponse)
+
+        const timeline: Timeline = {
```
nathanrussell-moj-digital, line 161: I know we have the test data builders for `TimelineEventResponse` and `TimelineEvent` 👍 ,  but they are not flexible enough at the moment for what we need to do in these tests (ie. have events of certain types/states). At some point in time we can possibly improve those builders to make them more flexible in this respect, at which point this test class can become a lot tidier 👍

server/services/timelineService.ts
```diff
@@ -1,28 +1,83 @@
...
+  constructor(
+    private readonly hmppsAuthClient: HmppsAuthClient,
+    private readonly educationAndWorkPlanClient: EducationAndWorkPlanClient,
+    private readonly prisonService: PrisonService,
+  ) {}
 
-  async getTimeline(prisonNumber: string, token: string): Promise<Timeline> {
+  async getTimeline(prisonNumber: string, token: string, username: string): Promise<Timeline> {
     try {
+      const systemToken = await this.hmppsAuthClient.getSystemClientToken(username)
```
nathanrussell-moj-digital, line 20: OK, this is why we need the username - we use it in order to get a system token. We need the system token to pass into the `PrisonService` so that it can retrieve prisons from the `prison-register-api` if it needs to

server/services/timelineService.ts
```diff
@@ -1,28 +1,83 @@
...
+    try {
+      return (await this.prisonService.getPrisonByPrisonId(prisonId, token)) || { prisonId, prisonName: undefined }
+    } catch (e) {
+      logger.error(`Error looking up prison ${prisonId}`, e)
+      // return a Prison with just the prison ID set. Failing to lookup the prison should not stop the Timeline service returning a Timeline
+      return { prisonId, prisonName: undefined }
+    }
   }
+
+  private filterTimelineEvents = (event: TimelineEventResponse): boolean =>
```
nathanrussell-moj-digital, line 81: I've changed the way we filter the events - this approach will be easier to maintain moving forwards as we can simply add to the `SUPPORTED_TIMELINE_EVENTS` array at the top of the file 👍

server/services/timelineService.ts
```diff
@@ -1,28 +1,83 @@
...
       return { problemRetrievingData: true } as Timeline
     }
   }
 
-  filterTimelineByActionPlanCreatedOrUpdatedEvent = (timelineResponse: TimelineResponse) => {
-    return timelineResponse.events.filter(
-      (event: { eventType: string }) =>
-        event.eventType === 'ACTION_PLAN_CREATED' || event.eventType === 'INDUCTION_UPDATED',
-    )
+  private addPrisonNameToPrisons = async (
```
nathanrussell-moj-digital, line 33: OK; strap yourself in, this function is complex!
The purpose and intent of the function is to iterate through the `TimelineEvent`s and for each one lookup the prison name from it's prisonId via the `PrisonService`. It will mutate each `TimelineEvent` by setting the prison name, and will then return the mutated collection.

That's what it does .... but there is complexity! mainly because the functions in `PrisonService` are `async`, which means you need to call them with `await` and return a `Promise`.

My original code for this function was this:
```
  private addPrisonNameToPrisons = async (
    events: Array<TimelineEvent>,
    token: string,
  ): Promise<Array<TimelineEvent>> => {
    return Promise.all(
      events.map(async (event): Promise<TimelineEvent> => {
        return {
          ...event,
          prison: await this.lookupPrison(event.prison.prisonId, token),
        }
      }),
    )
  }
```

It returns a `Promise` containing all (an array) of the mapped `TimelineEvent`s. In theory and on paper it works .... however because the promise and the call to the `PrisonService` (via the `lookupPrison` function) are async the mapping of each `TmelineEvent` and the call to `PrisonService` effectively run nanoseconds apart. This in turn means we end up with a race condition where several `TimelineEvent` mappings call the `PrisonService` at "the same time".
The code in the `PrisonService` method effectively says "if its in the cache return if from the cache, else call the prison-register-api to populate the cache, then return it from the cache". Given that the "check and set cache" is not an atomic operation you can end up with a race condition, which has the effect of spamming the `prison-register-api` !!! 😬 

The only way I could get round it is to do what I've done here, and I hope the comments help 🤞

server/services/prisonService.test.ts
```diff
@@ -0,0 +1,211 @@
...
+      mockedPrisonMapper.mockReturnValue(expectedPrison)
+
+      // When
+      const actual = await prisonService.getPrisonByPrisonId(prisonId, systemToken)
+
+      // Then
+      expect(actual).toEqual(expectedPrison)
+      expect(prisonRegisterStore.getActivePrisons).toHaveBeenCalled()
+      expect(mockedPrisonMapper).toHaveBeenCalledWith(moorlandPrisonResponse)
+      expect(prisonRegisterClient.getAllPrisons).not.toHaveBeenCalled()
```
mattwills-moj-digital, line 89: Nice 👍

server/services/prisonService.test.ts
```diff
@@ -0,0 +1,211 @@
...
+
+      // Then
+      expect(actual).toEqual(expectedPrison)
+      expect(prisonRegisterStore.getActivePrisons).toHaveBeenCalled()
+      expect(mockedPrisonMapper).toHaveBeenCalledWith(moorlandPrisonResponse)
+      expect(prisonRegisterClient.getAllPrisons).toHaveBeenCalled()
+      expect(prisonRegisterStore.setActivePrisons).toHaveBeenCalledWith(allPrisons, 1)
+    })
+
+    it('should not get prison by ID given retrieving from cache and API both throw errors', async () => {
```
mattwills-moj-digital, line 172: Crikey, I hope this never happens!

server/services/prisonService.ts
```diff
@@ -0,0 +1,60 @@
...
+      return toPrison(prisonResponse)
+    }
+
+    logger.info(`Could not find details for prison ${prisonId}`)
+    return undefined
+  }
+
+  private async getPrison(prisonId: string, token: string): Promise<PrisonResponse | undefined> {
+    return (
+      (await this.getCachedPrison(prisonId)) ||
```
mattwills-moj-digital, line 29: I missed the `||` initially. I may have been tempted to use an `if` statement, but this is fine 👍

server/services/prisonService.test.ts
```diff
@@ -0,0 +1,211 @@
...
+      active: true,
+    }),
+    aValidPrisonResponse({
+      prisonId: 'MDI',
+      prisonName: 'Moorland (HMP & YOI)',
+      active: true,
+    }),
+  ]
+
+  const activePrisons: Array<PrisonResponse> = [
```
mattwills-moj-digital, line 50: I'm a bit confused between the terminology of `active` and `cached`. Are they not the same thing?

mattwills-moj-digital (approved):
LGTM. Just a couple of minor comments.

server/services/prisonService.test.ts
nathanrussell-moj-digital: 💯 , agreed; but at least we have it covered 😁 👍

server/services/prisonService.test.ts
nathanrussell-moj-digital: No, defo not the same thing.
"cached" is a tech term, and is what you imagine it is (ie. an object in the Redis cache and therefore we dont need to look it up from it's source REST API)

"active" in this sense is a HMPPS term. Prisons have a field `active` which is a boolean and determines whether the prison is active of not. IE. some prisons are non-active for some reason (perhaps they have been closed). These prisons come back from the prison-register-api REST API call, but are not in use and have no prisoners in them.
These prisons we dont care about, so we dont cache them.

PR #348, opened 2023-11-27
RR-475 Display goal(s) created on the Timeline UI

chrimesdev:
## Description 

Display goal(s) created on the Timeline UI and display them with the most recent event first on the UI.

Removed the eventTypes that we don't use in the UI from the the `TimelineEvent` ViewModel and the tests/test data.

Changed the Timeline `timestamp` ViewModel to be a `Date` for the `dateOperator` function and well just because it probably should be a date!

### Screenshot

For the "Goal created" on 16 October 2023 and 8 November 2023 there was only 1 goal created in the single form flow - so "Goal created" is displayed

For the "Goals created" on the 27th November 2023 there was 2 goals created in the single form flow - so "Goal**s** created" is displayed

screencapture-localhost-3000-plan-A9412DY-view-timeline-2023-11-28-16_09_59


server/services/timelineService.ts
```diff
@@ -21,6 +21,29 @@ export default class TimelineService {
       const timelineResponse = await this.educationAndWorkPlanClient.getTimeline(prisonNumber, token)
       timelineResponse.events = timelineResponse.events.filter(this.filterTimelineEvents)
 
+      // Check for GOAL_CREATED events with the same correlationId, then change the event type to MULTIPLE_GOALS_CREATED
```
nathanrussell-moj-digital, line 24: It's cool that it's working 👍  Nice work 👍 

though I am struggling to read and understand this new block of code that does the magic 😁  I also notice this new code is not covered by any new tests

I wonder if we have some options here .... maybe rework this code so it's easier to follow? (and include tests for it in the main timelineService.test.ts); or factor it out into some external function that can be unit tested independently (and mocked in timelineService.test.ts)
Or even consider whether this is even the right place to do this ... 🤔  .... I'm wondering if this grouping is more of a UI concern than this main service and mapping, and perhaps it would be better placed somewhere closer to the view? Perhaps in the controller method, or maybe in the View object?
I don't know the answer to the above, but I'm not sure what we have here is quite right yet ... 🤔 

What are you thoughts about it?

nathanrussell-moj-digital (commented):
Loving everything in this PR; but not 100% sure on the stuff in timelineService that does the actual magic 🤔

server/@types/viewModels/index.d.ts
```diff
@@ -371,17 +371,9 @@ declare module 'viewModels' {
...
-      | 'GOAL_UPDATED'
-      | 'GOAL_STARTED'
-      | 'GOAL_COMPLETED'
-      | 'GOAL_ARCHIVED'
-      | 'INDUCTION_UPDATED'
-      | 'INDUCTION_CREATED'
+    eventType: 'ACTION_PLAN_CREATED' | 'GOAL_CREATED' | 'MULTIPLE_GOALS_CREATED' | 'GOAL_UPDATED' | 'INDUCTION_UPDATED'
     prison: Prison
-    timestamp: string
+    timestamp: Date
```
nathanrussell-moj-digital, line 376: Ah ha, this would probably help re: the sorting! 😁 Good one 👍

server/data/mappers/timelineMapper.test.ts
```diff
@@ -14,7 +15,7 @@ describe('timelineMapper', () => {
           eventType: 'ACTION_PLAN_CREATED',
           prisonId: 'MDI',
           actionedBy: 'RALPH_GEN',
-          timestamp: '2023-09-01T10:47:38.565Z',
+          timestamp: moment('2023-09-01T10:47:38.565Z').toDate(),
```
nathanrussell-moj-digital, line 18: Ah; _this_ line needs to stay as being a string. Its the value from the API (ie. the Response object) and will always be a string). It's only "our" objects (so in this case the view model object) that we can use a richer type with)

server/data/mappers/timelineMapper.test.ts
```diff
@@ -34,7 +35,7 @@ describe('timelineMapper', () => {
             prisonId: 'MDI',
             prisonName: undefined,
           },
-          timestamp: '2023-09-01T10:47:38.565Z',
+          timestamp: moment('2023-09-01T10:47:38.565Z').toDate(),
```
nathanrussell-moj-digital, line 38: 👍

server/data/mappers/timelineMapper.test.ts
```diff
@@ -57,7 +58,7 @@ describe('timelineMapper', () => {
       eventType: 'ACTION_PLAN_CREATED',
       prisonId: 'MDI',
       actionedBy: 'RALPH_GEN',
-      timestamp: '2023-09-01T10:47:38.565Z',
+      timestamp: moment('2023-09-01T10:47:38.565Z').toDate(),
```
nathanrussell-moj-digital, line 61: Same here - this needs to stay as being a `string`, not a `Date`

server/data/mappers/timelineMapper.test.ts
```diff
@@ -71,7 +72,7 @@ describe('timelineMapper', () => {
         prisonId: 'MDI',
         prisonName: undefined,
       },
-      timestamp: '2023-09-01T10:47:38.565Z',
+      timestamp: moment('2023-09-01T10:47:38.565Z').toDate(),
```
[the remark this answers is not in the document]
nathanrussell-moj-digital: But you'll also need to update `timelineMapper.ts` to turn the string from the API Response object into a `Date` object via `moment`

server/routes/timelineResolver.ts
```diff
@@ -0,0 +1,42 @@
+/* eslint-disable no-param-reassign */
+import type { TimelineResponse, TimelineEventResponse } from 'educationAndWorkPlanApiClient'
+import type { TimelineEvent } from 'viewModels'
+import dateComparator from './dateComparator'
+
+const filterTimelineEvents = (timelineResponse: TimelineResponse): TimelineResponse => {
```
nathanrussell-moj-digital, line 6: Your function arg here should not be the API Response type, it needs to be the view model type (`TimeLine`) - it's what comes out of the service method that is being passed into this function 👍

nathanrussell-moj-digital: And therefore your test above needs to change to use Timeline rather than TimelineResponse, at which point you'll have a `timestamp` field that is a `Date` which should then sort as you expect it to

server/routes/overview/overviewController.ts
```diff
@@ -128,7 +129,8 @@ export default class OverviewController {
     const { prisonNumber } = req.params
     const { prisonerSummary } = req.session
 
-    const timeline = await this.timelineService.getTimeline(prisonNumber, req.user.token, req.user.username)
+    const allTimelineEvents = await this.timelineService.getTimeline(prisonNumber, req.user.token, req.user.username)
+    const timeline = filterTimelineEvents(allTimelineEvents)
```
nathanrussell-moj-digital, line 133: I think I'd like to see an update to the overview controller test to reflect the fact that we have this new function that fiilters timeline events. But seeing as the existing test still works lets leave it for now. Lets concentrate on getting the core functionality working first 👍 
We can add the test in a separate PR afterwards 👍

server/data/mappers/timelineMapper.ts
```diff
@@ -19,11 +20,15 @@ const toTimelineEvent = (timelineEventResponse: TimelineEventResponse): Timeline
       prisonId: timelineEventResponse.prisonId,
       prisonName: undefined, // This mapper cannot look up / resolve the prisonName; that has to be post-processed
     },
-    timestamp: timelineEventResponse.timestamp,
+    timestamp: toDate(timelineEventResponse.timestamp),
```
nathanrussell-moj-digital, line 23: 👍

nathanrussell-moj-digital (approved):
Yeah, this is looking good now; nice work 👍 
Lets get this merged 👍

PR #349, opened 2023-11-29
RR-514 - Added basic/initial code for new style routes for Create Goal screens.

nathanrussell-moj-digital:
This PR the initial/basic new routes for the Create Goal process.
Its all sat behind a feature toggle, and the current routes are still there and operational.

In this PR I have opted to just define the routes using a hardcoded goal/step index of `1` throughout. This is enough to get the skeleton code there, and to ensure the cypress tests continue to work (though I did need to change the paths they use)

My next PR will start fleshing out the functionality by adding logic to the controller to increment indexes etc as you add goals and steps.

On the subject of the controller, because I'm envisaging quite a lot of change in the controller methods to make use of the indexes in the paths, I've opted to copy and paste the controller, so now we have 2 `CreateGoalControllers` - 1 for the original (current production) behaviour, and 1 for the new behaviour. This means that the next set of changes I need to make in order to operate on the indexes will not need to be wrapped in messy conditional logic based on the feature toggle. This will make the controller implementation and tests much simpler 👍


server/routes/createGoal/createGoalController.originalBehaviour.ts
```diff
@@ -0,0 +1,162 @@
...
+import CreateGoalView from './createGoalView'
+import AddStepView from './addStepView'
+import AddNoteView from './addNoteView'
+import { toCreateGoalDto } from './mappers/createGoalFormToCreateGoalDtoMapper'
+import validateCreateGoalForm from './createGoalFormValidator'
+import validateAddStepForm from './addStepFormValidator'
+import ReviewView from './reviewView'
+import futureGoalTargetDateCalculator from '../futureGoalTargetDateCalculator'
+
+/**
```
nathanrussell-moj-digital, line 15: This new class is a verbatim copy and paste of the original `createGoalController`  but with this comment block and a different class name. Functionally it's identical to the original.

server/routes/createGoal/createGoalController.originalBehaviour.test.ts
```diff
@@ -0,0 +1,506 @@
...
+import { toCreateGoalDto } from './mappers/createGoalFormToCreateGoalDtoMapper'
+import aValidPrisonerSummary from '../../testsupport/prisonerSummaryTestDataBuilder'
+import futureGoalTargetDateCalculator from '../futureGoalTargetDateCalculator'
+import CreateGoalControllerOriginalBehaviour from './createGoalController.originalBehaviour'
+
+jest.mock('./addStepFormValidator')
+jest.mock('./createGoalFormValidator')
+jest.mock('./mappers/createGoalFormToCreateGoalDtoMapper')
+
+/**
```
nathanrussell-moj-digital, line 23: This new test class is a verbatim copy and paste of the original `createGoalController.test.ts` but with this comment block. Functionally it's identical to the original.

server/routes/createGoal/createGoalController.test.ts
```diff
@@ -20,6 +20,9 @@ jest.mock('./addStepFormValidator')
 jest.mock('./createGoalFormValidator')
 jest.mock('./mappers/createGoalFormToCreateGoalDtoMapper')
 
+/**
+ * Unit tests for createGoalController.
+ */
```
nathanrussell-moj-digital, line 25: This is the original unit test class for the original `createGoalController` - the changes here are simply around the route paths, because thats the scope of the change that I've made in the implementation.

server/routes/createGoal/index.ts
```diff
@@ -14,8 +16,44 @@ import {
  * Route definitions for the pages relating to Creating A Goal
  */
 export default (router: Router, services: Services) => {
-  const createGoalController = new CreateGoalController(services.educationAndWorkPlanService)
+  if (config.featureToggles.newCreateGoalRoutesEnabled) {
```
nathanrussell-moj-digital, line 19: Using the new feature toggle, we either setup routes with the new index based paths using the controller; or setup the existing routes using the original behaviour controller.

mattwills-moj-digital (approved):
LGTM

PR #352, opened 2023-11-29
RR-514 - Logic to use the goal and step indexes

nathanrussell-moj-digital:
This PR updates `createGoalController` to start making use of the goal and step indexes that are in the route paths


mattwills-moj-digital (approved):
Simply marvellous.

PR #361, opened 2023-12-06
RR-532 fix issue with momentjs daylight saving time

chrimesdev:
## Description 

Fix issue with `momentjs` daylight saving time dates

The fix is `moment` to `moment.utc`

https://dsdmoj.atlassian.net/browse/RR-532


nathanrussell-moj-digital (approved):
LGTM; nice one 👍

PR #369, opened 2023-12-15
RR-539 display the line on print timeline

chrimesdev:
## Description

Display the line on print, for the Timeline.

Because of how the Timeline component is built in the MoJ Design System, the Timeline "line" isn't show on print. This is because it uses `::before` pseudo-elements to style the line, which gets ignored by print, because its a pseudo-element.

https://dsdmoj.atlassian.net/browse/RR-539

Screenshot 2023-12-15 at 10 47 41

### DIFF
--- a/assets/scss/print.scss
+++ b/assets/scss/print.scss
@@ -41,3 +41,18 @@
 .connect-dps-common-footer {
   @include govuk-visually-hidden();
 }
+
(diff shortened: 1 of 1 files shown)


nathanrussell-moj-digital (approved):
LGTM 👍

PR #374, opened 2024-01-04
RR-544 - Add goals from pre-induction overview page

nathanrussell-moj-digital:
This PR implements RR-544 to add functionality to the pre-induction Overview screen to allow adding Goals before the CIAG Induction is created (and subsequently displaying those goals on the pre-induction Overview screen)

No changes have been made to the post-induction Overview screen of process flow.

https://github.com/ministryofjustice/hmpps-education-and-work-plan-ui/assets/94835226/e413d31b-51ee-4c23-adf5-86a4e88ab526


helm_deploy/values-dev.yaml
```diff
@@ -24,6 +24,7 @@ generic-service:
     DPS_URL: "https://digital-dev.prison.service.justice.gov.uk"
     FRONTEND_COMPONENT_API_URL: "https://frontend-components-dev.hmpps.service.justice.gov.uk"
     ENVIRONMENT_NAME: "DEV"
+    CREATE_GOALS_WITHOUT_INDUCTION_ENABLED: "true"
```
nathanrussell-moj-digital, line 27: Feature toggled; only enabled in dev at the moment

server/testsupport/actionPlanTestDataBuilder.ts
```diff
@@ -1,6 +1,34 @@
...
+ * ```
+ * // create an action plan with 3 goals
+ * const actionPlan = aValidActionPlanWithOneGoal(
+ *  {
+ *    goals: [aValidGoal(), aValidGoal(), aValidGoal()],
+ *  }
+ * )
+ * ```
+ */
 const aValidActionPlanWithOneGoal = (
```
nathanrussell-moj-digital, line 32: In this PR I decided to deprecate this method with a suitable comment rather than rename it, otherwise this PR would have a load of unrelated change where I'd renamed this method.
We can remove this deprecated method and change all the existing uses of it to `aValidActionPlan()` in a separate PR

server/views/pages/overview/partials/overviewTab/overviewTabPostInduction.njk
```diff
@@ -3,103 +3,8 @@
 <section data-qa="post-induction-overview">
   {% if not actionPlan.problemRetrievingData %}
 
-    <div class="moj-page-header-actions">
```
nathanrussell-moj-digital, line 6: I've pulled out the action button header and goals list into includes (above) because they will also now be used in the pre-induction overview page (below)

server/views/pages/overview/partials/overviewTab/overviewTabPreInduction.njk
```diff
@@ -1,16 +1,38 @@
...
+    {{ govukNotificationBanner({
+      html: createInductionBanner
+    }) }}
+
+  {% else %}
+    {% if not actionPlan.problemRetrievingData %}
+
+      {% include './_actionsHeader.njk' %}
+
+      {{ govukNotificationBanner({
```
nathanrussell-moj-digital, line 24: Put the banner between the action buttons header and the goals list, as per the design

chrimesdev (approved):
LGTM

server/testsupport/actionPlanTestDataBuilder.ts
mattwills-moj-digital: I may have missed something, but `aValidActionPlanWithOneGoal()` seems ok to me?! Perhaps the goals array argument could be changed to a single `goal`, so it can be changed by a test if necessary, but otherwise I think it's ok! :)

mattwills-moj-digital (dismissed):

integration_tests/e2e/overview/preInductionOverview.cy.ts
```diff
@@ -13,14 +14,18 @@ context('Prisoner Overview page - Pre Induction', () => {
...
     cy.task('stubGetCiagProfile404Error')
-    cy.task('getActionPlan')
-    cy.task('stubLearnerProfile')
-    cy.task('stubLearnerEducation')
+    cy.task('getActionPlan', 'G6115VJ')
+    cy.task('getActionPlan', 'A00001A')
+    cy.task('stubLearnerProfile', 'G6115VJ')
+    cy.task('stubLearnerProfile', 'A00001A')
+    cy.task('stubLearnerEducation', 'G6115VJ')
+    cy.task('stubLearnerEducation', 'A00001A')
```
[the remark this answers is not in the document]
nathanrussell-moj-digital: Yeah, agreed. I think many of our integration tests could do with looking at in this respect because they all setup stuff like this. I'm sure there's a way of reducing the noise and duplication across all of our tests 👍 
Like you say, one for another day in another PR 👍

server/testsupport/actionPlanTestDataBuilder.ts
nathanrussell-moj-digital: I think that was the problem though - the method name says its an action plan with one goal, but the method arg allows you to specify any number of goals. And I was after setting up an action plan with no goals, so I really didnt want to use this method.

But you are right that instead of deprecating the method we could change the method arg to only take a single goal. Leave it with me  ..... 👍

PR #375, opened 2024-01-04
RR-506 Add another step whilst editing a goal

chrimesdev:
## Description

Add another step whilst editing a goal

https://dsdmoj.atlassian.net/browse/RR-506

https://github.com/ministryofjustice/hmpps-education-and-work-plan-ui/assets/11615501/6e890b8b-1b6b-4334-b1e0-28bff9967f32

### DIFF
--- a/server/routes/createGoal/createGoalController.ts
+++ b/server/routes/createGoal/createGoalController.ts
@@ -83,8 +83,9 @@ export default class CreateGoalController {
     req.session.newGoal.addStepForms = req.session.newGoal.addStepForms || []
 
     if (isEditMode(req)) {
-      req.session.newGoal.addStepForm =
-        req.session.newGoals[parseInt(goalIndex, 10) - 1].addStepForms[parseInt(stepIndex, 10) - 1]
+      req.session.newGoal.addStepForm = req.session.newGoals[parseInt(goalIndex, 10) - 1].addStepForms[
+        parseInt(stepIndex, 10) - 1
+      ] || { stepNumber: parseInt(stepIndex, 10) }
     } else if (!req.session.newGoal.addStepForm) {
       req.session.newGoal.addStepForm = { stepNumber: 1 }
     }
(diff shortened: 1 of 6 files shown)


server/routes/createGoal/createGoalController.ts
```diff
@@ -83,8 +83,9 @@ export default class CreateGoalController {
     req.session.newGoal.addStepForms = req.session.newGoal.addStepForms || []
 
     if (isEditMode(req)) {
-      req.session.newGoal.addStepForm =
-        req.session.newGoals[parseInt(goalIndex, 10) - 1].addStepForms[parseInt(stepIndex, 10) - 1]
+      req.session.newGoal.addStepForm = req.session.newGoals[parseInt(goalIndex, 10) - 1].addStepForms[
```
nathanrussell-moj-digital, line 86: I know I helped you with this earlier, but I wonder if we can change it slightly to make it a little easier to read. How about something like:
```
      const newGoalForm =  req.session.newGoals[parseInt(goalIndex, 10) - 1]
      let addStepForm = newGoalForm.addStepForms[parseInt(stepIndex, 10) - 1]
      // In the case of adding a new step in edit mode there won't be an existing AddStepForm on the session so we need to instantiate one
      if (!addStepForm) {
        addStepForm = { stepNumber: parseInt(stepIndex, 10) }
      }
      req.session.newGoal.addStepForm = addStepForm
```

I know its quite long and wordy (and perhaps there's a more concise way of expressing it?), but I think its perhaps more readable like this?

What do you think?

chrimesdev: Makes sense and more readable

nathanrussell-moj-digital: Having said that .... I'm not sure now ... (sorry! 😁 )

What I'm not sure about is whether this code needs to exist here at all .... its hard to explain in text so I'll check out your branch and have a tinker, and then perhaps we can do a little pairing so I can better explain my thought .... give me a few mins ....

nathanrussell-moj-digital (approved):
LGTM 👍

PR #377, opened 2024-01-05
RR-550 update dependencies to latest versions and match the typescript template

chrimesdev:
## Description 

Update dependencies to latest versions and match the typescript template.

There have been a few minor code changes to acomodate the new versions, comments have been left.

https://dsdmoj.atlassian.net/browse/RR-550


server/data/cache/redisStore.ts
```diff
@@ -24,6 +24,8 @@ export default abstract class RedisStore {
   }
 
   private async ensureConnected(): Promise<void> {
-    return !this.redisClient.isOpen ? this.redisClient.connect() : null
+    if (!this.redisClient.isOpen) {
+      await this.redisClient.connect()
```
chrimesdev, line 28: https://github.com/ministryofjustice/hmpps-template-typescript/blob/main/server/data/tokenStore/redisTokenStore.ts#L15

server/middleware/authorisationMiddleware.ts
```diff
@@ -1,4 +1,4 @@
-import jwtDecode from 'jwt-decode'
+import { jwtDecode } from 'jwt-decode'
```
chrimesdev, line 1: https://github.com/ministryofjustice/hmpps-template-typescript/blob/486b956683dee06fcaa1332f23233a2f59f048f8/server/middleware/authorisationMiddleware.ts#L1

.eslintrc.json
```diff
@@ -98,6 +98,7 @@
         "printWidth": 120,
         "semi": false
       }
-    ]
+    ],
+    "no-empty-function": ["error", { "allow": ["constructors", "arrowFunctions"] }]
```
chrimesdev, line 102: https://github.com/ministryofjustice/hmpps-template-typescript/blob/486b956683dee06fcaa1332f23233a2f59f048f8/.eslintrc.json#L100C1-L100C1

cypress.config.ts
```diff
@@ -22,7 +22,6 @@ export default defineConfig({
   reporterOptions: {
     configFile: 'reporter-config.json',
   },
-  videoUploadOnPasses: false,
```
chrimesdev, line 25: Redundant config in newer Cypress, replaced with https://github.com/ministryofjustice/hmpps-template-typescript/blob/486b956683dee06fcaa1332f23233a2f59f048f8/package.json#L26

server/data/cache/redisStore.ts
chrimesdev: Also suggested by GitHub co-pilot!

> The error message you're seeing is indicating that the return type of the ensureConnected function is not void as declared, but instead it's of type RedisClientType.
> In TypeScript, the void type is used when a function doesn't return a value. However, in your ensureConnected function, you're returning either the result of this.redisClient.connect() or null.
> Here's the corrected version of your ensureConnected function:

```
private async ensureConnected(): Promise<void> {
  if (!this.redisClient.isOpen) {
    await this.redisClient.connect();
  }
}
```

> In this version, the function doesn't explicitly return a value, so its return type is void. If this.redisClient.isOpen is false, the function waits for this.redisClient.connect() to complete, but it doesn't return the result.

cypress.config.ts
nathanrussell-moj-digital: I'm not sure about this - I deliberately added that line a while back, and its purpose is to keep videos of failing tests (in the same way that it keeps screenshots of failing tests)

Having a video of a failing test is more useful than the final screenshot as it shows the journey and/or context of the test, so fixing it is easier

Can we revert this line, and also the corresponding line in `packacge.json` ? (`"int-test": "cypress run",`)

Unless I'm missing something?

nathanrussell-moj-digital: Ah ha, we need to add `videos: true` to enable video support now. Done 👍

nathanrussell-moj-digital (approved):
LGTM 👍

PR #380, opened 2024-01-05
RR-501 fix bug for checking for the step index on step view

chrimesdev:
## Description

Fix bug for checking for the step index on step view

Bug comment: https://dsdmoj.atlassian.net/browse/RR-501?focusedCommentId=356513

### DIFF
--- a/server/routes/createGoal/createGoalController.ts
+++ b/server/routes/createGoal/createGoalController.ts
@@ -84,8 +84,8 @@ export default class CreateGoalController {
 
     if (isEditMode(req)) {
       req.session.newGoal.addStepForm =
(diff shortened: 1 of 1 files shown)


nathanrussell-moj-digital:
Your fix here _might_ be right, but I remember putting that code in place (in that order) specifically for the main create journey (I think); and I want us to be sure we dont introduce a regression in that by making this fix.
You are probably right in what you've done here, but let's look at it together on Monday 👍

nathanrussell-moj-digital (approved):
👍

nathanrussell-moj-digital (approved):
👍

PR #384, opened 2024-01-08
RR-526 - Refactor CIAG Induction mappers into 1 source file

nathanrussell-moj-digital:
This PR is an enabler for the rest of RR-526 where we need to write mappers that map from the new `InductionResponse` into the UI view models `WorkAndInterests` and `EducationAndTraining`

Currently we have some mapper files which map from the CIAG Induction response classes into the UI view models `WorkAndInterests` and `EducationAndTraining` via functions called `toWorkAndInterests`, `toEducationAndTraining` and  `toInductionQuestionSet`
The function names are fine as are the return types. But because typescript does not allow function overloading I cant overload these functions with new input args (ie. InductionResponse)

Ultimately we are going to delete these functions (ie. those that map from the CIAG Induction response classes)

But before I can write our new ones I need to get these ones out of the way!

So, I've simply moved the contents of the 3 source files (and their corresponding tests) into a source file of its own. 

Once this is merged then I can start to write the new mappers to map from the new `InductionResponse` into the UI view models `WorkAndInterests` and `EducationAndTraining` 👍


mattwills-moj-digital (approved):
What's not to love?

PR #388, opened 2024-01-10
RR-526 - Added inductionService

nathanrussell-moj-digital:
This PR adds a new service class `InductionService`

In terms of it's interface it is equivalent to the existing `CiagInductionService` having methods to get the work & interests, education & training, and to determine if a prisoner has an induction or not, but in terms of it's implementation it uses the PLP API client rather than the CIAG API client

At the moment this just introduces this new service class and it's unit tests. It is not wired in yet. That will be my next PR where I remove the `CiagInductionService` and replace its use with this new service class etc.

Also in this PR I've re-imported the PLP API swagger spec to get the recently added `prisonNumber` field in the `InductionResponse` 👍

### DIFF
--- a/server/@types/educationAndWorkPlanApi/index.d.ts
+++ b/server/@types/educationAndWorkPlanApi/index.d.ts
@@ -1707,6 +1707,11 @@ export interface components {
        * @example 814ade0a-a3b2-46a3-862f-79211ba13f7b
        */
       reference: string
+      /**
+       * @description The ID of the prisoner
+       * @example A1234BC
+       */
+      prisonNumber: string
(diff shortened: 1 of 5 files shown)


server/@types/educationAndWorkPlanApi/index.d.ts
```diff
@@ -1707,6 +1707,11 @@ export interface components {
        * @example 814ade0a-a3b2-46a3-862f-79211ba13f7b
        */
       reference: string
+      /**
```
nathanrussell-moj-digital, line 1710: I reimported the swagger spec, hence this change

mattwills-moj-digital (approved):
AToAB ❤️

PR #389, opened 2024-01-10
RR-526 - Use new inductionService in OverviewController

nathanrussell-moj-digital:
This PR wires the new `InductionService` into the `OverviewController` so that it gets the Induction from the new API rather than the old CIAG based one.
This change is feature toggled so that we can easily switch between the old and new APIs as part of testing

This PR also updates the wiremock stubs used by the integration tests (removed the old CIAG API based stubs, replaced with stubs that reflect the new API) and updates all the integration tests that use them.

I've also taken the opportunity to abstract some of the CIAG terminology used in the integration tests to simply Induction, which in turn has resulted in a couple of changes to the nunjucks templates markup in respect of `data-qa` attributes (eg: `hasCiagInductionApiUnavailableMessageDisplayed` has been renamed to `hasInductionUnavailableMessageDisplayed` etc)


chrimesdev (dismissed):

integration_tests/mockApis/educationAndWorkPlanApi.ts
```diff
@@ -212,6 +212,317 @@ const stubGetTimeline500Error = (prisonNumber = 'G6115VJ'): SuperAgentRequest =>
...
+      method: 'GET',
+      urlPattern: `/inductions/${prisonNumber}`,
+    },
+    response: {
+      status: 404,
+      headers: { 'Content-Type': 'application/json;charset=UTF-8' },
+      jsonBody: {
+        status: 404,
+        errorCode: null,
+        userMessage: `CIAG profile does not exist for offender ${prisonNumber}`,
```
mattwills-moj-digital, line 500: `Induction not found for prisoner [A8126D]`

mattwills-moj-digital (approved):

PR #390, opened 2024-01-11
RR-392 display the prison name from the prison search register api

chrimesdev:
## Description 

Display the prison name from the prison search register API

https://dsdmoj.atlassian.net/browse/RR-392

screencapture-localhost-3000-plan-G9981UK-view-support-needs-2024-01-11-09_48_53

### DIFF
--- a/server/routes/overview/overviewController.test.ts
+++ b/server/routes/overview/overviewController.test.ts
@@ -17,6 +17,7 @@ import aValidLongQuestionSetWorkAndInterests from '../../testsupport/workAndInte
 import aValidPrisonerSummary from '../../testsupport/prisonerSummaryTestDataBuilder'
 import { aValidShortQuestionSetEducationAndTraining } from '../../testsupport/educationAndTrainingTestDataBuilder'
 import TimelineService from '../../services/timelineService'
+import PrisonService from '../../services/prisonService'
 import aValidTimeline from '../../testsupport/timelineTestDataBuilder'
 import filterTimelineEvents from '../timelineResolver'
 
@@ -41,12 +42,16 @@ describe('overviewController', () => {
   const timelineService = {
     getTimeline: jest.fn(),
   }
+  const prisonService = {
+    getPrisonByPrisonId: jest.fn(),
+  }
 
   const controller = new OverviewController(
     curiousService as unknown as CuriousService,
     educationAndWorkPlanService as unknown as EducationAndWorkPlanService,
     ciagInductionService as unknown as CiagInductionService,
     timelineService as unknown as TimelineService,
+    prisonService as unknown as PrisonService,
   )
 
   const req = {
--- a/server/routes/overview/overviewController.test.ts
+++ b/server/routes/overview/overviewController.test.ts
@@ -43,7 +43,7 @@ describe('overviewController', () => {
     getTimeline: jest.fn(),
   }
   const prisonService = {
-    getPrisonByPrisonId: jest.fn(),
+    lookupPrison: jest.fn(),
   }
 
   const controller = new OverviewController(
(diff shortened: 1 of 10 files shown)


server/routes/overview/overviewController.test.ts
```diff
@@ -41,12 +42,16 @@ describe('overviewController', () => {
   const timelineService = {
     getTimeline: jest.fn(),
   }
+  const prisonService = {
+    getPrisonByPrisonId: jest.fn(),
```
nathanrussell-moj-digital, line 46: It would be good to see the relevant tests in this test class make some assertions on this mock and/or stub its response.
What I mean by that is that injecting the mock into the controller is fine (and indeed you needed to do that otherwise it doesnt compile properly). But _why_ does the controller class need the Prison Service? What is is doing with it? Is it doing the right / expected thing with it?

You are making use of the Prison Service in the `getSupportNeedsView` method of the controller, so what you want to do is find the relevant tests in this test class, then setup this mock. IE. when the controller method runs it will invoke `prisonService.getPrisonByPrisonId` at least twice (at least once for the health and support needs, and at least once for the neurodiversity stuff). You need to set the mock up so that it returns a `Prison`
You then also need to assert (in the `// Then` section) that the mock was called the expected number of times and with the expected arguments (ie. in this case the prison ID and token)

server/routes/overview/overviewController.ts
```diff
@@ -99,6 +101,31 @@ export default class OverviewController {
     const { prisonerSummary } = req.session
 
     const supportNeeds = await this.curiousService.getPrisonerSupportNeeds(prisonNumber, req.user.username)
+
+    // Loop through the healthAndSupport needs array and update the prison name for each need
+    await Promise.all(
+      supportNeeds.healthAndSupportNeeds.map(async supportNeed => {
+        const prison = await this.prisonService.getPrisonByPrisonId(supportNeed.prisonId, req.user.token)
```
nathanrussell-moj-digital, line 108: This call (and the one below for neurodiversities below) are using the wrong token
It works for you because your user token has some additional privs (including those to be able to call the prison-register-api), but not all users have these roles
Have a look at how this service is called by the timeline, but you want to be using the system token, not a user token

server/routes/overview/overviewController.ts
```diff
@@ -99,6 +101,31 @@ export default class OverviewController {
...
+        if (prison) {
```
nathanrussell-moj-digital, line 109: I may not have written this prisonService and timelineService in the best way originally 😬 , but by calling the prisonService directly like this you are missing a trick in respect of try/catch error handling.
As you have it at the moment if there is an error calling the prisonService the error is not caught and the page will fail (and it might even crash the node app! 😬 )
Have a look at how the timeline controller does this, but there is an intermediate method that it calls (in timelineService) called `lookupPrison` which does the try/catch

I'm not suggesting this method for getting support needs needs to use the timelineService (that would be wrong), but have a look at how it currently works and see if you can do something similar here
(hint - it might be a case of moving `lookupPrison` out of timelineService and into prisonService so that it can be used by things other than the timeline)

server/services/prisonService.test.ts
```diff
@@ -208,4 +214,36 @@ describe('prisonService', () => {
       expect(prisonRegisterStore.setActivePrisons).toHaveBeenCalledWith(allPrisons, 1)
     })
   })
+
+  describe('lookupPrison', () => {
+    it('should lookup prison by ID', async () => {
```
nathanrussell-moj-digital, line 219: Thank you for having a stab at writing the tests, but theres a couple of things not quite right here

Firstly, what you've written is a test of a mock - you have ended up testing a mock! 🤣 
On line 224 you setup the `prisonService.lookupPrison` as a mock, but then on line 227 you invoke that same mock!

Line 227 is correct - we want to test the new method you've added to prisonService, ie. lookupPrison

But the thing you need to mock is not the method you are testing, but instead the internal dependency of that method. In this case it would probably be the prisonRegisterStore (I'd have to check which methods call what etc, but I think thats right). 

The 2nd thing thats not right is your second test. What we want to be testing here is the catch block of the `prisonService.lookupPrison` method. In order to do that we need to setup the mock (probably prisonRegisterStore - see above) to throw an exception. By doing that we then force prisonService.lookupPrison to go into it's catch block 👍 

I'll checkout your branch in a mo and see if I can help with it

nathanrussell-moj-digital:
@chrimesdev ; re my PR comment, have a look at my last commit and see what you think. Can you see / follow the difference between what you did in these tests vs. how I've updated them?

nathanrussell-moj-digital (approved):

server/services/prisonService.ts
```diff
@@ -27,8 +27,8 @@ export default class PrisonService {
   }
 
   async lookupPrison(prisonId: string, username: string): Promise<Prison> {
-    const systemToken = await this.hmppsAuthClient.getSystemClientToken(username)
     try {
+      const systemToken = await this.hmppsAuthClient.getSystemClientToken(username)
```
chrimesdev, line 31: Good spot

PR #393, opened 2024-01-19
RR-526 - Enable feature toggle in all envs to use new Induction API

nathanrussell-moj-digital:
This PR enables the `USE_NEW_INDUCTION_API_ENABLED` feature toggle in all environments which tells the PLP UI to use the new Induction API to retrieve inductions, rather than the old/migrated CIAG API

This feature has been enabled in `dev` for over a week now, so I am confident we can roll this out and promote to the other environments


mattwills-moj-digital (approved):
GS

PR #394, opened 2024-01-22
RR-576 update to latest version of govuk-frontend v5

chrimesdev:
## Description

Update to latest version of govuk-frontend `v5.0.0` which includes some breaking changes.

Mainly around dropping support for Internet Explorer, changing the paths of where things are served up from the `govuk-frontend` package and the handling of the UI JavaScript.

Release notes/guide: https://github.com/alphagov/govuk-frontend/releases/tag/v5.0.0

Change tracker: https://docs.google.com/spreadsheets/d/1TU-6a0wog2A0jMDJO0Bf-m9XhUHxJjDfYBHSGlGiXME/edit#gid=0

### JIRA ticket

https://dsdmoj.atlassian.net/browse/RR-576


nathanrussell-moj-digital (commented):
From a technical perspective what you've done looks absolutely fine.

I am however slightly concerned by the change that we need to do re using the govuk javascript in the browser as an ES6 module.

[CanIUse](https://caniuse.com/es6-module) tells us that ES6 modules are not supported in IE 11 or below, and therefore the only MS browser that this will work with is Edge.

This is also noted in the govUK frontend v5 release notes, where they say:

> If you still need to provide support for older versions of Internet Explorer, you should stay on the latest [GOV.UK Frontend v4 release](https://github.com/alphagov/govuk-frontend/releases).

I'll be honest, I don't know our user community and their browser use (perhaps I should 😬 ). What I do know is that they are "internal" users, who are using HMPPS/Prison provided computers. How up to date those computers are in respect of OS/browsers etc, I really dont know.

I think before we merge this PR I think we should find out.

Can you check with James/Liora/our URs etc to see if they know what browsers people use in prisons / on prison provided kit please?

In the mean time I'll see if there are any stats/evidence I can find from the App Insights logs 👍

nathanrussell-moj-digital:
I've checked in App Insights re: browser usage, and it's all `Edge 120` (the Mac / Chrome would be me as part of smoke testing the other week)
Screenshot 2024-01-23 at 09 02 04

On that basis I'm happy to approve, but I'd still like us to check/validate this knowledge re: browser usage with the team please.

nathanrussell-moj-digital (approved):
LGTM 👍

PR #400, opened 2024-01-30
RR-611 Feature toggle and service method

mattwills-moj-digital:
This PR adds a new feature toggle `includePrisonTimelineEventsEnabled` to ensure `Prison Movement` events do not appear on the timeline until this feature is complete. It also adds the requiring filtering to `timelineService.ts` and updates the corresponding unit test.


server/services/timelineService.test.ts
```diff
@@ -7,6 +7,7 @@ import { EducationAndWorkPlanClient } from '../data'
 import { toTimeline } from '../data/mappers/timelineMapper'
 import aValidTimelineResponse from '../testsupport/timelineResponseTestDataBuilder'
 import aValidTimeline from '../testsupport/timelineTestDataBuilder'
+import config from '../config'
 
 jest.mock('../data/mappers/timelineMapper')
```
mattwills-moj-digital, line 13: I've changed a lot in this test, which is difficult to make out from the diff (partly because I've changed the grouping and also replaced some of the JSON with builders). It might be easier to view the file.

server/services/timelineService.test.ts
```diff
@@ -52,48 +53,141 @@ describe('timelineService', () => {
...
+            contextualInfo: 'Learn French',
             actionedByDisplayName: 'Ralph Gen',
           },
+        ],
+      }
+      const timeline: Timeline = aValidTimeline()
+      mockedTimelineMapper.mockReturnValue(timeline)
+
+      // When
+      await timelineService.getTimeline(prisonNumber, userToken, username)
```
nathanrussell-moj-digital, line 120: Do we no longer want to capture the return value from the method as `actual` and then make some kind of assertion on it? 🤔

server/services/timelineService.test.ts
```diff
@@ -52,48 +53,141 @@ describe('timelineService', () => {
...
+
+      // Then
+      expect(mockedTimelineMapper).toHaveBeenCalledWith(expectedFilteredTimelineResponse)
+    })
+
+    it('should get timeline without unsupported events', async () => {
```
nathanrussell-moj-digital, line 126: I'm not sure I follow the test name in respect of "without unsupported", and then the apparent significance of the prison timeline events feature toggle being off ?

server/services/timelineService.ts
```diff
@@ -69,6 +71,11 @@ export default class TimelineService {
     return Promise.all(eventPromises)
   }
 
-  private filterTimelineEvents = (event: TimelineEventResponse): boolean =>
-    SUPPORTED_TIMELINE_EVENTS.includes(event.eventType)
+  private filterTimelineEvents = (event: TimelineEventResponse): boolean => {
```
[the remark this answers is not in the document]
mattwills-moj-digital: I wasn't 100% happy with what I had. I originally wanted to combine two arrays, rather than having the "PLP" events listed twice in two separate arrays, as you have above (the first of which ends up being spread over several lines due to the ESLint rules!).

I originally tried the `push()` function, but it returns an integer (number of items added), rather than the array itself, however, I've just found the `concat()` function instead. What do you think of this?

```
// TODO RR-610 - remove feature toggle and revert back to one SUPPORTED_TIMELINE_EVENTS array
const PLP_TIMELINE_EVENTS = ['ACTION_PLAN_CREATED', 'INDUCTION_UPDATED', 'GOAL_UPDATED', 'GOAL_CREATED']
const PRISON_TIMELINE_EVENTS = ['PRISON_ADMISSION', 'PRISON_RELEASE', 'PRISON_TRANSFER']

const SUPPORTED_TIMELINE_EVENTS = config.featureToggles.includePrisonTimelineEventsEnabled
  ? PLP_TIMELINE_EVENTS.concat(PRISON_TIMELINE_EVENTS)
  : PLP_TIMELINE_EVENTS
```

server/services/timelineService.test.ts
mattwills-moj-digital: We could do, but it will be whatever's return from the mocked mapper: 
```
mockedTimelineMapper.mockReturnValue(timeline)
```
I guess I could add it for regression purposes, but the point of this test is to make sure that the mapper is called with the correct filtered items.

server/services/timelineService.test.ts
mattwills-moj-digital: The intention of the test is to make sure that 'Prison Movement' events are filtered out and not passed to the mapper.

server/services/timelineService.ts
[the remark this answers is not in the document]
mattwills-moj-digital: Unfortunately neither of the above work. I suspect these constants are initialised before `config.featureToggles.includePrisonTimelineEventsEnabled` is resolved?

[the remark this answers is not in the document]
mattwills-moj-digital: I've pushed a change which is hopefully easier to read?

server/services/timelineService.test.ts
mattwills-moj-digital: I've added an assertion for the returned value.

nathanrussell-moj-digital (approved):
LGTM 👍

PR #401, opened 2024-01-30
RR-611 Update timelineResolver ensure correct ordering

mattwills-moj-digital:
Now that we will soon be displaying Prison movement related events on the Timeline, the existing logic of always ensuring that the `ACTION_PLAN_CREATED` event appears last is no longer valid. Instead, we need to ensure that it appears after any related `GOAL_CREATED` events (i.e. events that occurred at exactly the same time), but still in reverse chronological order in relation to other unconnected events.

### DIFF
--- a/server/routes/timelineResolver.ts
+++ b/server/routes/timelineResolver.ts
@@ -66,10 +66,12 @@ const sortTimelineEvents = (timelineEvents: Array<TimelineEvent>): Array<Timelin
 }
 
 // Ensure the ACTION_PLAN_CREATED event appears after any related GOAL_CREATED events that were created at the same time. In theory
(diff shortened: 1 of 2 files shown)


server/routes/timelineResolver.ts
```diff
@@ -7,9 +8,9 @@ const filterTimelineEvents = (timeline: Timeline): Timeline => {
     return timeline
   }
 
-  // 'Flatten' multiple GOAL_CREATED events with the same correlationId into single MULTIPLE_GOALS_CREATED events
+  // Merge multiple GOAL_CREATED events with the same correlationId into single MULTIPLE_GOALS_CREATED events
```
nathanrussell-moj-digital, line 11: 👍 
Merge is definitely a better term in this context 👍 Thank you

PR #409, opened 2024-02-13
RR-569 remove create goals without induction feature toggle

chrimesdev:
## Description 

Remove create goals without induction feature toggle

https://dsdmoj.atlassian.net/browse/RR-569


server/routes/postInductionCreation/index.ts
```diff
@@ -14,10 +13,6 @@ export default (router: Router, services: Services) => {
     const userToken = req.user.token
     const { prisonNumber } = req.params
 
-    if (!config.featureToggles.createGoalsWithoutInductionEnabled) {
-      return res.redirect(createGoalRoute(prisonNumber))
-    }
-
     return (await prisonerHasActionPlan(prisonNumber, userToken, services.educationAndWorkPlanService))
       ? res.redirect(`/plan/${prisonNumber}/view/overview`) // Action Plan with goal(s) exists already. Redirect to the Overview page
       : res.redirect(createGoalRoute(prisonNumber)) // Action Plan goals do not exist yet. Redirect to the Create Goal flow routes.
```
nathanrussell-moj-digital, line 18: Minor, but given that this line is now the only line that uses the function `createGoalRoute`, and all that function does is return a string of the route for creating the first goal, I'd be tempted to put is back inline (like we do on the previous line)

(the only reason we factored it out into a function was because 2 places were using it. Thats not longer the case, and I don't believe there's anything particularly complex in the function that would preclude doing it inline here)

mattwills-moj-digital (approved):
LGTM

nathanrussell-moj-digital (approved):
LGTM 👍

PR #412, opened 2024-02-15
RR-413 use a smaller font size on the last updated field

chrimesdev:
## Description

Use a smaller font size on the last updated field

https://dsdmoj.atlassian.net/browse/RR-413

### DIFF
--- a/server/views/pages/overview/partials/educationAndTrainingTab/_qualificationsAndEducationHistory_inductionLongQuestionSet.njk
+++ b/server/views/pages/overview/partials/educationAndTrainingTab/_qualificationsAndEducationHistory_inductionLongQuestionSet.njk
@@ -88,7 +88,7 @@ as the follow up questions are different.
           </div>
         </dl>
 
-        <p class="govuk-hint">Last updated: {{ educationAndTraining.data.longQuestionSetAnswers.updatedAt | formatDate('D MMMM YYYY') }}<span class="govuk-!-display-none-print"> by {{ educationAndTraining.data.longQuestionSetAnswers.updatedBy }}</span></p>
+        <p class="govuk-hint govuk-body govuk-!-font-size-16">Last updated: {{ educationAndTraining.data.longQuestionSetAnswers.updatedAt | formatDate('D MMMM YYYY') }}<span class="govuk-!-display-none-print"> by {{ educationAndTraining.data.longQuestionSetAnswers.updatedBy }}</span></p>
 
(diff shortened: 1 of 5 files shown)


server/views/pages/overview/partials/educationAndTrainingTab/_qualificationsAndEducationHistory_inductionLongQuestionSet.njk
```diff
@@ -88,7 +88,7 @@ as the follow up questions are different.
           </div>
         </dl>
 
-        <p class="govuk-hint">Last updated: {{ educationAndTraining.data.longQuestionSetAnswers.updatedAt | formatDate('D MMMM YYYY') }}<span class="govuk-!-display-none-print"> by {{ educationAndTraining.data.longQuestionSetAnswers.updatedBy }}</span></p>
+        <p class="govuk-hint govuk-body govuk-!-font-size-16">Last updated: {{ educationAndTraining.data.longQuestionSetAnswers.updatedAt | formatDate('D MMMM YYYY') }}<span class="govuk-!-display-none-print"> by {{ educationAndTraining.data.longQuestionSetAnswers.updatedBy }}</span></p>
```
leejacobson-moj, line 91: Just noting in case you're not already aware, but `govuk-!-font-size-16` isn't responsive. `govuk-body-s` is and would be 16px on desktop and 14px on mobile.

Not sure it matters here, but `govuk-body-s` is generally preferred for that reason unless the font size should be 16px across all viewports.

leejacobson-moj (approved):
Looks good. Approved.

server/views/pages/overview/partials/educationAndTrainingTab/_qualificationsAndEducationHistory_inductionLongQuestionSet.njk
chrimesdev: `govuk-body-s` is getting overriden by the `govuk-hint` class as its more specific, so I've had to go with a utility class which uses `!important` to override `govuk-hint`.

chrimesdev: I think some other services have created a new class name `app-secondary-text-color` so they can use that instead of `govuk-hint` - which we could do here

mattwills-moj-digital (approved):

PR #414, opened 2024-02-15
RR-640 - render multiline text

nathanrussell-moj-digital:
PR to render multiline text (goal and step titles; goal notes)
Sat behind a feature toggle in `dev` only so that the team can properly play with it and test it


server/filters/formatMultilineTextFilter.ts
```diff
@@ -0,0 +1,7 @@
+export default function formatMultilineTextFilter(value: string): string {
+  if (!value) {
+    return undefined
+  }
+
+  return value.replace(/\n/gm, '<br />')
+}
```
nathanrussell-moj-digital, line 7: Its a really really simple filter - take the text and go a global regex search and replace on `\n`, replacing it with the HTML `<br />`

server/views/pages/goal/review/index.njk
```diff
@@ -28,7 +28,7 @@
                 Description
               </dt>
               <dd class="govuk-summary-list__value" data-qa="goal-{{ goalLoop.index }}-description-value">
-                {{ goal.title }}
+                {{ goal.title | formatMultilineText | safe }}
```
nathanrussell-moj-digital, line 31: The only gotcha is that we need to also use the `safe` filter, because the text now has HTML in it (ie. the `<br />` from the filter)

server/filters/formatMultilineTextFilter.ts
chrimesdev: Regex checks out 👍🏻

chrimesdev:
LGTM 👍🏻 

The only thing we may need to consider if someone (for whatever reason) adds lots of new lines (see screenshot). Whether we're happy with the rendering of or we want to do some clever formatting - but its probably going to be rare.

screencapture-localhost-3000-plan-A5502DZ-goals-review-2024-02-15-17_51_54

chrimesdev (approved):

mattwills-moj-digital (approved):

nathanrussell-moj-digital:
> LGTM 👍🏻
> 
> The only thing we may need to consider if someone (for whatever reason) adds lots of new lines (see screenshot). Whether we're happy with the rendering of that, or we want to do some clever formatting? But its probably going to be rare.
> 

Yeah, I had the same thought (though didnt run up an example like that 😁 )
But this is exactly why I want it feature toggled so we can allow the wider team to do precisely this type of thing; then they can decide for themselves whether we roll it out or not 👍

nathanrussell-moj-digital:
Also, there is a built in filter that does exactly what my filter is doing:
https://mozilla.github.io/nunjucks/templating.html#nl2br

It has a terrible name 🤣 , but I think I'd rather use an inbuilt filter than one we've custom written.
However, I couldnt work out how to feature toggle an inbuilt filter like I've done here. My thinking is we can use this one to allow the team to test and see if its something they want to do. If it is, then we remove the feature toggle, and the new custom filter (even though it has a great name! 😁 ), and replace the calls to it in the nunjucks with `nl2br`

PR #415, opened 2024-02-16
RR-640 - enable render multiline text in `dev`

nathanrussell-moj-digital:

PR #418, opened 2024-02-19
RR-640 - enable render multiline text in all envs

nathanrussell-moj-digital:
This PR  enables rendering of multiline text in all envs.

This was enabled in `dev` last week via this PR #415 and has been tested by James. James has now asked that we enable this is all envs and rollout to `prod`

Ordinarily this would be just a case of removing the feature toggle and associated logic. However in this case there is a little more to it ....

In #415 I wrote a nunjucks filter called `formatMultilineTextFilter` which replaces new line characters with `<br />`; and then I conditionally added it to the nunjucks environment filters, based on whether the feature toggle was enabled.

At the time I knew there was an inbuilt filter that does exactly what my new filter does - the terribly named [nlbr](https://mozilla.github.io/nunjucks/templating.html#nl2br)
I didn't want to re-invent the wheel, but at the time I didn't know how to conditionally add a filter to the nunjucks environment, given that it would already have been added by the standard nunjucks setup. 
The other thing I didnt like is the name `nl2br` as its not obvious what it does, especially when reading it in the context of reading a nunjucks template file.
Hence I did what I did at the time.

What I've done here is remove the feature toggle; and also removed my custom filter `formatMultilineTextFilter`. But in our nunjucks setup, where I'd defined our filter `formatMultilineText` to use the (now deleted custom filter), I've made it use the existing inbuilt one.
This means I am using standard inbuilt filter functionality (`nl2br`) , but I've given it a new name of `formatMultilineText`, which means the nunjucks views are better/clearer to read in this respect.


mattwills-moj-digital (approved):
GS

chrimesdev (approved):
LGTM!

PR #419, opened 2024-02-19
RR-567 add the separate skill type filters/data to the functional skills view

chrimesdev:
## Description

Create a new resolver to get functional skills by type, to allow us to pass separate skill types into the functional skills view page.

Adding the views and using this data will be done in a following PR.

https://dsdmoj.atlassian.net/browse/RR-567

### DIFF
--- a/server/routes/functionalSkills/functionalSkillsController.ts
+++ b/server/routes/functionalSkills/functionalSkillsController.ts
@@ -1,7 +1,7 @@
 import { RequestHandler } from 'express'
 import { CuriousService } from '../../services'
 import FunctionalSkillsView from './functionalSkillsView'
-import { allFunctionalSkills, mostRecentFunctionalSkills } from '../functionalSkillsResolver'
+import { allFunctionalSkills, mostRecentFunctionalSkills, functionalSkillsByType } from '../functionalSkillsResolver'
 
 export default class FunctionalSkillsController {
   constructor(private readonly curiousService: CuriousService) {}
@@ -17,10 +17,17 @@ export default class FunctionalSkillsController {
     const latestFunctionalSkillsFromCurious = mostRecentFunctionalSkills(functionalSkillsFromCurious)
     const allFunctionalSkillsFromCurious = allFunctionalSkills(functionalSkillsFromCurious)
 
+    const englishSkills = functionalSkillsByType(allFunctionalSkillsFromCurious.assessments, 'ENGLISH')
+    const mathsSkills = functionalSkillsByType(allFunctionalSkillsFromCurious.assessments, 'MATHS')
+    const digitalSkills = functionalSkillsByType(allFunctionalSkillsFromCurious.assessments, 'DIGITAL_LITERACY')
+
     const view = new FunctionalSkillsView(
       prisonerSummary,
       latestFunctionalSkillsFromCurious,
       allFunctionalSkillsFromCurious,
+      englishSkills,
+      mathsSkills,
+      digitalSkills,
     )
     res.render('pages/functionalSkills/index', { ...view.renderArgs })
   }
--- a/server/routes/functionalSkills/functionalSkillsController.ts
+++ b/server/routes/functionalSkills/functionalSkillsController.ts
@@ -17,7 +17,7 @@ export default class FunctionalSkillsController {
     const latestFunctionalSkillsFromCurious = mostRecentFunctionalSkills(functionalSkillsFromCurious)
     const allFunctionalSkillsFromCurious = allFunctionalSkills(functionalSkillsFromCurious)
 
-    const problemRetrievingData = !!functionalSkillsFromCurious.problemRetrievingData
+    const { problemRetrievingData } = functionalSkillsFromCurious
     const englishSkills = functionalSkillsByType(functionalSkillsFromCurious.assessments, 'ENGLISH')
     const mathsSkills = functionalSkillsByType(functionalSkillsFromCurious.assessments, 'MATHS')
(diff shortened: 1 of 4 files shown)


server/routes/functionalSkillsResolver.ts
```diff
@@ -77,4 +77,8 @@ const assessmentsGroupedByTypeSortedByDateDesc = (assessments: Array<Assessment>
   return map
 }
 
-export { allFunctionalSkills, mostRecentFunctionalSkills }
+const functionalSkillsByType = (assessments: Array<Assessment>, type: Assessment['type']): Array<Assessment> => {
```
nathanrussell-moj-digital, line 80: That's cool 👍 , I didn't know you could reference a field to use it's type
```
type: Assessment['type']
```
(clearly the word "type" is overloaded in this comment, but I know what I'm trying to say 😁 . I'm trying to say its cool 😁 )

nathanrussell-moj-digital: I know we didnt talk about this earlier, but I wonder if this method should return the assessments sorted? Probably sorted by assessment date, given the UI use case is that the default sort order should be by date?

If we are going to sort them, then I think you'll want to destructure the array into a new array before filtering and sorting (I know we dont do that elsewhere in this class, but I think its possibly a bug, thinking about one of the problems we had with sorting arrays last week)
```
return [...assessments].filter(.....).sort(....)
```
And I think you can use the `dateComparator` that we already have 🤔

server/routes/functionalSkillsResolver.test.ts
```diff
@@ -209,4 +209,47 @@ describe('functionalSkillsResolver', () => {
...
+        { type: 'MATHS', grade: 'Level 3', assessmentDate: YESTERDAY },
+      ] as Array<Assessment>
+
+      const expected = [{ type: 'ENGLISH', grade: 'Level 1', assessmentDate: TEN_DAYS_AGO }] as Array<Assessment>
+
+      // When
+      const actual = functionalSkillsByType(assessments, 'ENGLISH')
+
+      // Then
+      expect(actual).toEqual(expected)
```
nathanrussell-moj-digital, line 232: Given that you change the implementation to sort by date (see my other comment), then I think it would be worth changing this test so that we are filtering by type `DIGITAL_LITERACY`, because there are 3 of them with various assessment dates. This would then test both the filtering and sorting 👍

server/routes/functionalSkillsResolver.ts
chrimesdev: Good early spot with the sort order, wouldn't have noticed that until the UI changes!

server/routes/functionalSkillsResolver.test.ts
chrimesdev: Makes sense

nathanrussell-moj-digital: Thats a much better test; perfect ❤️ 
Thanks 👍

server/routes/functionalSkills/functionalSkillsController.ts
```diff
@@ -17,10 +17,17 @@ export default class FunctionalSkillsController {
...
 
+    const englishSkills = functionalSkillsByType(functionalSkillsFromCurious.assessments, 'ENGLISH')
+    const mathsSkills = functionalSkillsByType(functionalSkillsFromCurious.assessments, 'MATHS')
+    const digitalSkills = functionalSkillsByType(functionalSkillsFromCurious.assessments, 'DIGITAL_LITERACY')
+
     const view = new FunctionalSkillsView(
       prisonerSummary,
       latestFunctionalSkillsFromCurious,
       allFunctionalSkillsFromCurious,
+      englishSkills,
```
nathanrussell-moj-digital, line 28: OK, you're gonna hate me for not mentioning this before (but I only remembered last night), but we're going to need one more field in the view class `FunctionalSkillsView`
We need `problemRetrievingData: boolean` ..... because currently it exists in the 2 fields `latestFunctionalSkillsFromCurious` and `allFunctionalSkillsFromCurious` ; and the nunjucks view reads it from `allFunctionalSkillsFromCurious.problemRetrievingData`

But ... shortly (in a couple of PRs time), you're going to remove these 2 fields from the view class

So, I think you want a new `problemRetrievingData: boolean` on `FunctionalSkillsView`, and you'll provide its value here on the constructor call from `functionalSkillsFromCurious.problemRetrievingData`

Happy to talk it through if I've not made much sense here 😁

chrimesdev: I've added this; let me know what you think of the eslint error fix approach though in the below comment.

server/routes/functionalSkills/functionalSkillsController.ts
```diff
@@ -17,6 +17,7 @@ export default class FunctionalSkillsController {
     const latestFunctionalSkillsFromCurious = mostRecentFunctionalSkills(functionalSkillsFromCurious)
     const allFunctionalSkillsFromCurious = allFunctionalSkills(functionalSkillsFromCurious)
 
+    const problemRetrievingData = !!functionalSkillsFromCurious.problemRetrievingData
```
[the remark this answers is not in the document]
nathanrussell-moj-digital: Yeah, I know what you mean. What if you declared the type? Would this work?

```
const problemRetrievingData: Boolean = functionalSkillsFromCurious.problemRetrievingData
```

nathanrussell-moj-digital (approved):
LGTM 👍

PR #420, opened 2024-02-19
RR-642 - Remove the CIAG API generated types; replace with types from PLP API

nathanrussell-moj-digital:
This PR (almost 😁 ) removes the last hang-over of the original CIAG API from the PLP UI 👍 🎉 

It removes the generated types from the CIAG API swagger spec (which would fail if we had to generate them again because the API really no longer exists any more), and instead uses the types from the PLP API's imported types.

We are still left with the `ciagClient` as it still has the method `getCiagInductionsForPrisonNumbers` which is used by the Prisoner List page, but that API endpoint has been in PLP API for some time, and the client class has been configured to use if from there for some time as well.

I think what I really wanted to do was to delete `ciagClient` altogether and move the method `getCiagInductionsForPrisonNumbers` into `educationAndWorkPlanClient` (PLP API Client); but I think we'd be better off writing a new endpoint in the PLP API for retrieving induction summaries for a list of prison numbers, then doing away with all the `/ciag/induction` endpoints altogether 👍


README.md
```diff
@@ -65,12 +65,10 @@ To update the types from the Open API docs run the following commands:
...
 `npx openapi-typescript https://testservices.sequation.net/sequation-virtual-campus2-api/v3/api-docs -o server/@types/curiousApi/index.d.ts`
 
-`npx openapi-typescript https://ciag-induction-api-dev.hmpps.service.justice.gov.uk/v3/api-docs -o server/@types/ciagInductionApi/index.d.ts`
-
 Note that you will need to run prettier over the generated files and possibly handle other errors before compiling.
 
-The types are inherited for use in `server/@types/educationAndWorkPlanApi/index.d.ts`, `server/@types/prisonerSearchApi/index.d.ts` 
-and `server/@types/curiousApi/index.d.ts` which may also need tweaking for use.
+The types are inherited for use in `server/@types/educationAndWorkPlanApi/index.d.ts`, `server/@types/prisonerSearchApi/index.d.ts`, 
+`server/@types/prisonRegisterApi/index.d.ts` and `server/@types/curiousApi/index.d.ts` which may also need tweaking for use.
```
mattwills-moj-digital, line 71: 🙌

server/testsupport/ciagInductionListResponseTestDataBuilder.ts
```diff
@@ -1,9 +1,9 @@
-import type { CiagInduction, CiagInductionListResponse } from 'ciagInductionApiClient'
+import type { CiagInductionSummaryListResponse, CiagInductionSummaryResponse } from 'educationAndWorkPlanApiClient'
 import { aLongQuestionSetCiagInduction } from './ciagInductionTestDataBuilder'
 
 export default function aValidCiagInductionListResponse(options?: {
-  ciagProfileList?: CiagInduction[]
-}): CiagInductionListResponse {
+  ciagProfileList?: CiagInductionSummaryResponse[]
```
mattwills-moj-digital, line 5: This change makes sense, but I don't understand how it was an array of `CiagInduction` before?! 🤯

server/testsupport/ciagInductionTestDataBuilder.ts
```diff
@@ -10,7 +10,7 @@ const aLongQuestionSetCiagInduction = (options?: {
   modifiedByDateTime?: string
   workInterestModifiedBy?: string
   workInterestModifiedByDateTime?: string
-}): CiagInduction => {
+}): CiagInductionSummaryResponse => {
```
mattwills-moj-digital, line 13: In contrast, I don't understand how this is now a `Summary` response rather than a full induction. Perhaps I need to look at it with a fresh pair of eyes in the morning!

server/testsupport/ciagInductionListResponseTestDataBuilder.ts
nathanrussell-moj-digital: I'll try to explain ....

Previously the class types were from the CIAG API's imported swagger spec (so they were not the PLP classes, not even the PLP "versions" of the CIAG classes)
The "outer" object was a `CiagInductionListResponse` which had a single field `ciagProfileList` which was an array of `CiagInduction` (the full object !!)
This was based on the original implementation where we were calling the original CIAG API

At some point in time we implemented the "CIAG API" but in the PLP API. The paths and data types were identical, but we chose to give the classes different (better) names.
Our equivalent classes are `CiagInductionSummaryListResponse` and `CiagInductionSummaryResponse`.
What we didn't do at the time, which is what this change is about, is changing the UI codebase to use the new types from the PLP swagger spec. We changed the API endpoint, so it was calling the PLP API, but we didnt change these types here.

Because the types were identical (more on that in a sec) apart from name, it just worked. Perhaps more by luck than judgement, but it worked.

Now .... are the class types identical? 😁 🤔  I think you know the answer to that is no! 🤣 which I guess is perhaps where your question has come from ...

Our "outer wrapper" class `CiagInductionSummaryListResponse` is identical to the CIAG API class `CiagInductionListResponse` because they both contain a single field `ciagProfileList` which contains an array of `CiagInduction` (CIAG API class) / `CiagInductionSummaryResponse` (PLP API class)

Are our `CiagInductionSummaryResponse` PLP API class and CIAG API class `CiagInduction` identical ?
No ...  our `CiagInductionSummaryResponse` is (as the name suggests) a summary, rather than the full induction. But it contains a subset of the fields from `CiagInduction`, crucially the "important" ones that the UI reads.

Does that make sense?

nathanrussell-moj-digital: Ah .... I misunderstood your question ..... it was about the test data builder, and whether they were building the right thing, given that the PLP's summary object is indeed a summary, rather than a full induction!
Gotcha! Just making that change now 👍

mattwills-moj-digital (approved):

PR #422, opened 2024-02-20
RR-567 - Unit test for functionalSkillsController

nathanrussell-moj-digital:
Unit test for functionalSkillsController; destructure arrays in functionalSkillsResolver

### DIFF
--- a/server/routes/functionalSkillsResolver.ts
+++ b/server/routes/functionalSkillsResolver.ts
@@ -13,7 +13,7 @@ import dateComparator from './dateComparator'
 const mostRecentFunctionalSkills = (allFunctionalSkills: FunctionalSkills): FunctionalSkills => {
   return {
     ...allFunctionalSkills,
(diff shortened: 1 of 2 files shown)


server/routes/functionalSkillsResolver.ts
```diff
@@ -13,7 +13,7 @@ import dateComparator from './dateComparator'
 const mostRecentFunctionalSkills = (allFunctionalSkills: FunctionalSkills): FunctionalSkills => {
   return {
     ...allFunctionalSkills,
-    assessments: mostRecentAssessments(allFunctionalSkills.assessments || []),
+    assessments: mostRecentAssessments([...(allFunctionalSkills.assessments || [])]),
```
nathanrussell-moj-digital, line 16: As expected, the tests in the controller proved that we need to destructure the arrays before sorting and filtering in each of these method calls. 
Without destructuring we end up mutating the source object! 😬

chrimesdev (approved):

PR #425, opened 2024-02-21
RR-649 Remove the FRONTEND_COMPONENTS_API_FEATURE_TOGGLE_ENABLED feature toggle

chrimesdev:
## Description

Remove the `frontendComponentsApiToggleEnabled` feature toggle from the Nunjucks views and remove the old fallback Header. It also removes the original PLP header styles.

We still need a fallback of the Header and Footer incase of `feComponents.problemRetrievingData`

https://dsdmoj.atlassian.net/browse/RR-649


nathanrussell-moj-digital:
Why are we not removing the feature toggle? Isn't that what we spoke about?

chrimesdev:
> Why are we not removing the feature toggle? Isn't that what we spoke about?

Maybe I misunderstood the assignment, feature toggle now removed completely!

nathanrussell-moj-digital (approved):
LGTM 👍

PR #426, opened 2024-02-22
RR-647 - Add inductionDto and associated code

nathanrussell-moj-digital:
This PR adds the `inductionDto` and associated code (enums, mapper & service) that we will need for migrating the CIAG UI screens to PLP UI
Pretty much all of this has been copied verbatim from the code we added in CIAG UI (with a few minor tweaks), so should look pretty familiar


server/@types/dto/index.d.ts
```diff
@@ -30,3 +30,163 @@ declare module 'dto' {
     sequenceNumber: number
   }
 }
+
+declare module 'inductionDto' {
```
nathanrussell-moj-digital, line 34: Copied verbatim from CIAG UI, just wrapped in a different module name

server/data/mappers/inductionDtoMapper.ts
```diff
@@ -0,0 +1,59 @@
+import type { InductionResponse } from 'educationAndWorkPlanApiClient'
+import type { InductionDto } from 'inductionDto'
+
+const toInductionDto = (inductionResponse: InductionResponse): InductionDto => {
```
nathanrussell-moj-digital, line 4: This mapper, and it's test, copied verbatim from CIAG UI

server/services/inductionService.ts
```diff
@@ -19,18 +21,28 @@ export default class InductionService {
...
     try {
-      const induction = await this.getInduction(prisonNumber, token)
+      const induction = await this.retrieveInduction(prisonNumber, token)
       return toEducationAndTraining(induction)
     } catch (error) {
       return { problemRetrievingData: true } as EducationAndTraining
     }
   }
 
+  async getInduction(prisonNumber: string, token: string): Promise<InductionDto> {
```
nathanrussell-moj-digital, line 31: This is about the only part that has not been copied from CIAG UI (though it is based on their version)
We will need a service method to retrieve an Induction and map it into an `InductionDto`, and we will need it for pretty much every story that we are about to get into re: Updating the Induction; so it made sense to add it now.

We previously has a private method called `getInduction`, so I renamed that one to `retrieveInduction`, and then wrote this version of `getInduction`

server/testsupport/inductionDtoTestDataBuilder.ts
```diff
@@ -0,0 +1,233 @@
+import type { InductionDto } from 'inductionDto'
```
nathanrussell-moj-digital, line 1: Copied from CIAG UI

server/services/inductionService.ts
mattwills-moj-digital: I found this a bit confusing initially. How about putting the error handling in the client and then I think you can get rid of `retrieveInduction()`?

mattwills-moj-digital (approved):
LGTM

PR #428, opened 2024-02-22
RR-647 Update Induction service, mapper and client methods

mattwills-moj-digital:


server/data/mappers/updateInductionMapper.ts
```diff
@@ -0,0 +1,18 @@
...
+    reference: updateInductionDto.reference,
+    prisonId: updateInductionDto.prisonId,
+    workOnRelease: updateInductionDto.workOnRelease,
+    previousQualifications: updateInductionDto.previousQualifications,
+    previousTraining: updateInductionDto.previousTraining,
+    previousWorkExperiences: updateInductionDto.previousWorkExperiences,
+    inPrisonInterests: updateInductionDto.inPrisonInterests,
+    personalSkillsAndInterests: updateInductionDto.personalSkillsAndInterests,
+    futureWorkInterests: updateInductionDto.futureWorkInterests,
+  } as UpdateInductionRequest
```
nathanrussell-moj-digital, line 15: Do you need the cast `as UpdateInductionRequest` on this line? If you do, it might suggest you are missing some fields?

Also, given that it looks like its a direct field to field mapping, would this work (might be too concise for your liking?)

```
return { ...updateInductionDto }
```

nathanrussell-moj-digital:
Ah, you have a linting error (its actually on my branch), but it's going to prevent you merging this I think.
You might have to temporarily remove this unused variable from the `InPrisonWorkUpdateController` 
Screenshot 2024-02-22 at 17 33 46

As discussed earlier, its the `inductionDto` that represents the un-edited Induction, and I put it there because you'll need it in that method to map the changes into it in order to call you service.

You might need to remove it in order to be able to merge this, and then put it back when you are ready to use it

mattwills-moj-digital:
> Ah, you have a linting error (its actually on my branch), but it's going to prevent you merging this I think.

Yeah I spotted that after I pushed. Easy fix 👍

PR #438, opened 2024-02-26
NO_JIRA - fix the build by removing hard coded dates

nathanrussell-moj-digital:
PR to fix a couple of tests that use 26 Feb 2024 as their date values.
These tests started failing today (because today is 26th Feb 2024 !!! 🤣 )

### DIFF
--- a/integration_tests/e2e/goal/createAndEditGoal.cy.ts
+++ b/integration_tests/e2e/goal/createAndEditGoal.cy.ts
@@ -83,12 +83,39 @@ context('Create and edit a goal', () => {
     const reviewPage = Page.verifyOnPage(ReviewPage)
     reviewPage.clickChangeGoalTargetDateLink()
 
+    const today = new Date()
+    const nextYear = new Date(
+      today.getTime() +
+        365 * // days
+          24 * // hours
+          60 * // minutes
+          60 * // seconds
+          1000, // milliseconds
+    )
+    const targetCompletionDay = `${nextYear.getDate()}`
+    const targetCompletionMonth = `${nextYear.getMonth() + 1}` // month is zero indexed so need to add 1
+    const targetCompletionYear = `${nextYear.getFullYear()}`
+    const expectedMonthName = [
+      'January',
+      'February',
+      'March',
+      'April',
+      'May',
+      'June',
+      'July',
+      'August',
+      'September',
+      'October',
+      'November',
+      'December',
+    ][nextYear.getMonth()]
+
     const editGoalPage = Page.verifyOnPage(CreateGoalPage)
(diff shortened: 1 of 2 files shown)


integration_tests/e2e/goal/createAndEditGoal.cy.ts
```diff
@@ -83,12 +83,39 @@ context('Create and edit a goal', () => {
     const reviewPage = Page.verifyOnPage(ReviewPage)
     reviewPage.clickChangeGoalTargetDateLink()
 
+    const today = new Date()
```
nathanrussell-moj-digital, line 86: Cypress test was failing because it was setting the target completion date to 26 Feb 2024 which fails the validation (target completion dates have to be in the future, and today is 26 Feb 2024 and is not in the future! 🤣 )

Changed the test to work out and use "today + 1 year", so this test should not fail in the future 🤞

server/routes/createGoal/createGoalFormValidator.test.ts
```diff
@@ -1,10 +1,16 @@
...
 import validateGoalTitle from '../../validators/goalTitleValidator'
+import goalTargetCompletionDateValidator from '../../validators/goalTargetCompletionDateValidator'
 import validateCreateGoalForm from './createGoalFormValidator'
 
 jest.mock('../../validators/goalTitleValidator')
+jest.mock('../../validators/goalTargetCompletionDateValidator')
+
 describe('createGoalFormValidator', () => {
   const mockedValidateGoalTitle = validateGoalTitle as jest.MockedFunction<typeof validateGoalTitle>
+  const mockedValidateTargetCompletionDate = goalTargetCompletionDateValidator as jest.MockedFunction<
```
nathanrussell-moj-digital, line 11: The `createGoalFormValidator` unit test was failing because we were not mocking it's `goalTargetCompletionDateValidator` dependency. Therefore it was using the real `goalTargetCompletionDateValidator` validator, and with a date in some of the tests of 26 Feb 2024 (ie. today) it started failing.
Changed the test to mock `goalTargetCompletionDateValidator`

Dont worry! We are not testing mocks here! 🤣 
The unit under test `createGoalFormValidator` calls 2 other validators internally (`mockedValidateGoalTitle` and `mockedValidateTargetCompletionDate`) and returns values based on the return values from the 2 dependencies.

The `mockedValidateGoalTitle` was already mocked; we just need to mock `mockedValidateTargetCompletionDate`

The 2 dependant validators (`mockedValidateGoalTitle` and `mockedValidateTargetCompletionDate`) are unit tests properly themselves, so we have correct and adequate cover 👍

mattwills-moj-digital (approved):
GS

PR #442, opened 2024-02-26
RR-648 nunjucks template for the work interests page

chrimesdev:
## Description

Nunjucks template for the work interests page

https://dsdmoj.atlassian.net/browse/RR-648


nathanrussell-moj-digital (approved):
We'll take this and merge it; but this is not the page I was expecting you to do! 🤣 

This page that you've done is Work Interests (ie. what type work are they interested in doing upon release). We will need that page (but not yet 😁 )

The subtask that you were doing was for Work Experience (ie, what work have they done in the past)

Never mind; easy mistake to make as they are very similar pages! 🤣

PR #454, opened 2024-02-28
RR-672 Factors affecting ability to work route and controller etc

mattwills-moj-digital:
Following the view template in my previous PR, this ones adds the required functionality. I'll add cypress tests in my next PR.


server/routes/induction/common/affectAbilityToWorkController.ts
```diff
@@ -0,0 +1,44 @@
...
+import InductionController from './inductionController'
+import AffectAbilityToWorkView from './affectAbilityToWorkView'
+
+/**
+ * Abstract controller class defining functionality common to both the Create and Update Induction journeys.
+ */
+export default abstract class AffectAbilityToWorkController extends InductionController {
+  constructor() {
+    super()
+  }
```
leejacobson-moj, line 13: This shouldn't be needed. 
```suggestion
```

server/routes/induction/update/affectAbilityToWorkUpdateController.ts
```diff
@@ -0,0 +1,82 @@
...
+    next: NextFunction,
+  ): Promise<void> => {
+    const { prisonNumber } = req.params
+    const { prisonerSummary, inductionDto } = req.session
+    const { prisonId } = prisonerSummary
+
+    req.session.affectAbilityToWorkForm = { ...req.body }
+    if (!req.session.affectAbilityToWorkForm.affectAbilityToWork) {
+      req.session.affectAbilityToWorkForm.affectAbilityToWork = []
+    }
```
leejacobson-moj, line 40: If you prefer, you could use a Nullish coalescing assignment here (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment).

```suggestion
  req.session.affectAbilityToWorkForm.affectAbilityToWork ??= []
```

leejacobson-moj (approved):
Looks good to me. Added a couple of suggestions if you feel they're worth looking at.

server/routes/induction/update/affectAbilityToWorkUpdateController.ts
mattwills-moj-digital: That would be more succinct, but I'd like to solicit opinion on that 👍

server/routes/induction/common/affectAbilityToWorkController.ts
mattwills-moj-digital: Yep, removed.

Btw, I was worried that might fail the lint check, so I've just re-committed it manually!

leejacobson-moj (approved):

server/routes/induction/common/affectAbilityToWorkController.ts
```diff
@@ -0,0 +1,40 @@
...
+import type { AffectAbilityToWorkForm } from 'inductionForms'
+import InductionController from './inductionController'
+import AffectAbilityToWorkView from './affectAbilityToWorkView'
+
+/**
+ * Abstract controller class defining functionality common to both the Create and Update Induction journeys.
+ */
+export default abstract class AffectAbilityToWorkController extends InductionController {
+  /**
+   * Returns the Personal Interests view; suitable for use by the Create and Update journeys.
```
nathanrussell-moj-digital, line 12:
```
* Returns the Affects on Ability To Work view; suitable for use ...
```
perhaps?

server/routes/induction/update/affectAbilityToWorkFormValidator.test.ts
```diff
@@ -0,0 +1,84 @@
...
+        // When
+        const actual = validateAffectAbilityToWorkForm(spec, prisonerSummary)
+
+        // Then
+        expect(actual).toEqual(expected)
+      })
+    })
+  })
+
+  //  //  LIMITED_BY_OFFENSE = 'LIMITED_BY_OFFENSE',
```
nathanrussell-moj-digital, line 29: Comments left in by mistake?

server/routes/induction/update/affectAbilityToWorkFormValidator.test.ts
```diff
@@ -0,0 +1,84 @@
...
+      { affectAbilityToWork: ['NO_RIGHT'], affectAbilityToWorkOther: undefined },
+      { affectAbilityToWork: ['NO_RIGHT_TO_WORK', 'a-non-supported-value'], affectAbilityToWorkOther: undefined },
+      { affectAbilityToWork: ['NO_RIGHT_TO_WORK', 'NONE'], affectAbilityToWorkOther: undefined },
+    ).forEach(spec => {
+      it(`form data: ${JSON.stringify(spec)}`, () => {
+        // Given
+        const expected: Array<Record<string, string>> = [
+          {
+            href: '#affectAbilityToWork',
+            text: `Select factors affecting Jimmy Lightfingers's ability to work or select 'None of these'`,
```
nathanrussell-moj-digital, line 50: One day we'll fix this (but not in these PRs)

ie. the incorrect placement of the apostrophe when surnames end in an `s` -  in this case it shouldn't it be `Jimmy Lightfingers' ability to work` ? 😁 :trollface:

server/routes/induction/update/affectAbilityToWorkUpdateController.ts
nathanrussell-moj-digital: I think if we do it, then yes, Matt is right, we should do it in the other controllers we are writing at the moment (they all follow this pattern).
But also we should also do it in conjunction with the next line, as they basically work together. As it stands the code is this:

```
            if (!req.session.affectAbilityToWorkForm.affectAbilityToWork) {
	      req.session.affectAbilityToWorkForm.affectAbilityToWork = []
	    }
            if (!Array.isArray(req.session.affectAbilityToWorkForm.affectAbilityToWork)) {
              req.session.affectAbilityToWorkForm.affectAbilityToWork = [
                 req.session.affectAbilityToWorkForm.affectAbilityToWork,
              ]
            }
```
The purpose of these 2 if statements is to make sure the form object in the session is "correct" before we start. Specifically we are looking to "correct" up the array field - `affectAbilityToWork` in this case.

This is to make our validators simpler (otherwise they would have to do this type of thing), but getting the array in order beforehand will also help our mapper code later on.

The problem is to do with how the browser submits the form with checkbox fields, and how it is deserialized into the form object.
The html form binds the checkboxes to the input field name `affectAbilityToWork`
If the user selects and the browser submits more than one checkbox value, then the form submission contains many `affectAbilityToWork` fields which are deserialized into an array - happy days
If the user does not select any checkboxes at all, the browser does not submit the field at all, and hence the property is null. This is the first `if` statement.
If the user only selects one checkbox, the browser submits it .... but only a single `affectAbilityToWork` field, which gets deserialized into a string property, not an array property. This is what the 2nd if statement is fixing.

So, more than happy to do something with the current pattern (though I would argue that as it currently stands its more readable than other approaches?), but we should do it in conjunction with the 2nd if statement as well. Perhaps refactor it out to a private function, or better still, refactor it out into a single  generic function somewhere so that it works with any form type and property and we can use it from all the controllers?

nathanrussell-moj-digital (approved):
Some minor comments, but LGTM 👍

server/routes/induction/update/affectAbilityToWorkUpdateController.ts
mattwills-moj-digital: I wasn't sure about the readability either, but that might be because of my lack of familiarity in this area and obviously it's always good to challenge that. Regardless, I'll leave it as it is for now and we can take a view on it separately.

server/routes/induction/update/affectAbilityToWorkFormValidator.test.ts
mattwills-moj-digital: Controversial! I think both are correct actually: https://www.wikihow.com/James-or-James%27s

server/routes/induction/update/affectAbilityToWorkFormValidator.test.ts
mattwills-moj-digital: OMG, I'm so disappointed with myself! 😱

mattwills-moj-digital:
I've addressed the minor comments in my next PR: https://github.com/ministryofjustice/hmpps-education-and-work-plan-ui/pull/455

PR #458, opened 2024-02-29
RR-646 - Cypress tests for updating Previous Work Experience Detail

nathanrussell-moj-digital:
This PR adds the cypress tests for updating the Previous Work Experience Detail on an Induction

### DIFF
--- a/integration_tests/e2e/induction/updatePreviousWorkExperienceDetail.cy.ts
+++ b/integration_tests/e2e/induction/updatePreviousWorkExperienceDetail.cy.ts
@@ -1,0 +1,105 @@
+import Page from '../../pages/page'
+import WorkAndInterestsPage from '../../pages/overview/WorkAndInterestsPage'
+import { putRequestedFor } from '../../mockApis/wiremock/requestPatternBuilder'
+import { urlEqualTo } from '../../mockApis/wiremock/matchers/url'
+import { matchingJsonPath } from '../../mockApis/wiremock/matchers/content'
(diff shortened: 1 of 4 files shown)


integration_tests/e2e/induction/updatePreviousWorkExperienceDetail.cy.ts
```diff
@@ -0,0 +1,105 @@
...
+    cy.task('stubUpdateInduction')
+    cy.task('stubGetInductionLongQuestionSet')
+    cy.signIn()
+  })
+
+  it('should update Previous Work Experience Detail given page submitted with changed job role and detail and no validation errors', () => {
+    // Given
+    const prisonNumber = 'G6115VJ'
+
+    // stubGetInductionLongQuestionSet has previous work experiences of Office and Other
```
mattwills-moj-digital, line 32: Very minor typo (which we can address later if you prefer): should read `Office manager`, not `Office`

mattwills-moj-digital (approved):
AToAB

integration_tests/e2e/induction/updatePreviousWorkExperienceDetail.cy.ts
nathanrussell-moj-digital: No probs, I'll address it in my next PR 👍 , but its actually meant to be describing the list of types, rather than any specific role (because the test is based around a work experience type, by virtue of the path parameter), so I'll change it to:

```
// stubGetInductionLongQuestionSet has previous work experiences whose types are OFFICE and OTHER
```

PR #459, opened 2024-02-29
RR-668 Controller etc for Reasons Not To Get Work page

mattwills-moj-digital:


server/@types/dto/index.d.ts
```diff
@@ -60,7 +60,7 @@ declare module 'inductionDto' {
 
   export interface WorkOnReleaseDto extends ReferencedAndAuditable {
     hopingToWork: HopingToGetWorkValue
-    notHopingToWorkReasons: ReasonToNotGetWorkValue[]
+    notHopingToWorkReasons: ReasonNotToGetWorkValue[]
```
mattwills-moj-digital, line 63: Previously we had a mix of `reason(s)NotTo` and `reason(s)ToNot`, but I'm consistently using the former (except for `notHopingToWorkReasons` here, as that models the API, rather than the UI codebase).

PR #461, opened 2024-03-01
RR-680 Route, controller and validator for Future Work Interest Types

mattwills-moj-digital:
This PR adds the controller etc for the future work interest "types" page (which is distinct from a separate page for providing "details" of the work type):

image

I'll add cypress tests in a separate PR, but this appears to work ok on my local machine.

PR #462, opened 2024-03-04
RR-646 - Initial controller for Previous Work Experience Types

nathanrussell-moj-digital:
This PR is the first part of the Previous Work Experience Types controller.

In general it follows the same approach that we've been using so far for the Induction Update screens, but the POST (submit) is more complex because there are a number of scenarios:

* Form submitted with no changes at all
This is addressed in this PR
* Form submitted only removals to Previous Work Experience (ie. checkbox(es) unticked, but no other changes)
This is addressed in this PR
* Form submitted with other changes (ie. zero or more removals AND (one or more additions OR the value for `OTHER` was changed)
In this scenario we need to present the Previous Work Experience detail page for each Work Experience type
This is partially implemented in this PR, in that we work out the list of Previous Work Experiences that we need to show the detail page for, but not actually go there.

The reason we do not go to the Detail pages in this PR is that I want to keep the PRs as small as possible and add functionality iteratively. This is particularly true in this case as I suspect the page "flow" / form handling will be quite complicated and a massive PR will mask this complexity leading to a difficult review process.

### DIFF
--- a/server/routes/induction/update/previousWorkExperienceTypesUpdateController.test.ts
+++ b/server/routes/induction/update/previousWorkExperienceTypesUpdateController.test.ts
@@ -1,0 +1,310 @@
+import { NextFunction, Request, Response } from 'express'
+import type { SessionData } from 'express-session'
+import type { PreviousWorkExperienceDto } from 'inductionDto'
+import { InductionService } from '../../../services'
+import PreviousWorkExperienceTypesUpdateController from './previousWorkExperienceTypesUpdateController'
+import validatePreviousWorkExperienceTypesForm from './previousWorkExperienceTypesFormValidator'
+import toCreateOrUpdateInductionDto from '../../../data/mappers/createOrUpdateInductionDtoMapper'
+import aValidPrisonerSummary from '../../../testsupport/prisonerSummaryTestDataBuilder'
+import { aLongQuestionSetInductionDto } from '../../../testsupport/inductionDtoTestDataBuilder'
+import { aLongQuestionSetUpdateInductionDto } from '../../../testsupport/updateInductionDtoTestDataBuilder'
+
+jest.mock('./previousWorkExperienceTypesFormValidator')
+jest.mock('../../../data/mappers/createOrUpdateInductionDtoMapper')
+
+describe('previousWorkExperienceTypesUpdateController', () => {
+  const mockedFormValidator = validatePreviousWorkExperienceTypesForm as jest.MockedFunction<
+    typeof validatePreviousWorkExperienceTypesForm
+  >
+  const mockedCreateOrUpdateInductionDtoMapper = toCreateOrUpdateInductionDto as jest.MockedFunction<
+    typeof toCreateOrUpdateInductionDto
+  >
+
+  const inductionService = {
+    updateInduction: jest.fn(),
+  }
+
+  const controller = new PreviousWorkExperienceTypesUpdateController(inductionService as unknown as InductionService)
+
+  const req = {
+    session: {} as SessionData,
+    body: {},
+    user: {} as Express.User,
+    params: {} as Record<string, string>,
+    flash: jest.fn(),
+  }
+  const res = {
+    redirect: jest.fn(),
+    render: jest.fn(),
+  }
+  const next = jest.fn()
+
+  let errors: Array<Record<string, string>>
+
+  beforeEach(() => {
+    jest.resetAllMocks()
+    req.session = {} as SessionData
+    req.body = {}
+    req.user = {} as Express.User
+    req.params = {} as Record<string, string>
+
+    errors = []
+  })
+
+  describe('getPreviousWorkExperienceTypesView', () => {
+    it('should get the Previous Work Experience Types view given there is no PreviousWorkExperienceTypesForm on the session', async () => {
+      // Given
+      const prisonNumber = 'A1234BC'
+      req.params.prisonNumber = prisonNumber
+
+      const prisonerSummary = aValidPrisonerSummary()
+      req.session.prisonerSummary = prisonerSummary
+      const inductionDto = aLongQuestionSetInductionDto({ hasWorkedBefore: true })
+      req.session.inductionDto = inductionDto
+      req.session.previousWorkExperienceTypesForm = undefined
+
+      const expectedPreviousWorkExperienceTypesForm = {
+        typeOfWorkExperience: ['CONSTRUCTION', 'OTHER'],
+        typeOfWorkExperienceOther: 'Retail delivery',
+      }
+
+      const expectedView = {
+        prisonerSummary,
+        form: expectedPreviousWorkExperienceTypesForm,
+        backLinkUrl: '/plan/A1234BC/view/work-and-interests',
+        backLinkAriaText: 'Back to <TODO - check what CIAG UI does here>',
+        errors,
+      }
+
+      // When
+      await controller.getPreviousWorkExperienceTypesView(
+        req as undefined as Request,
+        res as undefined as Response,
+        next as undefined as NextFunction,
+      )
+
+      // Then
+      expect(res.render).toHaveBeenCalledWith(
+        'pages/induction/previousWorkExperience/workExperienceTypes',
+        expectedView,
+      )
+      expect(req.session.previousWorkExperienceTypesForm).toBeUndefined()
+      expect(req.session.inductionDto).toEqual(inductionDto)
+    })
+
+    it('should get the Previous Work Experience Types view given there is an PreviousWorkExperienceTypesForm already on the session', async () => {
+      // Given
+      const prisonNumber = 'A1234BC'
+      req.params.prisonNumber = prisonNumber
+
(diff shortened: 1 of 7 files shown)


server/routes/induction/update/previousWorkExperienceTypesUpdateController.ts
```diff
@@ -0,0 +1,263 @@
...
+        req.session.previousWorkExperienceTypesForm.typeOfWorkExperience,
+      ]
+    }
+    const { previousWorkExperienceTypesForm } = req.session
+
+    const errors = validatePreviousWorkExperienceTypesForm(previousWorkExperienceTypesForm, prisonerSummary)
+    if (errors.length > 0) {
+      req.flash('errors', errors)
+      return res.redirect(`/prisoners/${prisonNumber}/induction/previous-work-experience`)
+    }
```
nathanrussell-moj-digital, line 51: Up to this point, the submit handler of this controller should be no different to any of the others we are doing in this epic. IE. get the form from the request body, fix up any array fields, and validate the form 👍 

The complexity comes next ! 🤣

server/routes/induction/update/previousWorkExperienceTypesUpdateController.ts
```diff
@@ -0,0 +1,251 @@
...
+          return next(createError(500, `Error updating Induction for prisoner ${prisonNumber}. Error: ${e}`))
+        }
+      } else {
+        logger.debug('No changes to Previous Work Experiences were submitted')
+      }
+
+      req.session.previousWorkExperienceTypesForm = undefined
+      req.session.inductionDto = undefined
+      return res.redirect(`/plan/${prisonNumber}/view/work-and-interests`)
+    }
```
nathanrussell-moj-digital, line 74: The PR adds the unit test coverage for everything up until here (including this `if` block for "no changes" and "only removals")

The following section (which is not complete) is not covered, but will be tested in my next PR when we introduce the "mini-flow"

server/routes/induction/update/previousWorkExperienceTypesUpdateController.test.ts
```diff
@@ -0,0 +1,310 @@
...
+      expect(mockedCreateOrUpdateInductionDtoMapper).toHaveBeenCalledWith(prisonerSummary.prisonId, updatedInduction)
+      expect(updatedInduction.previousWorkExperiences.experiences).toEqual(expectedPreviousWorkExperiences)
+
+      expect(inductionService.updateInduction).toHaveBeenCalledWith(prisonNumber, updateInductionDto, 'some-token')
+      expect(res.redirect).toHaveBeenCalledWith(`/plan/${prisonNumber}/view/work-and-interests`)
+      expect(req.session.skillsForm).toBeUndefined()
+      expect(req.session.inductionDto).toBeUndefined()
+    })
+
+    it('should update Induction given form is submitted where the only change is a removal of the work type OTHER', async () => {
```
mattwills-moj-digital, line 259: Just wondering why you're differentiating between removing OTHER and non OTHER? I could understand editing OTHER, but not removing?

server/routes/induction/update/previousWorkExperienceTypesUpdateController.ts
```diff
@@ -0,0 +1,251 @@
...
+      return res.redirect(`/prisoners/${prisonNumber}/induction/previous-work-experience`)
+    }
+
+    if (
+      noChangesToPreviousWorkExperienceSubmitted(inductionDto, previousWorkExperienceTypesForm) ||
+      onlyRemovalsWereMadeToPreviousWorkExperienceSubmitted(inductionDto, previousWorkExperienceTypesForm)
+    ) {
+      if (onlyRemovalsWereMadeToPreviousWorkExperienceSubmitted(inductionDto, previousWorkExperienceTypesForm)) {
+        logger.debug(
+          'Previous Work Experiences changes are only removed types so Induction can be updated without starting a flow',
```
mattwills-moj-digital, line 59: Minor `...without asking for work experience details` might more sense to the new reader?

server/routes/induction/update/previousWorkExperienceTypesUpdateController.test.ts
nathanrussell-moj-digital: That is a good point; I think it was from my original thought process (which we spoke about on Friday) .... I'll have a quick think about renaming it (or removing it completely) .... 🤔

server/routes/induction/update/previousWorkExperienceTypesUpdateController.ts
nathanrussell-moj-digital: Yes, 💯 , will do that now 👍

server/routes/induction/update/previousWorkExperienceTypesUpdateController.ts
```diff
@@ -0,0 +1,251 @@
...
+        logger.debug('No changes to Previous Work Experiences were submitted')
+      }
+
+      req.session.previousWorkExperienceTypesForm = undefined
+      req.session.inductionDto = undefined
+      return res.redirect(`/plan/${prisonNumber}/view/work-and-interests`)
+    }
+
+    /* We need to show the Details page for each of:
+         - previous work experience on the original induction that has not been removed by the form submission
```
mattwills-moj-digital, line 77: Deciding when to show the details page is probably up for debate, but I guess you don't want to ask for guidance from our illustrious design/UR team?!

I'm conscious I suggested we ask for the details of each job type (new and existing) when they make changes (i.e. whey they don't just remove some). I think is the right decision, but there is an inherent inconsistency here...

What happens if they want to remove one and edit the job role of another? I assume they'd have to remove one from the list and then go back to the "Work and interests" page, before clicking on the change link for "Specific job role of interest". That might be fine, but if they added a job type, then they'd also be able to edit the existing job roles in the same flow. Does that make sense?

server/routes/induction/update/previousWorkExperienceTypesUpdateController.ts
```diff
@@ -0,0 +1,251 @@
...
+         - any additional job types that have been added by the form submission
+         - and OTHER if it's value has changed
+     */
+    const workExperienceTypesToShowDetailsFormFor = [
+      ...typesOfPreviousWorkExperienceToShowDetailsFormFor(inductionDto, previousWorkExperienceTypesForm),
+    ].sort(previousWorkExperienceTypeScreenOrderComparator) // sort them by the order presented on screen (which is not alphabetic on the enum values)
```
mattwills-moj-digital, line 83: Makes sense 👍

server/routes/induction/update/previousWorkExperienceTypesUpdateController.ts
nathanrussell-moj-digital: > What happens if they want to remove one and edit the job role of another? 
They would do the "edit" on the Change link of the job type in question, and not via this change link (which is for the list of job types)

I think the complexity is actually because you talked me into changing the process slightly in this respect, where the "queue" of detail pages is to include all job types on the induction, even those that you've not changed here

Might be easier to chat through on a call .. 🤔

server/routes/induction/update/previousWorkExperienceTypesUpdateController.ts
```diff
@@ -0,0 +1,251 @@
...
+   * Updates the Induction using the InductionService.
+   * @throws Error if InductionService throws an error. The Error from InductionService is rethrown.
+   */
+  private updateInduction = async (
+    inductionDto: InductionDto,
+    previousWorkExperienceTypesForm: PreviousWorkExperienceTypesForm,
+    req: Request,
+    prisonerSummary: PrisonerSummary,
+  ) => {
+    const updatedInduction = updatedInductionDtoWithRemovedPreviousWorkExperiencesAndChangesToOther(
```
mattwills-moj-digital, line 103: Nice descriptive method name! 😂 😉 

In all seriousness, I'm wondering if this a bit too involved and could be simplified? In terms of work experience types, the submitted form is the "final source of truth". It just lacks the job details. Would it not be simpler/easier to just build up a list of job types and extract the missing roles from the existing ones in the session?

I did something similar in my PR: https://github.com/ministryofjustice/hmpps-education-and-work-plan-ui/pull/461/files#r1510958161

server/routes/induction/update/previousWorkExperienceTypesUpdateController.test.ts
nathanrussell-moj-digital: Removed the test 👍

server/routes/induction/update/previousWorkExperienceTypesUpdateController.ts
nathanrussell-moj-digital: Done 👍

server/routes/induction/update/previousWorkExperienceTypesUpdateController.ts
nathanrussell-moj-digital: Not sure I follow ... lets chat... 👍

mattwills-moj-digital (approved):
LGTM

PR #466, opened 2024-03-05
RR-646 - Refactored induction mapping and session handling to support being able to redirect to Details page

nathanrussell-moj-digital:
This PR refactors the induction mapping and session handling to support being able to redirect to Details page

### DIFF
--- a/server/routes/induction/common/previousWorkExperienceDetailController.ts
+++ b/server/routes/induction/common/previousWorkExperienceDetailController.ts
@@ -89,11 +89,11 @@ const toPreviousWorkExperienceDetailForm = (
   inductionDto: InductionDto,
   typeOfWorkExperience: TypeOfWorkExperienceValue,
 ): PreviousWorkExperienceDetailForm => {
-  const previousWorkExperience = inductionDto.previousWorkExperiences.experiences.find(
+  const previousWorkExperience = inductionDto.previousWorkExperiences?.experiences.find(
     experience => experience.experienceType === typeOfWorkExperience,
   )
   return {
-    jobRole: previousWorkExperience.role,
-    jobDetails: previousWorkExperience.details,
+    jobRole: previousWorkExperience?.role || '',
+    jobDetails: previousWorkExperience?.details || '',
   }
 }
--- a/server/routes/induction/update/previousWorkExperienceDetailUpdateController.test.ts
+++ b/server/routes/induction/update/previousWorkExperienceDetailUpdateController.test.ts
@@ -92,7 +92,7 @@ describe('previousWorkExperienceDetailUpdateController', () => {
         'pages/induction/previousWorkExperience/workExperienceDetail',
         expectedView,
       )
-      expect(req.session.personalInterestsForm).toBeUndefined()
+      expect(req.session.previousWorkExperienceDetailForm).toBeUndefined()
       expect(req.session.inductionDto).toEqual(inductionDto)
     })
 
@@ -135,7 +135,7 @@ describe('previousWorkExperienceDetailUpdateController', () => {
         'pages/induction/previousWorkExperience/workExperienceDetail',
         expectedView,
       )
-      expect(req.session.personalInterestsForm).toBeUndefined()
+      expect(req.session.previousWorkExperienceDetailForm).toBeUndefined()
       expect(req.session.inductionDto).toEqual(inductionDto)
     })
 
--- a/server/routes/induction/update/previousWorkExperienceTypesUpdateController.ts
+++ b/server/routes/induction/update/previousWorkExperienceTypesUpdateController.ts
@@ -2,7 +2,6 @@ import createError from 'http-errors'
 import { NextFunction, Request, RequestHandler, Response } from 'express'
 import type { InductionDto, PreviousWorkExperienceDto } from 'inductionDto'
 import type { PreviousWorkExperienceTypesForm } from 'inductionForms'
-import type { PrisonerSummary } from 'viewModels'
 import logger from '../../../../logger'
 import PreviousWorkExperienceTypesController from '../common/previousWorkExperienceTypesController'
 import { InductionService } from '../../../services'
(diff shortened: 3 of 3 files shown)


server/routes/induction/common/previousWorkExperienceDetailController.ts
```diff
@@ -89,11 +89,11 @@ const toPreviousWorkExperienceDetailForm = (
   inductionDto: InductionDto,
   typeOfWorkExperience: TypeOfWorkExperienceValue,
 ): PreviousWorkExperienceDetailForm => {
-  const previousWorkExperience = inductionDto.previousWorkExperiences.experiences.find(
+  const previousWorkExperience = inductionDto.previousWorkExperiences?.experiences.find(
```
nathanrussell-moj-digital, line 92: Belt'n'braces - just makes sure we dont NPE or equivalent when trying to render the page where there is no job role/details for the given work experience type

server/routes/induction/update/previousWorkExperienceDetailUpdateController.test.ts
```diff
@@ -92,7 +92,7 @@ describe('previousWorkExperienceDetailUpdateController', () => {
         'pages/induction/previousWorkExperience/workExperienceDetail',
         expectedView,
       )
-      expect(req.session.personalInterestsForm).toBeUndefined()
+      expect(req.session.previousWorkExperienceDetailForm).toBeUndefined()
```
nathanrussell-moj-digital, line 95: Fixing a copy and paste typo from an earlier PR! 🤣

server/routes/induction/update/previousWorkExperienceTypesUpdateController.ts
```diff
@@ -58,14 +57,20 @@ export default class PreviousWorkExperienceTypesUpdateController extends Previou
         logger.debug(
           'Previous Work Experiences changes are only removed types so Induction can be updated without asking for work experience details',
         )
-        // call the service to update the Induction
+
+        // create an updated InductionDto with any changes to Previous Work Experiences and then map it to a CreateOrUpdateInductionDTO to call the API
```
nathanrussell-moj-digital, line 61: Change the way the Previous Work Experience **Types** controller works re: submit handler....

Here, we are still only updating the Induction when there are only removals, but we are now doing it in the exact same way we do with all our other Update controllers; in that we map to the Induction (I've change the mapper method, see below), then map it to the CreateOrUpdateDto, then call the service.

Previously the updating of the Induction via the service call was in a private method because I'd envisaged calling it from different places for different circumstances (ie. for removals only, and for additions, and for changes to Other etc). This is now not the case, and I think the only time this submit hander will actually call the API to update the Induction is here when the only change is removals 👍

server/routes/induction/update/previousWorkExperienceTypesUpdateController.ts
```diff
@@ -58,14 +57,20 @@ export default class PreviousWorkExperienceTypesUpdateController extends Previou
...
+        const updatedInduction = updatedInductionDtoWithPreviousWorkExperiences(
```
nathanrussell-moj-digital, line 62: I've renamed (and refactored) the mapper method from `updatedInductionDtoWithRemovalsAndChangesToOther` to `updatedInductionDtoWithPreviousWorkExperiences`
The new name is consistent with similar mapper methods in the other Update controller classes, and it now no longer only handles removes and changes to Other .... it also handles additions 😁 , and we use it a few lines down in respect of the "mini-flow" where new work experience types have been added

server/routes/induction/update/previousWorkExperienceTypesUpdateController.ts
```diff
@@ -89,38 +94,30 @@ export default class PreviousWorkExperienceTypesUpdateController extends Previou
       `Previous Work Experiences changes resulting in going to the Detail pages for ${workExperienceTypesToShowDetailsFormFor}`,
     )
 
-    // TODO - replace with correct redirect once we've implemented form redirect/handling for a "queue" of Previous Work Experience Detail pages
-    return res.redirect(`/plan/${prisonNumber}/view/work-and-interests`)
-  }
+    // Update the InductionDTO in the session with changes to Previous Work Experiences, but do not persist to the API
+    // The user will be redirected to each Previous Work Experience Detail page in turn
+    req.session.inductionDto = updatedInductionDtoWithPreviousWorkExperiences(
```
nathanrussell-moj-digital, line 99: This updates the Induction DTO on the session by calling the `updatedInductionDtoWithPreviousWorkExperiences` method, which adds in any new work experience types.

At the moment we are only redirecting to the first new job type detail page, but I will change that in my next PR (to iterate through them all).
I just wanted to get this first part done to prove whether it will work or not ... and I think it does 👍 😁 🎉

server/routes/induction/update/previousWorkExperienceTypesUpdateController.ts
mattwills-moj-digital: So if they remove and add some experience types, then the changes won't be persisted until they've provided details for the first work experience on the subsequent screen?

server/routes/induction/update/previousWorkExperienceTypesUpdateController.ts
mattwills-moj-digital: The word `previous` caught me out here. I initially thought it was work experiences that were submitted previously! 😂

server/routes/induction/update/previousWorkExperienceTypesUpdateController.ts
```diff
@@ -183,7 +190,7 @@ const addedPreviousWorkExperienceTypes = (
 
 /**
  * Given the current Induction and the [PreviousWorkExperienceTypesForm], returns `true` if both the Induction and the
- * [PreviousWorkExperienceTypesForm] inclyde a Previous Work Experience type of OTHER, and the it's value is different.
+ * [PreviousWorkExperienceTypesForm] inclyde a Previous Work Experience type of OTHER, and if it's value is different.
```
mattwills-moj-digital, line 193: There are a couple more typos - one of them involves an apostrophe! 😉

mattwills-moj-digital (dismissed):
This looks good and might work perfectly ok, but I can't be sure until we test the flow 👍

server/routes/induction/update/previousWorkExperienceTypesUpdateController.ts
nathanrussell-moj-digital: > So if they remove and add some experience types, then the changes won't be persisted until they've provided details for the first work experience on the subsequent screen?

Not sure yet ..... at the moment I'll tentatively say Yes .... but I'm not sure it works in all cases, and I want to bounce it with you after lunch.

It doesnt change/impact the changes I'm proposing in this PR though. I think we need these changes regardless

server/routes/induction/update/previousWorkExperienceTypesUpdateController.ts
nathanrussell-moj-digital: Done

nathanrussell-moj-digital:
> This looks good and might work perfectly ok, but I can't be sure until we test the flow 👍

It works for the most part. There is a bit that I'm not happy with that I wanted to show you after lunch, but I don't believe the changes in this PR are affecting it. I believe we need these changes regardless

mattwills-moj-digital (approved):

PR #467, opened 2024-03-05
RR-682 add isAccredited to the viewModel / mapper for InPrisonEducation

chrimesdev:
## Description

Add `isAccredited` to the viewModel for InPrisonEducation, which is required for the new views. The mapper for this data will be done in the next PR.

Also added it to the mapper to prevent error due to adding to the ViewModel.

https://dsdmoj.atlassian.net/browse/RR-683


nathanrussell-moj-digital (approved):
LGTM 👍

PR #478, opened 2024-03-11
RR-706 - Replaced multiline text filter with css

nathanrussell-moj-digital:
This PR replaces the use of the `formatMultilineText` and `safe` filters with a pure css solution.

The problem with using `formatMultilineText` is that it replaces `\n` with `<br />`. Because the text now contains html that needs to be rendered we needed to use the `safe` filter. And because we are using the `safe` filter on user entered text (eg: Goal title), it means that any html that the user has entered into the field is rendered as markup, rather than being escaped. This is essentially a XSS vulnerability.

Because of how the styles are inherited and cascaded through the elements, I've had to replace a couple of GDS components with their handwritten html equivalent so that I could apply the style on the correct element. (Otherwise it was being applied to a parent/grandparent element, which meant the spacing was applied incorrectly)

### DIFF
--- a/assets/scss/local.scss
+++ b/assets/scss/local.scss
@@ -26,3 +26,7 @@
 .app-u-section-break-8 {
   border-top: 8px solid $govuk-border-colour !important;
 }
+
+.app-u-multiline-text {
+  white-space: pre-line;
+}
--- a/server/views/pages/goal/review/index.njk
+++ b/server/views/pages/goal/review/index.njk
@@ -28,7 +28,7 @@
                 Description
               </dt>
               <dd class="govuk-summary-list__value" data-qa="goal-{{ goalLoop.index }}-description-value">
-                {{ goal.title | formatMultilineText | safe }}
+                <span class="app-u-multiline-text">{{ goal.title }}</span>
               </dd>
               <dd class="govuk-summary-list__actions">
                 <a class="govuk-link" href="/plan/{{ prisonerSummary.prisonNumber }}/goals/{{ goalLoop.index }}/create?mode=edit" data-qa="change-goal-{{ goalLoop.index }}-description">
@@ -55,7 +55,7 @@
                   Step {{ step.sequenceNumber }}
                 </dt>
                 <dd class="govuk-summary-list__value" data-qa="goal-{{ goalLoop.index }}-step-{{ step.sequenceNumber }}-value">
-                  {{ step.title | formatMultilineText | safe }}
+                  <span class="app-u-multiline-text">{{ step.title }}</span>
                 </dd>
                 <dd class="govuk-summary-list__actions">
                   <a class="govuk-link" href="/plan/{{ prisonerSummary.prisonNumber }}/goals/{{ goalLoop.index }}/add-step/{{ step.sequenceNumber }}?mode=edit" data-qa="change-goal-{{ goalLoop.index }}-step-{{ step.sequenceNumber }}">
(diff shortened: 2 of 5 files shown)


server/views/pages/goal/review/index.njk
```diff
@@ -28,7 +28,7 @@
                 Description
               </dt>
               <dd class="govuk-summary-list__value" data-qa="goal-{{ goalLoop.index }}-description-value">
-                {{ goal.title | formatMultilineText | safe }}
+                <span class="app-u-multiline-text">{{ goal.title }}</span>
```
nathanrussell-moj-digital, line 31: To be sure this works consistently, I've wrapped the text in a `<span>` element that has the desired style. This is so that I can force the start of the text to be immediately after the opening tag that contains the style.

EG: this renders extra line spaces above the text because there is a load of white space between the `<dd>` that contains the class, and the text:
```
              <dd class="govuk-summary-list__value app-u-multiline-text" data-qa="goal-{{ goalLoop.index }}-description-value">{{ goal.title }}</dd>
```

I could have "fixed" it by putting it all on one line like this, but it felt somewhat "hidden" and would be too easy for a dev in the future to do it on several lines (as we have it currently)
```
              <dd class="govuk-summary-list__value app-u-multiline-text" data-qa="goal-{{ goalLoop.index }}-description-value">
                {{ goal.title }}
              </dd>
```

server/views/pages/goal/update/review.njk
```diff
@@ -34,43 +34,44 @@
...
+            {{ goalStatusHtml | safe }}
+          </dd>
+        </div>
+
+        <div class="govuk-summary-list__row">
+          <dt class="govuk-summary-list__key">
+            Description
+          </dt>
+          <dd class="govuk-summary-list__value">
+            <span class="app-u-multiline-text">{{ data.title }}</span>
```
nathanrussell-moj-digital, line 52: To ensure the class is placed on the element immediately wrapping the text, and to ensure it's all on one line to prevent additional newline whitespace being rendered (see comment above) , I had to replace the use of the `govukSummaryList` with it's hand written html equivalent. 

There are 2 places in this component where I needed to do this, here on line 52, and the `notes` field on line 70

server/views/pages/overview/partials/overviewTab/_goalsList.njk
```diff
@@ -56,14 +56,16 @@
...
       {% if goal.note %}
-        {{ govukDetails({
-          summaryHtml: "View note<span class='govuk-visually-hidden'> for Goal " + goal.sequenceNumber + "</span>",
-          text: goal.note | formatMultilineText | safe,
-          attributes: {
-            "data-qa": "overview-notes-expander-" + goal.sequenceNumber
-          },
-          classes: "app-notes-expander"
-        }) }}
+        <details class="govuk-details app-notes-expander" data-qa="overview-notes-expander-{{ goal.sequenceNumber }}">
```
nathanrussell-moj-digital, line 59: As noted above, I've had to replace the GDS component `govukDetails` with its handwritten html equivalent for the same reason (ie. ensuring the `note` field on line 66 is correctly wrapped in an element containing the class)

assets/scss/local.scss
```diff
@@ -26,3 +26,7 @@
 .app-u-section-break-8 {
   border-top: 8px solid $govuk-border-colour !important;
 }
+
+.app-u-multiline-text {
```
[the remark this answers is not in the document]
chrimesdev: Yep I would class this as a utility class (which can be reused) so `app-u` is the correct naming convention!

server/views/pages/goal/review/index.njk
chrimesdev: Makes sense - tested and works well!

chrimesdev (approved):

PR #484, opened 2024-03-12
RR-686 - Cypress tests for courses & qualifications

chrimesdev:
## Description

Update Cypress tests for Work & Education tab; courses and qualifications

https://dsdmoj.atlassian.net/browse/RR-686


integration_tests/e2e/overview/educationAndTrainingTab.cy.ts
```diff
@@ -38,7 +38,47 @@ context('Prisoner Overview page - Education And Training tab', () => {
       educationAndTrainingPage //
         .activeTabIs('Education and training')
         .hasFunctionalSkillsDisplayed()
-        .hasCompletedInPrisonQualificationsDisplayed()
+        .hasCompletedInPrisonQualificationsLast12MonthsDisplayed()
```
nathanrussell-moj-digital, line 41: I don't think I 100% follow this ....
How does this test pass? What stub is the test setting up that makes it display the "courses completed in the last 12 months" table? 
(Same question for the other tests where you've added the assertion `.hasCompletedInPrisonQualificationsLast12MonthsDisplayed()` )

integration_tests/e2e/overview/educationAndTrainingTab.cy.ts
```diff
@@ -38,7 +38,47 @@ context('Prisoner Overview page - Education And Training tab', () => {
...
+      const overviewPage = Page.verifyOnPage(OverviewPage)
+
+      // When
+      overviewPage.selectTab('Education and training')
+      const educationAndTrainingPage = Page.verifyOnPage(EducationAndTrainingPage)
+
+      // Then
+      educationAndTrainingPage //
+        .activeTabIs('Education and training')
+        .hasNoCoursesAndQualificationsLast12MonthsMessageDisplayed()
```
nathanrussell-moj-digital, line 81: Again, I don't follow this - where is the call to a stub in this test that says the prisoner has no completed courses in the last 12 months?
(also, "no completed courses in the last 12 months" is subtly different to "no courses at all" - do we have that covered?)

integration_tests/e2e/overview/educationAndTrainingTab.cy.ts
```diff
@@ -78,7 +118,7 @@ context('Prisoner Overview page - Education And Training tab', () => {
       educationAndTrainingPage //
         .activeTabIs('Education and training')
         .hasFunctionalSkillsDisplayed()
-        .hasCompletedInPrisonQualificationsDisplayed()
+        .hasCompletedInPrisonQualificationsLast12MonthsDisplayed()
```
nathanrussell-moj-digital, line 121: This is definitely wrong - if this test is "the API call to learner education returned a 404" .... how does it manage to display the table that describes the prisoners completed courses in the last 12 months?

integration_tests/e2e/overview/educationAndTrainingTab.cy.ts
```diff
@@ -99,7 +139,7 @@ context('Prisoner Overview page - Education And Training tab', () => {
         .activeTabIs('Education and training')
         .doesNotHaveFunctionalSkillsDisplayed()
         .hasCuriousUnavailableMessageDisplayed()
-        .hasCompletedInPrisonQualificationsDisplayed()
+        .hasCompletedInPrisonQualificationsLast12MonthsDisplayed()
```
nathanrussell-moj-digital, line 142: Likewise - if the Curious learner profile API as returned a 500 (which is what this test is), how are we able to display the table showing courses completed in the last 12 months?
Why are we not displaying the "curious is unavailable message" in place of the completed courses table?

integration_tests/e2e/overview/educationAndTrainingTab.cy.ts
chrimesdev: I've only updated the function name here to be more accurate description as to what its looking for on the page; so the test must have always been wrong. I can look at rewriting the tests to work.

integration_tests/mockApis/curiousApi.ts
```diff
@@ -316,6 +317,125 @@ const stubLearnerEducation = (prisonNumber = 'G6115VJ', page = 0): SuperAgentReq
     },
   })
 
+const stubLearnerEducationWithCoursesQualificationsInLast12Months = (
```
nathanrussell-moj-digital, line 320: I appreciate this could make for a very long method name, but I think the key word/phrase "completed" is missing here.
The important thing about this stub is that it is returning details of learner education (courses) that have been completed in the last 12 months.

I wonder if we can get this into the method name somehow without making it a massive long method name! 🤣

integration_tests/mockApis/curiousApi.ts
```diff
@@ -316,6 +317,125 @@ const stubLearnerEducation = (prisonNumber = 'G6115VJ', page = 0): SuperAgentReq
...
+            qcfCertificateIndicator: false,
+            lrsGLH: 0,
+            attendedGLH: 100,
+            actualGLH: 200,
+            outcome: 'No achievement',
+            outcomeGrade: null,
+            employmentOutcome: null,
+            withdrawalReasons: 'Other',
+            prisonWithdrawalReason: 'Significant ill health causing them to be unable to attend education',
+            completionStatus: 'The learner has withdrawn from the learning activities leading to the learning aim',
```
nathanrussell-moj-digital, line 371: Something is not right here ....
The tech notes in the story say this:

> **Completed courses**
> Filter courses from API response by `completionStatus === Completed`

So, I would expect this line in your stub to say 
```
completionStatus: 'Completed',
```

But because it doesn't ..... it's making me wonder how the code is even working? Does your implementation (in your previous PR) not filter on `completionStatus` ?

integration_tests/pages/overview/EducationAndTrainingPage.ts
```diff
@@ -40,13 +40,18 @@ export default class EducationAndTrainingPage extends Page {
...
-  hasCompletedInPrisonQualificationsDisplayed(): EducationAndTrainingPage {
-    this.completedInPrisonQualificationsTable().should('be.visible')
+  hasCompletedInPrisonQualificationsLast12MonthsDisplayed(): EducationAndTrainingPage {
+    this.completedInPrisonQualificationsLast12MonthsTable().should('be.visible')
     return this
   }
 
-  doesNotCompletedInPrisonQualificationsDisplayed(): EducationAndTrainingPage {
-    this.completedInPrisonQualificationsTable().should('not.exist')
+  hasCompletedInPrisonQualificationRecordDisplayed(expected: string): EducationAndTrainingPage {
```
nathanrussell-moj-digital, line 48: So this is just asserting the table is visible .. 🤔 
I wonder if that is part of the problem in this PR .... to be able to assert there are completed courses displayed, we need to assert the table has some records (`<tr>`s perhaps?), and/or does also contain the "no records" message

integration_tests/pages/overview/EducationAndTrainingPage.ts
```diff
@@ -40,13 +40,18 @@ export default class EducationAndTrainingPage extends Page {
     return this
   }
 
-  hasCompletedInPrisonQualificationsDisplayed(): EducationAndTrainingPage {
-    this.completedInPrisonQualificationsTable().should('be.visible')
+  hasCompletedInPrisonQualificationsLast12MonthsDisplayed(): EducationAndTrainingPage {
+    this.completedInPrisonQualificationsLast12MonthsTable().should('be.visible')
```
nathanrussell-moj-digital, line 44: So this is just asserting the table is visible .. 🤔
I wonder if that is part of the problem in this PR .... to be able to assert there are completed courses displayed, we need to assert the table has some records (<tr>s perhaps?), and/or does also contain the "no records" message

integration_tests/mockApis/curiousApi.ts
chrimesdev: The code I have added uses the result of `const completedInPrisonEducation = completedInPrisonEducationRecords(allInPrisonEducation)` 

The existing `inPrisonEducationRecordsResolver` : `completedInPrisonEducationRecords` should handle the `completionStatus` filtering.

nathanrussell-moj-digital: Hmmm, well I don't understand whats going on here then .... 🤔  .... how is this a "completed" course? 🤔

integration_tests/pages/overview/EducationAndTrainingPage.ts
chrimesdev: I think because the designs have the empty state message "No courses or qualifications completed in last 12 months" as an actual table row; make it a little more difficult to test this. 

Rather than having a Table for data, then a paragraph element for the empty state message (which probably would also be more semantically correct) - which would clear up the confusion here I think

integration_tests/mockApis/curiousApi.ts
nathanrussell-moj-digital: We still need to get to the bottom of this - how and why is this course considered a "completed" course?

integration_tests/mockApis/curiousApi.ts
```diff
@@ -316,6 +317,209 @@ const stubLearnerEducation = (prisonNumber = 'G6115VJ', page = 0): SuperAgentReq
...
```
nathanrussell-moj-digital, line 455: As above, this does not look like a "completed" course 🤔

nathanrussell-moj-digital (approved):
OK, lets merge this, but we have to understand (and potentially fix) how we know if a course is completed before we can consider RR-590 complete (meeting booked for tomorrow morning)

PR #494, opened 2024-03-13
RR-722 - Display the Qualification Level page

mattwills-moj-digital:


server/views/pages/induction/prePrisonEducation/qualificationLevel.njk
```diff
@@ -0,0 +1,165 @@
...
+{# Setup page based on education level #}
+{% set options = [
+  {
+    value: "ENTRY_LEVEL",
+    text: "ENTRY_LEVEL" | formatQualificationLevel,
+    hint: {
+      text: "ENTRY_LEVEL" | formatQualificationLevelHint
+    },
+    checked: form.qualificationLevel === "ENTRY_LEVEL"
+  },
```
mattwills-moj-digital, line 27: This logic has been copied over from Gary's template, but with our approach for checkboxes (or in this case radio buttons) applied.

server/filters/formatQualificationLevelHintFilter.ts
```diff
@@ -0,0 +1,16 @@
+export default function formatQualificationLevelHintFilter(value: string): string {
```
mattwills-moj-digital, line 1: I thought we could use a separate filter for the "hints", but happy to canvas opinion on this.

nathanrussell-moj-digital: Yeah, I'm happy with this.
We don't have an equivalent of Gary's "contentLookup" , and I don't think we need one. I think thats what filters are for

We have other Induction pages that we have been migrating as part of this where we have hardcoded the hints in the template because we didnt have the corresponding filter etc, and at the time perhaps it was easier/quicker to just put the content in the template?

But I prefer this type of approach, and it's been there in the back of my mind that we should go back and write filters or similar for the hint text

This makes a good start and establishes that pattern. Good work 👍

server/@types/forms/index.d.ts
```diff
@@ -124,6 +124,11 @@ declare module 'inductionForms' {
     educationLevel: EducationLevelValue
   }
 
+  export interface QualificationLevelForm {
+    educationLevel: EducationLevelValue
```
nathanrussell-moj-digital, line 128: Why does the form need the education level? We currently have that modelled in it's own form `HighestLevelOfEducationForm` ... 🤔

server/routes/induction/update/qualificationLevelUpdateController.ts
```diff
@@ -0,0 +1,21 @@
+import { Request } from 'express'
+import { InductionService } from '../../../services'
+import QualificationLevelController from '../common/qualificationLevelController'
+
+/**
+ * Controller for the Update of the WorkExperience screen of the Induction.
```
nathanrussell-moj-digital, line 6: Minor copy and paste typo

server/@types/forms/index.d.ts
mattwills-moj-digital: The logic in the template uses it. For example:

```
{% if ['FURTHER_EDUCATION_COLLEGE', 'UNDERGRADUATE_DEGREE_AT_UNIVERSITY', 'POSTGRADUATE_DEGREE_AT_UNIVERSITY'].includes(form.educationLevel) or form.educationLevel === '' %}
  {% set title = "What level of further education college qualification does " + prisonerSummary.firstName + " " + prisonerSummary.lastName + " want to add?" %}
```

server/views/pages/induction/prePrisonEducation/qualificationLevel.njk
nathanrussell-moj-digital: Yes, that is the correct logic to use 👍

server/@types/forms/index.d.ts
mattwills-moj-digital: Ah yes, that makes sense 👍

nathanrussell-moj-digital (approved):

PR #496, opened 2024-03-14
RR-674 template for in progress courses with new column logic

chrimesdev:
## Description

Boilerplate nunjucks template for the In progress courses and qualifications tab, ready for data when we have it.

Plus additional logic in the macro to handle optional columns from the design, I think once we have data, we may be able to come up with a better approach for this. Tried to keep the Nunjucks template clean as possible whilst moving the logic into the macro - open to thoughts on this!

https://dsdmoj.atlassian.net/browse/RR-674

### Screenshots

screencapture-localhost-3000-plan-G9981UK-in-prison-courses-and-qualifications-2024-03-14-10_52_47

screencapture-localhost-3000-plan-G9981UK-in-prison-courses-and-qualifications-2024-03-14-10_52_13 (1)

### DIFF
--- a/server/views/pages/inPrisonCoursesAndQualifications/index.njk
+++ b/server/views/pages/inPrisonCoursesAndQualifications/index.njk
@@ -38,16 +38,37 @@
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-full">
 
+      {% set curiousHelpText %}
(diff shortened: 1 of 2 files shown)


nathanrussell-moj-digital (approved):
LGTM

PR #504, opened 2024-03-18
RR-731 - Support for back links in cypress tests

nathanrussell-moj-digital:
This PR adds support for testing back links in our cypress tests
I've created an `InductionPage` superclass that all of the Induction based pages now extend; and in `InductionPage` I've added some basic functions for testing and asserting the back links
And I've added a couple of calls to the assertion methods in some of the tests as examples.

The work we are doing in RR-731 (@chrimesdev) and RR-732 (when we get to it) will make use of this 👍

### DIFF
--- a/integration_tests/pages/induction/InductionPage.ts
+++ b/integration_tests/pages/induction/InductionPage.ts
@@ -1,0 +1,26 @@
+import Page, { PageElement } from '../page'
+
+export default abstract class InductionPage extends Page {
+  constructor(pageId: string, options?: { axeTest?: boolean }) {
+    super(pageId, options)
+  }
+
+  clickBackLinkTo = <T extends Page>(expected: new () => T): T => {
+    this.backLink().click()
+    return Page.verifyOnPage(expected)
+  }
+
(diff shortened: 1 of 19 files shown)


integration_tests/pages/induction/InductionPage.ts
```diff
@@ -0,0 +1,26 @@
+import Page, { PageElement } from '../page'
+
+export default abstract class InductionPage extends Page {
+  constructor(pageId: string, options?: { axeTest?: boolean }) {
+    super(pageId, options)
+  }
+
+  clickBackLinkTo = <T extends Page>(expected: new () => T): T => {
```
nathanrussell-moj-digital, line 8: I've not written any tests currently call to call this method, but if we did want to write a test that clicked the Back link we should be able to do something like:

```
const affectAbilityToWorkPage = Page.verifyOnPage(AffectAbilityToWorkPage)
...
...
const workAndInterestsPage = affectAbilityToWorkPage.clickBackLinkTo(WorkAndInterestsPage)

```

mattwills-moj-digital (approved):
AToAB 👍

PR #507, opened 2024-03-19
RR-731 set aria labels for the back link on update screens

chrimesdev:
## Description

Set `aria-label` for the back link on update screens

https://dsdmoj.atlassian.net/browse/RR-731


nathanrussell-moj-digital (approved):
LGTM 👍

PR #510, opened 2024-03-19
RR-732 - Dynamic back links for add qualification screens

mattwills-moj-digital:
This PR implements the `getBackLinkUrl()` methods for the add qualification related pages (i.e. "Qualification Level" and "Qualification Details").

Since these pages are part of a mini flow, I'm reusing the excellent `PageFlowQueue` that @nathanrussell-moj-digital added in a previous PR for the previous work experience functionality. It took me a while to grasp that the key thing here is to add the current and previous page when first building the `PageFlowQueue`. However, rather than defining the complete set of pages initially, I'm adding the "Qualification Detail" page to the array of pages when the user goes to that page, as I think this makes the flow simpler and potentially more dynamic for other flows in the future.

I aim to implement the `getBackLinkAriaText()` functionality in my next PR.

### DIFF
--- a/server/routes/induction/common/qualificationDetailsController.ts
+++ b/server/routes/induction/common/qualificationDetailsController.ts
@@ -14,7 +14,10 @@ export default abstract class QualificationDetailsController extends InductionCo
     res: Response,
     next: NextFunction,
   ): Promise<void> => {
-    const { prisonerSummary, qualificationLevelForm } = req.session
+    const { pageFlowQueue, prisonerSummary, qualificationLevelForm } = req.session
+    if (!pageFlowQueue) {
+      return res.redirect(`/plan/${prisonerSummary.prisonNumber}/view/education-and-training`)
+    }
 
     const qualificationDetailsForm = req.session.qualificationDetailsForm || {
       qualificationSubject: '',
--- a/server/routes/induction/update/highestLevelOfEducationUpdateController.ts
+++ b/server/routes/induction/update/highestLevelOfEducationUpdateController.ts
@@ -2,12 +2,14 @@ import createError from 'http-errors'
 import { NextFunction, Request, RequestHandler, Response } from 'express'
 import type { InductionDto } from 'inductionDto'
 import type { HighestLevelOfEducationForm } from 'inductionForms'
+import type { PageFlowQueue } from 'viewModels'
 import HighestLevelOfEducationController from '../common/highestLevelOfEducationController'
 import { InductionService } from '../../../services'
 import validateHighestLevelOfEducationForm from './highestLevelOfEducationFormValidator'
 import toCreateOrUpdateInductionDto from '../../../data/mappers/createOrUpdateInductionDtoMapper'
 import EducationLevelValue from '../../../enums/educationLevelValue'
 import logger from '../../../../logger'
+import { getNextPage } from '../../pageFlowQueue'
 
 /**
  * Controller for the Update of the Highest Level of Education screen of the Induction.
@@ -60,6 +62,9 @@ export default class HighestLevelOfEducationUpdateController extends HighestLeve
 
       try {
         await this.inductionService.updateInduction(prisonNumber, updateInductionDto, req.user.token)
+        req.session.highestLevelOfEducationForm = undefined
+        req.session.inductionDto = undefined
+        return res.redirect(`/plan/${prisonNumber}/view/education-and-training`)
       } catch (e) {
         logger.error(`Error updating Induction for prisoner ${prisonNumber}`, e)
         return next(createError(500, `Error updating Induction for prisoner ${prisonNumber}. Error: ${e}`))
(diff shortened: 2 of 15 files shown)


server/routes/induction/update/highestLevelOfEducationUpdateController.ts
```diff
@@ -60,6 +62,9 @@ export default class HighestLevelOfEducationUpdateController extends HighestLeve
 
       try {
         await this.inductionService.updateInduction(prisonNumber, updateInductionDto, req.user.token)
+        req.session.highestLevelOfEducationForm = undefined
+        req.session.inductionDto = undefined
+        return res.redirect(`/plan/${prisonNumber}/view/education-and-training`)
```
mattwills-moj-digital, line 67: I've simply moved this up from lines 77-79

server/routes/induction/common/qualificationDetailsController.ts
```diff
@@ -14,7 +14,10 @@ export default abstract class QualificationDetailsController extends InductionCo
     res: Response,
     next: NextFunction,
   ): Promise<void> => {
-    const { prisonerSummary, qualificationLevelForm } = req.session
+    const { pageFlowQueue, prisonerSummary, qualificationLevelForm } = req.session
+    if (!pageFlowQueue) {
```
nathanrussell-moj-digital, line 18: This check in this controller, and the next one, could (probably should) be moved out to a route request handler (the same as we do currently with `checkPrisonerSummaryExistsInSession`, `checkUpdateGoalFormExistsInSession` etc)
It's fine as it is here and now in this PR, but we should probably look to refactor it out to `routerRequestHandlers.ts` where it can be tested etc, in the future 👍

server/routes/induction/update/qualificationDetailsUpdateController.test.ts
```diff
@@ -51,6 +51,13 @@ describe('qualificationDetailsUpdateController', () => {
       req.session.inductionDto = inductionDto
       const qualificationLevelForm = { qualificationLevel: QualificationLevelValue.LEVEL_3 }
       req.session.qualificationLevelForm = qualificationLevelForm
+      req.session.pageFlowQueue = {
+        pageUrls: [
+          `/prisoners/${prisonNumber}/induction/qualifications`,
+          `/prisoners/${prisonNumber}/induction/qualification-level`,
+        ],
+        currentPageIndex: 2,
```
nathanrussell-moj-digital, line 59: Minor, but the `currentPageIndex` is zero indexed, so this should be `1`

server/routes/induction/update/qualificationLevelUpdateController.test.ts
```diff
@@ -54,6 +54,10 @@ describe('qualificationLevelUpdateController', () => {
       const inductionDto = aLongQuestionSetInductionDto()
       req.session.inductionDto = inductionDto
       req.session.qualificationLevelForm = undefined
+      req.session.pageFlowQueue = {
+        pageUrls: [`/prisoners/${prisonNumber}/induction/qualifications`],
```
nathanrussell-moj-digital, line 58: Should this have at least 1 elements in it? and the currentPageIndex is zero indexed

server/routes/pageFlowQueue.ts
```diff
@@ -4,6 +4,17 @@ import type { PageFlowQueue } from 'viewModels'
  * Module exposing utility functions that operate on [PageFlowQueue] instances
  */
 
+/**
+ * Adds the provided page url to the last position in the specified [PageFlowQueue].
+ * Does not change the current page (current page index).
+ */
+const addPage = (pageFlowQueue: PageFlowQueue, pageUrl: string): PageFlowQueue => {
```
nathanrussell-moj-digital, line 11: Do we need a unit test for this new method?

mattwills-moj-digital: Not sure that's necessary. It's also tested indirectly.

server/routes/induction/update/qualificationDetailsUpdateController.test.ts
mattwills-moj-digital: It is indeed 0 indexed, but there are a couple of things to consider:

1. `getPreviousPage()` subtracts 1 from the index: `return pageFlowQueue.pageUrls[pageFlowQueue.currentPageIndex - 1]`. If I change this number to 1 then the test breaks.
2. Normally the `routerRequestHandler` would increment the page index, but that's not involved in this unit test.

server/routes/induction/update/qualificationLevelUpdateController.test.ts
mattwills-moj-digital: I've changed this to:

```
pageUrls: [
   `/prisoners/${prisonNumber}/induction/qualifications`,
   `/prisoners/${prisonNumber}/induction/qualification-level`,
],
```

But I think the index needs to remain at 1, since the `routerRequestHandler` is not involved here.

server/routes/induction/update/qualificationDetailsUpdateController.test.ts
mattwills-moj-digital: I've added an additional URI, since we agreed late yesterday that the full flow (including `/qualifications-details` which is currently the last page in this flow) should be included.

server/routes/pageFlowQueue.ts
mattwills-moj-digital: Done, although I think we need to give careful consideration to our testing strategy: https://blog.cleancoder.com/uncle-bob/2017/10/03/TestContravariance.html

nathanrussell-moj-digital (approved):
LGTM 👍

PR #511, opened 2024-03-19
Temporarily set anomaly score for rule 921110 to 5 in dev

leejacobson-moj:
## Overview

Temporarily set modsec anomaly score for rule 921110 to 5 in dev to test request blocking logic


nathanrussell-moj-digital (approved):
LGTM 👍 

Let's hope this works, and if it does then perhaps we can do the same with rule `942230` (SQL injection) which we currently have disabled altogether 🤞

PR #516, opened 2024-03-20
RR-732 - Return aria text for qualifications back links

mattwills-moj-digital:
This PR introduces `dynamicAriaTextResolver` to get the aria text for a given URI back link. I've started with the add qualification mini flow, which results in the aria text:

image

The plan is to add the aria text for the remaining back links. For example, we need to replace wherever we have:

```
getBackLinkAriaText(_req: Request): string {
   return 'Back to <TODO - check what CIAG UI does here>'
}
```
With:
```
getBackLinkAriaText(req: Request): string {
   return getDynamicBackLinkAriaText(req, this.getBackLinkUrl(req))
}
```
And also add the required URI and text to the `dynamicAriaTextResolver`.


nathanrussell-moj-digital (approved):
LGTM 👍

PR #517, opened 2024-03-21
RR-590 - Fixes the logic to determine status of Curious courses

nathanrussell-moj-digital:
This PR fixes the logic to determine status of Curious courses, along with changes to associated code

The basic idea is that a Curious course has one of 4 statuses - In Progress, Completed, Withdrawn and Temporarily Withdrawn.
The problem is that the Curious API contains a field called `completionStatus` that, according to their swagger spec, returns what looks like it should be an enum:
Screenshot 2024-03-21 at 08 21 10

However, on looking at example records we get back from Curious `dev`, the field appears to be some kind of free text, where we have seen examples like:
```
completionStatus: 'The learner has withdrawn from the learning activities leading to the learning aim'
```

To date we have only needed to know whether a Curious course has been completed, so we have used the `courseCompletionDate` field - if the field has a value we assume the course was completed.

This no longer works for us, as we now need to know the full range of statuses, as we need to group and display on screen courses that are In Progress, Completed, Withdrawn and Temporarily Withdrawn.

Meganexus have not been very forthcoming with any help in this area; but Ben Pearce has done some fine investigation and analysis work, and has discovered that the Curious UI has a drop down for this status with 4 choices .... where the onscreen label appears to be the value we are receiving back in the API:
image

From this it looks like the API does indeed return an enum (of sorts) for the `completionStatus` field (though it is nothing like what their swagger spec says! 🙄 ). I say "enum of sorts" because it's not the way we would do it, and presumably if the screen values ever change, then so does the API response.


integration_tests/e2e/overview/educationAndTrainingTab.cy.ts
```diff
@@ -354,7 +354,7 @@ context('Prisoner Overview page - Education And Training tab', () => {
   describe('should display a link to view all courses and qualifications', () => {
     it(`should link to the courses and qualifications page`, () => {
       // Given
-      cy.task('stubGetInductionShortQuestionSet')
+      cy.task('stubLearnerEducationWithCoursesQualificationsCompletedInLast12Months')
```
nathanrussell-moj-digital, line 357: This cypress test might need some love in a future PR as I'm not convinced either the test or the implementation is 100% correct. The change I've made to the stub here is enough to keep it passing, but I think we'll be looking at this test again in the near future.

server/@types/viewModels/index.d.ts
```diff
@@ -107,7 +107,7 @@ declare module 'viewModels' {
     courseName: string
     courseCode: string
     courseStartDate: Date
-    courseCompleted: boolean
+    courseStatus: 'COMPLETED' | 'IN_PROGRESS' | 'WITHDRAWN' | 'TEMPORARILY_WITHDRAWN'
```
nathanrussell-moj-digital, line 110: Update our view model by removing the `courseCompleted: boolean` field, and replacing it with a `courseStatus` field of an enum('ish) type.
Essentially this is the main ask of this PR - we need our view model representation of a Curious course to have a definitive status so that we can group and display them onscreen.

Everything else in this PR is to support this view model change, be it changes to the mapper, tests, test data builders etc

server/data/mappers/inPrisonEducationMapper.test.ts
```diff
@@ -1,139 +1,172 @@
...
-      grade: null, // expect grade to be null (because there is no outcome grade or outcome)
-      source: 'CURIOUS',
-    }
-
-    // When
-    const actual = toInPrisonEducation(apiLearnerEducation)
-
-    // Then
-    expect(actual).toEqual(expectedInPrisonEducation)
+  describe('toCourseStatus', () => {
```
nathanrussell-moj-digital, line 141: New tests for the new `toCourseStatus` function

server/data/mappers/inPrisonEducationMapper.test.ts
```diff
@@ -1,139 +1,172 @@
...
-      grade: null,
-      source: 'CURIOUS',
-    }
-
-    // When
-    const actual = toInPrisonEducation(apiLearnerEducation)
-
-    // Then
-    expect(actual).toEqual(expectedInPrisonEducation)
+  describe('toInPrisonEducation', () => {
```
nathanrussell-moj-digital, line 10: This looks like a lot of change, but its actually just indentation because I've wrapped the existing tests in a `describe('toInPrisonEducation', () => {` block to separate them from the new tests I've written below

server/data/mappers/inPrisonEducationMapper.ts
```diff
@@ -19,4 +19,40 @@ const toInPrisonEducation = (apiLearnerEducation: LearnerEducation): InPrisonEdu
   }
 }
 
-export default toInPrisonEducation
+/**
```
nathanrussell-moj-digital, line 22: Hopefully my doc here is sufficient and will prevent someone tripping over this in the future

mattwills-moj-digital: It's very good 🙌

server/data/mappers/inPrisonEducationMapper.test.ts
mattwills-moj-digital: I wish whitespace changes were hidden by default, without having to select that option and reload the page (or adding `&w=1` to the URL).

mattwills-moj-digital (approved):
A fine piece of work 🙌

PR #522, opened 2024-03-21
RR-590 - Refactor how Curious In-Prison courses are mapped into view models

nathanrussell-moj-digital:
This PR (is big! sorry!) refactors how we map the Curious `LearnerEducation` records into a view model, and then how we display these on screen.

Previously we mapped from Curious' data into a view model called `InPrisonEducationRecords` which in turn had an array of `InPrisonEducation`. 

There were several problems with this, especially with what we are currently being asked to do re: grouping them by status.

The term "InPrison**Education**" is not something we or the business use elsewhere. Its more often referred to as an In Prison **Course**

The other problem is that the service layer mapped all returned `LearnerEducation`s into `InPrisonEducation`s without any consideration or differentiation of the course status.
In current production we have 2 use cases for this data - we need to display a maximum of the 2 most recent completed courses on the Overview page; and we need to display all completed courses on the main education & training page.
Currently we do this in the controller. IE. the controller calls the service method which gets all courses (`InPrisonEducation`s); and within the controller, via some "resolvers" (sic !), we essentially filter out what we don't need.
It's not great to be honest.

We are now being asked to further group the course data by status - completed, in progress, withdrawn, and temporarily withdrawn. And further, of the completed courses, we need to be able to show the courses completed in the last 12 months.

This is a lot more filtering and general data munging; and the controllers were looking like they were going to get pretty messy, with calls to new "resolvers" 🤢 🤮 

I've stepped back from this and looked again at what we are trying to do, and decided that the view model itself was wrong. Not only did it have the wrong names etc (easily fixed), but the grouping of courses by type etc really needed to be there in the view model. That would mean the controllers can simply ask the service for the view model, and then not have to worry about further processing/filtering etc, before passing it into the nunjucks view. It can just pass it as is, and let the nunjucks view reference and display the relevant properties.

So; the refactoring here is really about the view model, but that obviously impacts the view model itself, the mappers, the service, the controllers, and the nunjucks templates.

In theory it shouldn't have impacted the cypress tests, but I've taken the opportunity to get in there and tidy up some naming etc, and the grouping of some of the tests.


integration_tests/e2e/overview/educationAndTrainingTab.cy.ts
```diff
@@ -350,22 +373,4 @@ context('Prisoner Overview page - Education And Training tab', () => {
       Page.verifyOnPage(QualificationsListPage)
     })
   })
-
-  describe('should display a link to view all courses and qualifications', () => {
```
nathanrussell-moj-digital, line 354: This test is covered in my refactored tests above ☝️

integration_tests/mockApis/curiousApi.ts
```diff
@@ -317,7 +317,7 @@ const stubLearnerEducation = (prisonNumber = 'G6115VJ', page = 0): SuperAgentReq
     },
   })
 
-const stubLearnerEducationWithCoursesQualificationsCompletedInLast12Months = (
+const stubLearnerEducationWithCompletedCoursesInLast12Months = (
```
nathanrussell-moj-digital, line 320: Simplified and improved the names

integration_tests/pages/overview/EducationAndTrainingPage.ts
```diff
@@ -33,8 +33,21 @@ export default class EducationAndTrainingPage extends Page {
     return this
   }
 
-  hasFunctionalSkillsDisplayed(): EducationAndTrainingPage {
+  hasFunctionalSkillWithAssessmentScoreDisplayed(
```
nathanrussell-moj-digital, line 36: Improved the naming and functionality of the assertion methods in the page class, so as to give a better API. It means the cypress tests that use them are much more readable 👍

integration_tests/pages/overview/EducationAndTrainingPage.ts
```diff
@@ -43,25 +56,25 @@ export default class EducationAndTrainingPage extends Page {
...
-  doesNotCompletedInPrisonQualificationsLast12MonthsDisplayed(): EducationAndTrainingPage {
-    this.completedInPrisonQualificationsLast12MonthsTable().should('not.exist')
+  doesNotHaveCompletedCoursesInLast12MonthsDisplayed(): EducationAndTrainingPage {
+    this.completedInPrisonCoursesInLast12MonthsTable().should('not.exist')
     return this
   }
 
   hasCuriousUnavailableMessageDisplayed(): EducationAndTrainingPage {
-    this.curiousUnavailableMessage().should('be.exist')
+    this.curiousUnavailableMessage().should('be.visible')
```
nathanrussell-moj-digital, line 72: `be.exist` is horrible to read! `be.visible` is much better 👍

server/@types/viewModels/index.d.ts
```diff
@@ -91,17 +91,19 @@ declare module 'viewModels' {
   }
 
   /**
-   * A prisoner's record of In-Prison courses and education, which is made up of a collection of InPrisonEducation.
+   * A prisoner's record of In-Prison courses, which is made up of collections of [InPrisonCourse].
```
nathanrussell-moj-digital, line 94: Essentially this is the crux of the change; changing these view models

server/data/mappers/inPrisonCourseMapper.ts
```diff
@@ -1,17 +1,17 @@
...
 
-const toInPrisonEducation = (apiLearnerEducation: LearnerEducation): InPrisonEducation => {
+const toInPrisonCourse = (apiLearnerEducation: LearnerEducation): InPrisonCourse => {
   return {
     prisonId: apiLearnerEducation.establishmentId,
     prisonName: apiLearnerEducation.establishmentName,
     courseCode: apiLearnerEducation.courseCode,
     courseName: apiLearnerEducation.courseName,
-    courseStartDate: moment(apiLearnerEducation.learningStartDate).toDate(),
+    courseStartDate: moment(apiLearnerEducation.learningStartDate).utc().toDate(),
```
nathanrussell-moj-digital, line 11: Ensure the dates are UTC (one of the tests burnt me with this, hence the change 😁 )

server/routes/inPrisonEducationRecordsResolver.test.ts
nathanrussell-moj-digital, line 1: Adios "resolver" !!! 😁

server/routes/overview/overviewController.ts
```diff
@@ -117,22 +110,11 @@ export default class OverviewController {
     const allFunctionalSkills = await this.curiousService.getPrisonerFunctionalSkills(prisonNumber, req.user.username)
     const functionalSkills = mostRecentFunctionalSkills(allFunctionalSkills)
 
-    const allInPrisonEducation = await this.curiousService.getLearnerEducation(prisonNumber, req.user.username)
-    const completedInPrisonEducation = completedInPrisonEducationRecords(allInPrisonEducation)
-    const completedInPrisonEducationWithinLast12Months = await this.setPrisonNamesOnEducationRecords(
-      completedInPrisonEducationRecordsWithinLast12Months(completedInPrisonEducation),
-      req,
-    )
+    const inPrisonCourses = await this.curiousService.getPrisonerInPrisonCourses(prisonNumber, req.user.username)
```
nathanrussell-moj-digital, line 113: The controller code is much simpler now, now that it doesnt have to faff around with filtering and mutating the course data it gets from the service 👍

server/routes/overview/overviewController.ts
```diff
@@ -154,23 +136,4 @@ export default class OverviewController {
     const view = new TimelineView(prisonerSummary, timeline)
     res.render('pages/overview/index', { ...view.renderArgs })
   }
-
-  async setPrisonNamesOnEducationRecords(
```
nathanrussell-moj-digital, line 158: The prison name lookup is done by the service layer, our controller doesnt need to worry about it 👍

server/routes/overview/overviewController.ts
nathanrussell-moj-digital: And my next PR, which is waiting in the wings for this one to be merged first, will make the controller code even simpler still in respect of getting data like this from services 👍  (bet you can't wait! 🤣 )

mattwills-moj-digital (approved):
LGTM 👍

PR #524, opened 2024-03-22
RR-590 - Retrieve curious In Prison Courses via router request handler

nathanrussell-moj-digital:
This PR changes the way we retrieve the Curious In Prison Courses; from being a responsibility of the controller method, to being a router request handler that puts the result in [`res.locals`](https://expressjs.com/en/api.html#res.locals)

This is a better pattern, especially for data like this where we simply want to pass it to the view. It means the controller is not responsible for this and can just assume the data is already there on `res.locals`. Some of our controllers and associated unit tests are getting a little "fat" in this respect, so this is a good thing.

For the time being I'm only doing this for retrieving the Curious In Prison Course data, but there are other similar service requests that we currently make in the controller that we could refactor in a similar manner. I'll add some tickets to do those in the future 👍

### DIFF
--- a/server/routes/inPrisonCoursesAndQualifications/index.ts
+++ b/server/routes/inPrisonCoursesAndQualifications/index.ts
@@ -2,7 +2,7 @@ import type { Router } from 'express'
 import config from '../../config'
 import { Services } from '../../services'
 import { checkUserHasViewAuthority } from '../../middleware/roleBasedAccessControl'
-import { retrievePrisonerSummaryIfNotInSession } from '../routerRequestHandlers'
+import { retrieveCuriousInPrisonCourses, retrievePrisonerSummaryIfNotInSession } from '../routerRequestHandlers'
 import InPrisonCoursesAndQualificationsController from './inPrisonCoursesAndQualificationsController'
 
 /**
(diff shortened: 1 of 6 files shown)


server/routes/inPrisonCoursesAndQualifications/index.ts
```diff
@@ -18,6 +18,7 @@ export default (router: Router, services: Services) => {
     router.get('/plan/:prisonNumber/in-prison-courses-and-qualifications', [
       checkUserHasViewAuthority(),
       retrievePrisonerSummaryIfNotInSession(services.prisonerSearchService),
+      retrieveCuriousInPrisonCourses(services.curiousService),
```
nathanrussell-moj-digital, line 21: Retrieve the Curious In Prison Courses as part of the request handler chain, rather than in the controller itself 👍

server/routes/routerRequestHandlers.ts
```diff
@@ -221,6 +222,29 @@ const retrieveFunctionalSkillsIfNotInSession = (curiousService: CuriousService):
   }
 }
 
+/**
```
nathanrussell-moj-digital, line 225: This file (`routerRequestHandlers.ts`) and it's test are getting massive! And not in a [good way](https://www.youtube.com/watch?v=O7TklQTeuSE)!

I'll raise a jira shortly to split these out into their own source files etc 👍

nathanrussell-moj-digital: I've raised RR-750 for this refactoring work 👍

mattwills-moj-digital (approved):
AToAB

PR #527, opened 2024-03-25
RR-749: Activities API client

leejacobson-moj:
## Overview

Added activities API client and methods for retrieving prisoner allocations and waiting list applications ready for activities integration work.


server/data/restClient.ts
```diff
@@ -92,51 +93,52 @@ export default class RestClient {
     return ''
   }
 
-  async get({
+  async get<Response = unknown>({
```
leejacobson-moj, line 96: Updated this inline with the TS template. Instead of returning `Promise<unknown>` client calls can now return the specified response type.

server/testsupport/activitiesApi/allocationTestDataBuilder.ts
leejacobson-moj, line 1: I've placed this and `waitingListTestDataBuilder.ts` in an `activitiesApi` directory. Happy to put these directly in `testsupport` like the others if preferred. There's quite a few files in there and I thought it might be easier to find them if we started to categorise them under sub-directories.

nathanrussell-moj-digital (approved):
LGTM; though could you update the following sections in the `README` please:

`## Imported Types` - to document the command you ran / that someone will need to run in the future to import the A&A swagger spec

`### Client Roles` - I'm assuming the client you've introduced here will use the system token, rather then the user token? We should document the role required 👍

nathanrussell-moj-digital (approved):
LGTM 👍

PR #529, opened 2024-03-25
RR-738 - Controllers and route for Want to Add Qualifications page

mattwills-moj-digital:
A previous PR had added the nunjucks template for the `Want to Add Qualifications` page, but we didn't have the corresponding controllers & route. 

Note that the submit currently doesn't work without a `pageFlowQueue`, since it will only ever be called when switching from the long question set to the short one (i.e. when it is part of this main flow).

### DIFF
--- a/server/routes/induction/common/qualificationsListController.ts
+++ b/server/routes/induction/common/qualificationsListController.ts
@@ -37,6 +37,7 @@ export default abstract class QualificationsListController extends InductionCont
   }
 }
 
+// TODO - this is duplicated in WantToAddQualificationsController - needs putting somewhere common
 const mostRecentAssessments = (allAssessments: Array<Assessment>): Array<Assessment> => {
   const allAssessmentsGroupedByTypeSortedByDateDesc = assessmentsGroupedByTypeSortedByDateDesc(allAssessments)
 
--- a/server/routes/induction/common/wantToAddQualificationsController.ts
+++ b/server/routes/induction/common/wantToAddQualificationsController.ts
@@ -1,0 +1,74 @@
+import { NextFunction, Request, RequestHandler, Response } from 'express'
+import type { Assessment } from 'viewModels'
+import type { WantToAddQualificationsForm } from 'inductionForms'
+import InductionController from './inductionController'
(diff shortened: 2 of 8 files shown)


server/routes/induction/common/wantToAddQualificationsController.ts
```diff
@@ -0,0 +1,74 @@
...
+      this.getBackLinkUrl(req),
+      this.getBackLinkAriaText(req),
+      wantToAddQualificationsForm,
+      functionalSkills,
+    )
+    return res.render('pages/induction/prePrisonEducation/wantToAddQualifications', { ...view.renderArgs })
+  }
+}
+
+const mostRecentAssessments = (allAssessments: Array<Assessment>): Array<Assessment> => {
```
mattwills-moj-digital, line 40: This is currently copied from `QualificationsListController`, so we should put this somewhere common, but it's not my focus right now.

server/routes/induction/common/wantToAddQualificationsController.ts
```diff
@@ -0,0 +1,75 @@
...
+      this.getBackLinkUrl(req),
+      this.getBackLinkAriaText(req),
+      wantToAddQualificationsForm,
+      functionalSkills,
+    )
+    return res.render('pages/induction/prePrisonEducation/wantToAddQualifications', { ...view.renderArgs })
+  }
+}
+
+// TODO - this is duplicated in QualificationsListController - needs putting somewhere common
```
mattwills-moj-digital, line 40: I've added these TODOs for now, but it's not my primary focus at this minute.

server/routes/induction/common/qualificationsListController.ts
```diff
@@ -37,6 +37,7 @@ export default abstract class QualificationsListController extends InductionCont
   }
 }
 
+// TODO - this is duplicated in WantToAddQualificationsController - needs putting somewhere common
```
nathanrussell-moj-digital, line 40: Yeah; agreed. I wonder if the router request handler is the right place to do this (or perhaps a specific router request handler that returns just the most recent assessments .... 🤔

nathanrussell-moj-digital (approved):
LGTM

PR #531, opened 2024-03-26
RR-744 create a basic view for new add goals flow

chrimesdev:
## Description 

Create a basic view for new add goals flow. Also added a basic Route and Controller method to be able to access the new page. These will be updated in the next PRs.

https://dsdmoj.atlassian.net/browse/RR-744

### Screenshot

screencapture-localhost-3000-plan-G5045UF-goals-1-create-2024-03-26-09_05_47

### DIFF
--- a/server/routes/createGoal/createGoalController.ts
+++ b/server/routes/createGoal/createGoalController.ts
@@ -54,36 +54,6 @@ export default class CreateGoalController {
     return res.render('pages/goal/create/index', { ...view.renderArgs })
   }
 
-  // TODO: RR-734 - Create controller and view classes new create goal journey
-  newGetCreateGoalView: RequestHandler = async (req, res, next): Promise<void> => {
-    const { prisonNumber } = req.params
-    const { prisonerSummary } = req.session
-
-    req.session.newGoals = req.session.newGoals || []
-
-    if (!req.session.newGoal?.createGoalForm) {
-      req.session.newGoal = {
-        createGoalForm: { prisonNumber },
-      } as NewGoal
-    }
-
-    const today = moment().toDate()
-    const futureGoalTargetDates = [
-      futureGoalTargetDateCalculator(today, 3),
-      futureGoalTargetDateCalculator(today, 6),
-      futureGoalTargetDateCalculator(today, 12),
-    ]
-
-    const view = new CreateGoalView(
-      prisonerSummary,
-      req.session.newGoal.createGoalForm,
-      futureGoalTargetDates,
-      false,
-      req.flash('errors'),
-    )
-    return res.render('pages/goal/create/newIndex', { ...view.renderArgs })
-  }
-
   submitCreateGoalForm: RequestHandler = async (req, res, next): Promise<void> => {
     const { prisonNumber, goalIndex } = req.params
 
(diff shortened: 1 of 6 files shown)


server/views/pages/goal/create/newIndex.njk
nathanrussell-moj-digital, line 1: I'm not sure about the filename `newIndex` , but it will do for now (I understand why you've named it this way)

server/views/pages/goal/create/newIndex.njk
```diff
@@ -0,0 +1,190 @@
...
+
+      <h2 class="govuk-heading-s">Add goals</h2>
+      <p class="govuk-hint">
+        Goals should be SMART, meaning 'specific', 'measurable', 'achievable', 'relevant' and 'time-bound'. SMART goals allow progress to be better measured and evaluated.<br>
+        You can add the steps to achieve the goal on the next page.
+      </p>
+
+      <form class="form" method="post" novalidate="">
+        <input type="hidden" name="_csrf" value="{{ csrfToken }}" />
+        <input type="hidden" name="prisonNumber" value="{{ form.prisonNumber }}" />
```
nathanrussell-moj-digital, line 33: The need for `stepNumber` hidden field is unclear at the moment. I understand where it has come from (the update view), and we _might_ need it?, but I'm not sure until we have designed the process and the form object a little better.
For the time being can you remove it please? (though you know we'll probably end up adding it back in again in a couple of weeks! 🤣 )

server/views/pages/goal/create/newIndex.njk
```diff
@@ -0,0 +1,190 @@
...
+                items: targetCompletionDateItems,
+                errorMessage: errors | findError('targetCompletionDate')
+              }) }}
+
+            <div class="govuk-form-group">
+              {% set titleLabel = 'Step ' + form.stepNumber %}
+              {{ govukCharacterCount({
+                name: "title",
+                id: "title",
+                value: form.title,
```
nathanrussell-moj-digital, line 132: Presumably this wants to be the title of the step, rather than the title of the goal?
At the moment both fields are using the same property in the form object

Given we've not designed the form object yet, could you change it to `form.stepTitle`, but add a todo to it to reflect a specific step's title? (ie. we might use an array in the form object, so this form field will need to index into the array)

server/routes/createGoal/createGoalController.ts
```diff
@@ -54,6 +54,36 @@ export default class CreateGoalController {
     return res.render('pages/goal/create/index', { ...view.renderArgs })
   }
 
+  // TODO: RR-734 - Create controller and view classes new create goal journey
```
nathanrussell-moj-digital, line 57: The existing `CreateGoalController` is going to get way too big (it's too big already to be honest); so we don't want these new methods etc in this class.
Can you create a new controller class please, perhaps called `CreateGoalsController` or similar, and put your 2 new methods in it please?

server/routes/createGoal/createGoalController.ts
```diff
@@ -54,6 +54,36 @@ export default class CreateGoalController {
...
+  newGetCreateGoalView: RequestHandler = async (req, res, next): Promise<void> => {
+    const { prisonNumber } = req.params
+    const { prisonerSummary } = req.session
+
+    req.session.newGoals = req.session.newGoals || []
```
nathanrussell-moj-digital, line 62: OK; so you're assuming we'll be using the existing form objects 🤔 
Thats a fair working assumption at the moment, but we do have a subtask around defining the form object(s), so this might change
Leave it as it is for the time being - it's OK as a skeleton, but we might change this as and when we define the form objects

server/routes/createGoal/createGoalController.ts
```diff
@@ -54,6 +54,36 @@ export default class CreateGoalController {
...
+    }
+
+    const today = moment().toDate()
+    const futureGoalTargetDates = [
+      futureGoalTargetDateCalculator(today, 3),
+      futureGoalTargetDateCalculator(today, 6),
+      futureGoalTargetDateCalculator(today, 12),
+    ]
+
+    const view = new CreateGoalView(
```
nathanrussell-moj-digital, line 77: I think we might want a different view model class here (not sure yet? 🤔 ; but I'm wondering if the changes we are about to get into might conflict with what we already do for the current create goal (singular) journey, and therefore this view class)

Can you put a TODO on here please, to work out whether we need a new view class, or can re-use this existing one

server/routes/createGoal/createGoalController.ts
```diff
@@ -77,6 +107,21 @@ export default class CreateGoalController {
     return res.redirect(`/plan/${prisonNumber}/goals/${goalIndex}/add-step/1`)
   }
 
+  // TODO: RR-748 - Implement submit handler for new create goal journey
+  newSubmitCreateGoalForm: RequestHandler = async (req, res, next): Promise<void> => {
+    const { prisonNumber, goalIndex } = req.params
+
+    req.session.newGoal.createGoalForm = { ...req.body }
```
nathanrussell-moj-digital, line 114: As above, this assumes we will be using the existing form object and validating it in the same way. I'm really not sure that will be the case .... 🤔 

For the time being, can you remove all this stuff, and just make this submit handler method return the res.redirect

PR #536, opened 2024-03-27
RR-743 use the correct form field values from viewModel

chrimesdev:
## Description

Use the correct form field values from viewModel, we will need to update these values to add a loop once the session data is setup properly.

### DIFF
--- a/server/views/pages/createGoals/index.njk
+++ b/server/views/pages/createGoals/index.njk
@@ -41,15 +41,15 @@
           <div class="govuk-summary-card__content">
 
               {{ govukCharacterCount({
-                name: "title",
-                id: "title",
+                name: "goals[0].title",
+                id: "goals[0].title",
                 value: form.goals[0].title,
                 maxlength: 512,
                 label: {
                   text: "Describe the goal",
                   classes: "govuk-label--s"
                 },
-                errorMessage: errors | findError('title')
+                errorMessage: errors | findError('goals[0].title')
               }) }}
 
               {% set targetCompletionDateItems = [] %}
@@ -78,21 +78,21 @@
                     {
                       id: 'another-date-day',
                       label: 'Day',
-                      name: 'targetCompletionDate-day',
+                      name: 'goals[0].targetCompletionDate-day',
                       classes: 'govuk-input--width-2',
                       value: form.goals[0]['targetCompletionDate-day']
                     },
                     {
                       id: 'another-date-month',
                       label: 'Month',
-                      name: 'targetCompletionDate-month',
+                      name: 'goals[0].targetCompletionDate-month',
                       classes: 'govuk-input--width-2',
                       value: form.goals[0]['targetCompletionDate-month']
                     },
                     {
                       id: 'another-date-year',
                       label: 'Year',
-                      name: 'targetCompletionDate-year',
+                      name: 'goals[0].targetCompletionDate-year',
                       classes: 'govuk-input--width-4',
                       value: form.goals[0]['targetCompletionDate-year']
                     }
@@ -110,8 +110,8 @@
               %}
 
               {{ govukRadios({
-                name: "targetCompletionDate",
-                id: "targetCompletionDate",
+                name: "goals[0].targetCompletionDate",
+                id: "goals[0].targetCompletionDate",
                 value: form.goals[0].targetCompletionDate,
                 fieldset: {
                   legend: {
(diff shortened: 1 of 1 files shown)


server/views/pages/createGoals/index.njk
```diff
@@ -171,15 +172,17 @@
...
               </div>
             </div>
 
-
           </div>
         </div>
 
         {{ govukButton({
+          id: "submit-button",
+          name: "submit-button",
```
nathanrussell-moj-digital, line 184: I think this needs to be `name: "action"` so that in the controller code (yet to be written) we will know the difference between when the user presses the main Submit CTA (this button) vs. add another step vs. add another goal (not being added in this story, but will come in a subsequent story)

server/views/pages/createGoals/index.njk
```diff
@@ -43,7 +43,7 @@
               {{ govukCharacterCount({
                 name: "title",
```
nathanrussell-moj-digital, line 44: All of the `name` and `id` fields need changing to reflect _which_ goal's / step's field it is
I think you want something like:
```
  name: "goals[0].title",
  id: "goals[0].title",
  value: form.goals[0].title,
```
And use the same pattern/approach for all the other fields

server/views/pages/createGoals/index.njk
```diff
@@ -123,28 +123,29 @@
...
-                hint: {
-                  text: "Describe this step"
-                },
-                errorMessage: errors | findError('stepTitle')
-              }) }}
-
-            </div>
+            {% for step in form.steps %}
+              <div class="govuk-form-group">
+                <!-- TODO: Populate the stepNumber object to relect the correct number on the UI -->
```
nathanrussell-moj-digital, line 128: Can we not resolve these todo's by using the loop index?

server/views/pages/createGoals/index.njk
```diff
@@ -123,28 +123,29 @@
...
+                {% set stepTitleLabel = 'Step ' + form.stepNumber %}
+                <!-- TODO: Populate the stepTitle object to reflect the correct step title on the UI -->
+                {{ govukCharacterCount({
+                  name: "stepTitle",
+                  id: "stepTitle",
+                  value: form.goals[0].steps[0].title,
```
nathanrussell-moj-digital, line 134: As with my comment above re: the `name` and `id` fields, you'll need to do it here as well
But we need to use the zero based loop index for the `step` index
Something like this I think:
```
                  name: "goals[0].step[" + loop.index0 + "].title",
                  id: "goals[0].step[" + loop.index0 + "].title",
                  value: {{ form.goals[0].steps[loop.index0].title }},
```

Also, even though I've referred to it as `loop` here, it might be worth setting the loop variable as something like `stepLoop` or similar, because in the next set of stories when we do "add another goal", then we'll have a nested loop (ie. we'll be looping through the goals, and within that loop we'll be looping through the steps)

server/views/pages/createGoals/index.njk
```diff
@@ -123,28 +123,29 @@
...
+                  rows: 2,
+                  label: {
+                    text: stepTitleLabel,
+                    classes: "govuk-label--s",
+                    isPageHeading: false
+                  },
+                  hint: {
+                    text: "Describe this step"
+                  },
+                  errorMessage: errors | findError('stepTitle')
```
nathanrussell-moj-digital, line 145: `findError` will need to reference the `id` of _this_ step title field, so will want to be something like 
```
errorMessage: errors | findError('goals[0].step[' + loop.index0 + '].title')
```

server/views/pages/createGoals/index.njk
```diff
@@ -171,15 +172,17 @@
                   </strong>
                 </div>
                 {{ errors | findError('note') }}
-                <textarea class="govuk-textarea" id="note" name="note" rows="5" aria-describedby="warning-text">{{ form.note }}</textarea>
+                <textarea class="govuk-textarea" id="note" name="note" rows="5" aria-describedby="warning-text">{{ form.goals[0].note }}</textarea>
```
nathanrussell-moj-digital, line 175: As above, the `name`, `id` and `findErrors` will need to be something like `goal[0].note`

server/views/pages/createGoals/index.njk
chrimesdev: Currently we have no steps to loop through in the session, so the loop isn't actually working - I've removed it for now and just put placeholders in until we have some session data

nathanrussell-moj-digital (approved):
LGTM

PR #544, opened 2024-04-02
RR-480 - Reverted previous changes to Store classes

nathanrussell-moj-digital:
This PR reverts the previous changes to Store classes, making the Store classes consistent with HMPPS typescript template


server/data/cache/prisonRegisterStore.ts
nathanrussell-moj-digital, line 1: The following 4 files have not really been deleted; they've been moved as part of the refactoring

server/data/prisonRegisterStore/prisonRegisterStore.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  private async ensureConnected() {
+    if (!this.client.isOpen) {
+      await this.client.connect()
+    }
+  }
+
+  async setActivePrisons(activePrisons: Array<PrisonResponse>, durationDays = 1): Promise<string> {
+    await this.ensureConnected()
+    return this.client.set(ACTIVE_PRISONS, JSON.stringify(activePrisons), { EX: durationDays * 24 * 60 * 60 })
```
nathanrussell-moj-digital, line 22: Re: the comment in the Prisoner Profile PR, this is what we want to do I think?

server/data/tokenStore/inMemoryTokenStore.ts
```diff
@@ -0,0 +1,17 @@
...
+  map = new Map<string, { token: string; expiry: Date }>()
+
+  public async setToken(key: string, token: string, durationSeconds: number): Promise<void> {
+    this.map.set(key, { token, expiry: new Date(Date.now() + durationSeconds * 1000) })
+    return Promise.resolve()
+  }
+
+  public async getToken(key: string): Promise<string> {
+    if (!this.map.has(key) || this.map.get(key).expiry.getTime() < Date.now()) {
+      return Promise.resolve(null)
```
leejacobson-moj, line 13: I appreciate this isn't your code and is in the template, but wrapping the return value in the `Promise.resolve()` is unnecessary here since `async` will do that automatically.

leejacobson-moj (approved):
Looks good. Approved.

server/data/tokenStore/inMemoryTokenStore.ts
nathanrussell-moj-digital: Yeah; agreed.
I think my previous refactorings were around improving various bits of code (not sure whether I made this change or not in my previous version); but I think what I've seen here is that there are some classes etc that we shouldn't touch too much because it will deviate from the template too much, which in turn makes updating the project from the template hard!
I think perhaps a better approach is that we should look to make improvements in the template project, then let them filter down into the various projects over time. Thats perhaps the better way of doing what I was trying to do in my previous refactorings 😁

PR #546, opened 2024-04-03
RR-748 - Fix routes and basic controller with tests

nathanrussell-moj-digital:
This PR continues the work that Adam started re: the new create goals journey. Specifically it:

* Corrects the route for the new form / controller endpoints.
Adam had assumed the route would be the same as the existing journey, where the goal number and step number are path elements. This is not true as it's now a single page journey where each new goal and step are added to the same page (it will still be a full page request/response model; it's not being done with client side javascript / SPA). The correct route is `/plan/:prisonNumber/goals/create`
* Adds a unit test class for the controller that Adam started
Unit tests for the GET handler method, and form validation on the POST submit handler method.
Further tests will be added as we write more functionality into the controller.
* Adds `date-fns` as a dependency
To date we have used `moment`, but we have a ticket to replace `moment` with `date-fns`. Rather than migrate as a big bang switch over we'll migrate over time. Adding the library is the first step in this.
* Changes the `futureGoalTargetDate` calculator and it's test from using `moment` to use `date-fns` instead.

### DIFF
--- a/server/routes/createGoal/createGoalsController.ts
+++ b/server/routes/createGoal/createGoalsController.ts
@@ -1,6 +1,6 @@
 import type { RequestHandler } from 'express'
-import moment from 'moment'
-import type { CreateGoalsForm } from 'forms'
+import { UTCDate } from '@date-fns/utc'
+import { startOfDay } from 'date-fns'
 import CreateGoalsView from './createGoalsView'
 import futureGoalTargetDateCalculator from '../futureGoalTargetDateCalculator'
 import validateCreateGoalsForm from './createGoalsFormValidator'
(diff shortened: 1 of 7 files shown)


server/routes/createGoal/createGoalsController.ts
```diff
@@ -12,44 +12,37 @@ export default class CreateGoalsController {
     const { prisonNumber } = req.params
     const { prisonerSummary } = req.session
 
-    req.session.createGoalsForm = req.session.createGoalsForm || []
```
nathanrussell-moj-digital, line 15: This needed quite a bit of refactoring as Adam had copied most of it from the current create goal (singular) controller which works in quite a different way (it uses arrays of forms on the session)

server/routes/createGoal/createGoalsController.ts
```diff
@@ -12,44 +12,37 @@ export default class CreateGoalsController {
...
+      goals: [
+        {
+          steps: [],
+        },
+      ],
     }
+    req.session.createGoalsForm = undefined
 
-    const today = moment().toDate()
+    const today = startOfDay(new UTCDate())
```
leejacobson-moj, line 25:
```suggestion
    const today = startOfToday()
```

leejacobson-moj: date-fns has a `startOfToday` method you could use here instead, https://date-fns.org/v3.6.0/docs/startOfToday

leejacobson-moj (dismissed):
One small suggestion if you want to update, otherwise looks good. Approved.

server/routes/createGoal/createGoalsController.ts
nathanrussell-moj-digital: Done 👍

leejacobson-moj (approved):

PR #550, opened 2024-04-04
RR-734 - Initial cypress tests for new Create Goal journey

nathanrussell-moj-digital:
This PR adds the initial cypress tests for the new Create Goal journey/screen.

The original journey that we are replacing is spread over a number of screens, so there are existing cypress tests for those screens. However, because the feature toggle is now enabled in `feature.env` those cypress tests are failing (because the old screens / way of doing it has been replaced by the new way). Therefore those cypress tests have been disabled for now, and ultimately will be deleted when the feature is complete and goes live.


integration_tests/e2e/goal/createGoals.cy.ts
```diff
@@ -0,0 +1,57 @@
+import Page from '../../pages/page'
+import CreateGoalsPage from '../../pages/goal/CreateGoalsPage'
+import OverviewPage from '../../pages/overview/OverviewPage'
+
+context('Create goals', () => {
```
nathanrussell-moj-digital, line 5: This is the new cypress test for the new create goals screen.
At the moment it only has 2 scenarios because that is as much of the functionality that we've written this far:
1. Navigate to the page directly via its URL
1. Click to create a Goal from the Overview page, and partially complete the form, triggering a validation error

integration_tests/e2e/goal/review.cy.ts
```diff
@@ -106,7 +108,7 @@ context('Review goal(s)', () => {
     cy.visit(`/plan/${prisonNumber}/view/overview`)
     const overviewPage = Page.verifyOnPage(OverviewPage)
 
-    let createGoalPage = overviewPage.clickAddGoalButton()
+    const createGoalPage = overviewPage.clickAddGoalButton()
```
nathanrussell-moj-digital, line 111: I had to change this to a `const`, and comment the lines out below because of a compilation error (the previous page class was `CreateGoalPage` and the new one is `CreateGoalsPage` (plural), and the methods on those 2 page classes are different at the moment
It doesnt really matter though, because the whole cypress test is disabled, so the whole thing will be deleted at some point soon 👍

integration_tests/e2e/overview/postInductionOverview.cy.ts
```diff
@@ -132,6 +132,6 @@ context('Prisoner Overview page - Post Induction', () => {
     overviewPage.clickAddGoalButton()
 
     // Then
-    Page.verifyOnPage(CreateGoalPage)
+    Page.verifyOnPage(CreateGoalsPage)
```
nathanrussell-moj-digital, line 135: By clicking the Add Goal button on the Overview page, we arrive on the new Create Goals page, which has a different page class (CreateGoalsPage) (plural)

integration_tests/pages/page.ts
```diff
@@ -50,12 +50,12 @@ export default abstract class Page {
...
 
   hasErrorCount(expected: number) {
-    cy.get('.govuk-error-summary__list').should('have.length', expected)
+    cy.get('.govuk-error-summary__list li').should('have.length', expected)
     return this
   }
 
   hasFieldInError(field: string) {
-    cy.get(`#${field}-error`).should('exist')
+    cy.get(`#${Cypress.$.escapeSelector(field)}-error`).should('exist')
```
nathanrussell-moj-digital, line 58: We need to use the cypress method to escape the selectors, because the `id`s of the elements that our new test are looking for contain "special" characters (the `[` and `]`)
EG:
```
    createGoalPage //
      .hasErrorCount(2)
      .hasFieldInError('goals[0].title')
      .hasFieldInError('goals[0].steps[0].title')
```
in the cypress test above ☝️

server/routes/createGoal/createGoalsController.ts
```diff
@@ -35,8 +37,17 @@ export default class CreateGoalsController {
   // TODO: RR-748 - Implement submit handler for new create goal journey
   submitCreateGoalsForm: RequestHandler = async (req, res, next): Promise<void> => {
     const { prisonNumber } = req.params
-    req.session.createGoalsForm = { ...req.body }
-    const { createGoalsForm } = req.session
+
+    const createGoalsForm = { ...req.body } as CreateGoalsForm
```
nathanrussell-moj-digital, line 41: We have to do this type of thing in the induction controllers as well - basically when we parse the request body into the form object, it may not have its arrays initialised.
IE. if the form is submitted with missing/blank step titles, given that the step title is the only field in the steps array, we end up with a goal with no steps array at all (rather than an empty array)
Equally, if the entire form were submitted having filled out not fields at all, then we end up with no goals array at all (rather than an empty array)
In both of these cases, where the array property is null/undefined rather than empty, the validator fails horribly!

server/views/pages/createGoals/index.njk
```diff
@@ -43,7 +43,7 @@
           <div class="govuk-summary-card__content">
 
             {{ govukCharacterCount({
-              name: "goals[" + goalLoop.index0 + "].title",
+              name: "goals[" + goalLoop.index0 + "][title]",
```
nathanrussell-moj-digital, line 46: Adam got a bit confused with the HTML field name syntax vs. the `id` attribute syntax, bless him 😁 
He got especially confused with the syntax for the steps (below), though to be fair that is messy!

So, our form definition / object graph is this:
```
  export interface CreateGoalsForm {
    prisonNumber: string
    goals: Array<{
      title?: string
      targetCompletionDate?: string
      'targetCompletionDate-day'?: string
      'targetCompletionDate-month'?: string
      'targetCompletionDate-year'?: string
      steps: Array<{
        title?: string
      }>
      note?: string
    }>
    action?: 'add-another-step' | 'submit-form'
  }
```
eg:
```
{
  prisonNumber: 'A1234BC',
  goals: [
    {
      title: 'Goal 1',
      targetCompletionDate: '2024-12-31',
      steps: [
        { title: 'Step 1 for Goal 1' },
        { title: 'Step 2 for Goal 1' },
      ],
      note: 'A note about goal 1'
    },
    {
      title: 'Goal 2',
      targetCompletionDate: '2025-02-28',
      steps: [
        { title: 'Step 1 for Goal 2' },
        { title: 'Step 2 for Goal 2' },
      ],
      note: 'A note about goal 2'
    },
  ],
  action: 'submit-form',
}
```

The html form field name for array indexing needs to be quite specific for the express urlencoded body-parser to bind it into the object graph properly. Unfortunately it doesnt do dot notation! 😢 

To index into a goal and reference it's `title` field (which is a field directly off of a goal array element) it needs to be:
```
goals[0][title]
```

To index into a step within a goal and reference it's `title` field (which is a field directly off of a step array element) it needs to be:
```
goals[0][steps][0][title]
```

For the `id` element we can use whatever we like (as long as it's a valid & unique ID) - dot notation is nice as we are all familiar with it and it reflects the object graph well

integration_tests/e2e/goal/createGoals.cy.ts
```diff
@@ -0,0 +1,57 @@
...
+  it('should not be able to submit Create Goal page given validation errors', () => {
+    // Given
+    const prisonNumber = 'G6115VJ'
+    cy.signIn()
+
+    cy.visit(`/plan/${prisonNumber}/view/overview`)
+    const overviewPage = Page.verifyOnPage(OverviewPage)
+    const createGoalPage = overviewPage.clickAddGoalButton()
+
+    createGoalPage //
```
leejacobson-moj, line 43: What's with the `//` here and one line 52? Is that a typo or is there a reason for it?

integration_tests/pages/page.ts
```diff
@@ -50,12 +50,12 @@ export default abstract class Page {
   }
 
   hasErrorCount(expected: number) {
-    cy.get('.govuk-error-summary__list').should('have.length', expected)
+    cy.get('.govuk-error-summary__list li').should('have.length', expected)
```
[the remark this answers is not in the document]
leejacobson-moj: Nice catch 👍

leejacobson-moj (approved):
I think you might have a couple of stray comments (`//`), but otherwise looks good. Approved.

PR #551, opened 2024-04-04
RR-760: Added HMPPS audit client and service

leejacobson-moj:


server/config.ts
```diff
@@ -33,6 +33,20 @@ export interface ApiConfig {
   agent: AgentConfig
 }
 
+const auditConfig = () => {
+  const auditEnabled = get('AUDIT_ENABLED', 'false') === 'true'
```
nathanrussell-moj-digital, line 37: Minor; you could perhaps use the `toBoolean` function that we added a while back to make this syntax a little neater / easier to read.
We use it for pretty much all of the feature toggles, like this:
```
    includePrisonTimelineEventsEnabled: toBoolean(
      get('INCLUDE_PRISON_TIMELINE_EVENTS_ENABLED', false, requiredInProduction),
    ),
```
So in your case you might be able to do:
```
const auditEnabled = toBoolean(get('AUDIT_ENABLED', false))
```

But I accept that this particular part of the config is from the template, and if we change it here in our version, the next time we merge in the template we'll see it as merge conflict noise 🤷‍♂️

nathanrussell-moj-digital (approved):
Great stuff; looks good to me 👍

nathanrussell-moj-digital (approved):
LGTM 👍

PR #553, opened 2024-04-05
RR-734 - Ability to add steps when creating goal

nathanrussell-moj-digital:
This PR continues the fine work Adam started re: the new screen/journey for creating goal(s); but implementing the code behind the "Add another step" button

### DIFF
--- a/integration_tests/e2e/goal/createGoals.cy.ts
+++ b/integration_tests/e2e/goal/createGoals.cy.ts
@@ -17,6 +17,7 @@ context('Create goals', () => {
     cy.task('getActionPlan')
     cy.task('stubLearnerProfile')
     cy.task('stubLearnerEducation')
+    cy.task('stubGetAllPrisons')
   })
 
   it('should be able to navigate directly to Create Goal page', () => {
@@ -43,6 +44,7 @@ context('Create goals', () => {
     createGoalPage //
       .clearGoalTitle(1)
       .setTargetCompletionDate6to12Months(1)
+      .clearStepTitle(1, 1)
 
     // When
     createGoalPage.submitPage()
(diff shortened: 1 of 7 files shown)


integration_tests/e2e/goal/createGoals.cy.ts
```diff
@@ -17,6 +17,7 @@ context('Create goals', () => {
     cy.task('getActionPlan')
     cy.task('stubLearnerProfile')
     cy.task('stubLearnerEducation')
+    cy.task('stubGetAllPrisons')
```
nathanrussell-moj-digital, line 20: Not strictly related to this PR, but this was a missing stub from this cypress test suite (resulting in some 404s but not breaking the journey as failing to lookup the prison name is handled and is acceptable)

server/@types/forms/index.d.ts
```diff
@@ -57,7 +57,7 @@ declare module 'forms' {
       }>
       note?: string
     }>
-    action?: 'add-another-step' | 'submit-form'
+    action?: string
```
nathanrussell-moj-digital, line 60: We need to make this field a little loser now in respect of it's values. This is because there is an "add another step" button on each Goal on the form, and when one of these buttons is pressed it need to carry with it not only the action ('add-another-step'), but also the goal number that you want to add the step to.
The html form in the nunjucks template (below 👇 ) does this by suffixing the goal number on the button value:
```
value: "add-another-step|" + goalLoop.index0
```
Therefore if the form had 3 goals, there would be 3 "Add another step" buttons,  where their value attributes are:
1. `value="add-another-step|0`
2.`value="add-another-step|1`
3.`value="add-another-step|2`

server/config.ts
```diff
@@ -34,7 +34,7 @@ export interface ApiConfig {
 }
 
 const auditConfig = () => {
-  const auditEnabled = toBoolean(get('AUDIT_ENABLED', 'false'))
+  const auditEnabled = toBoolean(get('AUDIT_ENABLED', false))
```
nathanrussell-moj-digital, line 37: Not related to this PR, but this was stopping the cypress tests running locally for me
More specifically it was stopping the server starting with `npm run start-feature` because `AUDIT_ENABLED` is not set in `feature.env`

Because `AUDIT_ENABLED` is not set in `feature.env`, it uses the fallback value which before my change was a string. 
The string value `'false'` doesn't work as intended in the `toBoolean` method as it's seen as a truthy value! (Man, I hate js's truthy and falsy values !)
Because it was seen as `true` the app tried to start with auditing enabled, but the rest of the required env vars are not set!

The fix is to simply pass this default value in as a `boolean` value 👍

leejacobson-moj: That's weird. It doesn't seem to do that for me when running locally. 🤔 If it fixes it though that's fine with me 🤷‍♂️

PR #554, opened 2024-04-08
RR-650 - Improve prisonService

nathanrussell-moj-digital:
This PR improves the existing `prisonService`

The reason I'm doing this improvement is because we need the same `prisonService` in DPS Prisoner Profile for the work we are doing there, and I went to simply copy and paste it.
However, on doing so I identified some areas where it was not clear what the class and it's functions were doing. Namely:
* the `getPrisonByPrisonId` and `lookupPrison` methods were essentially doing the same thing; but it was not clear which should be used (not clear why you would use one function over the other)
* lack of class/function doc that would have helped explain

The refactoring here basically:
* keeps `lookupPrison` as that was the method that all dependent code used (but renames it)
* moves the code in `getPrisonByPrisonId` into `lookupPrison` (`lookupPrison` was the only code that was actually calling `getPrisonByPrisonId` so in effect it was a private method)
* deletes the original `getPrisonByPrisonId` method
* renames `lookupPrison` to `getPrisonByPrisonId` as it is a better, more descriptive method name
* adds doc where it adds value

Several test classes needed some change based on this refactoring, and I took the opportunity to change the mock style as suggested to me in a previous PR 👍

### DIFF
--- a/server/routes/functionalSkills/functionalSkillsController.test.ts
+++ b/server/routes/functionalSkills/functionalSkillsController.test.ts
@@ -2,7 +2,8 @@ import { NextFunction, Request, Response } from 'express'
 import { SessionData } from 'express-session'
 import moment from 'moment'
 import type { Assessment, Prison } from 'viewModels'
-import { CuriousService, PrisonService } from '../../services'
+import CuriousService from '../../services/curiousService'
+import PrisonService from '../../services/prisonService'
 import FunctionalSkillsController from './functionalSkillsController'
 import aValidPrisonerSummary from '../../testsupport/prisonerSummaryTestDataBuilder'
 import {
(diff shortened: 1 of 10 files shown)


server/routes/functionalSkills/functionalSkillsController.test.ts
```diff
@@ -71,7 +66,7 @@ describe('functionalSkillsController', () => {
       const prisonerSummary = aValidPrisonerSummary(prisonNumber)
       req.session.prisonerSummary = prisonerSummary
 
-      prisonService.lookupPrison.mockImplementation(mockPrisonLookup)
+      prisonService.getPrisonByPrisonId.mockImplementation(mockPrisonLookup)
```
nathanrussell-moj-digital, line 69: The method exposed by `prisonService` and used by all dependant code is now `getPrisonByPrisonId` rather than `lookupPrison` 👍

leejacobson-moj (approved):
Looks good. Approved. 👍

PR #558, opened 2024-04-08
Frontend template updates

leejacobson-moj:
## Overview

Adds frontend updates from the template. Changes include:

- Updated fallback header.
- Minor change to how the GovUK & MOJ frontend libraries are initialised.
- Removal of `govuk_frontend_toolkit`.
- Removed unnecessary `mojFrontendInit` nunjucks block.
- Dependency version bumps.


assets/js/govukFrontendInit.js
leejacobson-moj, line 1: The initAll() calls for the govUK and MOJ frontend were previously made from scripts added to `layout.njk`.

You'll notice those init calls have now been removed from the layout and moved into these init files, which are imported instead.

package.json
```diff
@@ -5,7 +5,7 @@
   "repository": "git@github.com:ministryofjustice/hmpps-education-and-work-plan-ui.git",
   "license": "MIT",
   "scripts": {
-    "prepare": "husky install",
+    "prepare": "husky",
```
leejacobson-moj, line 8: It appears `husky install` is to be deprecated in favour of `husky`,
https://github.com/typicode/husky/releases/tag/v9.0.1

server/middleware/setUpStaticResources.ts
```diff
@@ -28,10 +28,6 @@ export default function setUpStaticResources(): Router {
     router.use('/assets', express.static(path.join(process.cwd(), dir), cacheControl))
   })
 
-  Array.of('/node_modules/govuk_frontend_toolkit/images').forEach(dir => {
```
leejacobson-moj, line 31: Govuk frontend toolkit has been depreciated, https://github.com/alphagov/govuk_frontend_toolkit

assets/js/govukFrontendInit.js
leejacobson-moj, line 1: The initAll() calls for the govUK and MOJ frontend were previously made from scripts added to layout.njk.

You'll notice those init calls have now been removed from the layout and moved into these init files, which are imported instead.

server/views/pages/prisonerList/index.njk
```diff
@@ -24,5 +24,3 @@
   </div>
 
 {% endblock %}
-
-{% block mojFrontendInit %}{% endblock %}
```
leejacobson-moj, line 28: This was used to stop the MOJ frontend JS library from initialising on the prisoner list page. This prevented the MOJ Sortable Table JS from loading and breaking the custom sort.

A better way to do this is to simply remove the `data-module=="moj-sortable-table"` attribute from the table itself, which I have done.

PR #561, opened 2024-04-10
RR-736 - Cypress test for updating an induction from long question set to short question set

nathanrussell-moj-digital:
This PR implements the cypress test for updating an induction from long question set to short question set
(this is the functionality that Matt completed before he left, but didnt get time to write this cypress test for)

### DIFF
--- a/integration_tests/pages/induction/CheckYourAnswersPage.ts
+++ b/integration_tests/pages/induction/CheckYourAnswersPage.ts
@@ -1,0 +1,15 @@
+import Page, { PageElement } from '../page'
+import WorkAndInterestsPage from '../overview/WorkAndInterestsPage'
+
+export default class CheckYourAnswersPage extends Page {
+  constructor() {
+    super('induction-check-your-answers')
+  }
+
+  submitPage(): WorkAndInterestsPage {
+    this.submitButton().click()
+    return Page.verifyOnPage(WorkAndInterestsPage)
+  }
+
+  private submitButton = (): PageElement => cy.get('#submit-button')
+}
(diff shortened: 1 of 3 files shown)


integration_tests/pages/induction/CheckYourAnswersPage.ts
```diff
@@ -0,0 +1,15 @@
...
+  constructor() {
+    super('induction-check-your-answers')
+  }
+
+  submitPage(): WorkAndInterestsPage {
+    this.submitButton().click()
+    return Page.verifyOnPage(WorkAndInterestsPage)
+  }
+
+  private submitButton = (): PageElement => cy.get('#submit-button')
```
leejacobson-moj, line 14: Not one for this PR, but since this is a common UI element it might be worth adding this to `Page` so it's available to all page types that extend it?

leejacobson-moj (approved):
Looks good. Approved. 👍

integration_tests/pages/induction/CheckYourAnswersPage.ts
nathanrussell-moj-digital: Yeah, agreed.
I'm in the middle of the next PR to do with the CIAG migration (short question set to long question set), which is already quite a big change set 😬 So I'd rather not do it in my current piece of work.

I'll add a ticket to the backlog to address the submit button duplication so we don't forget it 👍

nathanrussell-moj-digital: I have created RR-770 👍

PR #562, opened 2024-04-11
RR-736 - Implementation and cypress test for updating an induction from short question set to long question set

nathanrussell-moj-digital:
This PR implements the necessary changes to the Induction controllers and cypress test to allow updating an induction from short question set to long question set.

### DIFF
--- a/integration_tests/e2e/induction/updateInductionQuestionSet.cy.ts
+++ b/integration_tests/e2e/induction/updateInductionQuestionSet.cy.ts
@@ -16,6 +16,20 @@ import InPrisonWorkPage from '../../pages/induction/InPrisonWorkPage'
 import InPrisonWorkValue from '../../../server/enums/inPrisonWorkValue'
 import InPrisonTrainingValue from '../../../server/enums/inPrisonTrainingValue'
 import CheckYourAnswersPage from '../../pages/induction/CheckYourAnswersPage'
+import WorkedBeforePage from '../../pages/induction/WorkedBeforePage'
+import YesNoValue from '../../../server/enums/yesNoValue'
+import PreviousWorkExperienceTypesPage from '../../pages/induction/PreviousWorkExperienceTypesPage'
+import TypeOfWorkExperienceValue from '../../../server/enums/typeOfWorkExperienceValue'
+import PreviousWorkExperienceDetailPage from '../../pages/induction/PreviousWorkExperienceDetailPage'
+import FutureWorkInterestTypesPage from '../../pages/induction/FutureWorkInterestTypesPage'
+import WorkInterestTypeValue from '../../../server/enums/workInterestTypeValue'
+import FutureWorkInterestRolesPage from '../../pages/induction/FutureWorkInterestRolesPage'
+import SkillsPage from '../../pages/induction/SkillsPage'
+import SkillsValue from '../../../server/enums/skillsValue'
+import PersonalInterestsPage from '../../pages/induction/PersonalInterestsPage'
+import PersonalInterestsValue from '../../../server/enums/personalInterestsValue'
+import AffectAbilityToWorkPage from '../../pages/induction/AffectAbilityToWorkPage'
+import AbilityToWorkValue from '../../../server/enums/abilityToWorkValue'
 
 /**
  * Cypress tests that change the Induction question set by updating the answer to 'Hoping to work on release'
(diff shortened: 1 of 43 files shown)


integration_tests/e2e/induction/updateInductionQuestionSet.cy.ts
```diff
@@ -136,4 +150,145 @@ context(`Change Induction question set by updating the answer to 'Hoping to work
         ),
     )
   })
+
+  it(`should update a short question set Induction into a long question set Induction given form submitted with 'Hoping to work on release' as Yes`, () => {
```
nathanrussell-moj-digital, line 154: This is the cypress test that drives the scenario we are implementing - that of updating an existing Induction that was based on the short question set (where "do you want to work" was answered as No or Not Sure), and updating "do you want to work" to Yes.

"Do you want to work" is the key question that drives the question set. When it is answered as No or Not Sure we ask the "short" question set. Where it is answered as Yes we ask the long question set.

This scenario is the updating of that one question - by changing that one question we actually have to kick off the complete Induction question set in "long question set" mode to capture the answers to questions that were not previously asked

server/routes/induction/common/affectAbilityToWorkController.ts
```diff
@@ -21,6 +21,12 @@ export default abstract class AffectAbilityToWorkController extends InductionCon
     const affectAbilityToWorkForm = req.session.affectAbilityToWorkForm || toAffectAbilityToWorkForm(inductionDto)
     req.session.affectAbilityToWorkForm = undefined
 
+    // Check if we are in the midst of changing the main induction question set (in this case from short route to long route)
```
nathanrussell-moj-digital, line 24: All of the controllers in the `common` folder are abstract superclass controllers, essentially with the functionality for the "GET" endpoints. The reason we have them as superclasses like this is that we currently extend them with controllers for the Update functionality (which we are doing at the moment)
But soon we will look at the "Create" journey functionality, and will create new controllers for the "create" functionality. But the "GET" endpoints for the create journey will be the same, hence these abstract superclasses.

Anyway, the changes in these `common` controllers is to add handling for the Back links (the Back links are different depending on how you got to a given page - either direct from a Change link, or as part of a journey).
To manage the back link locations we have a concept of a 'page history' that we hold in session. 
The code we are adding to these controllers is to detect if we are part of an "update journey" and if so add the current page to the page history.

server/routes/induction/common/personalInterestsController.ts
```diff
@@ -31,8 +37,8 @@ export default abstract class PersonalInterestsController extends InductionContr
 
 const toPersonalInterestsForm = (inductionDto: InductionDto): PersonalInterestsForm => {
   return {
-    personalInterests: inductionDto.personalSkillsAndInterests?.interests.map(interest => interest.interestType) || [],
-    personalInterestsOther: inductionDto.personalSkillsAndInterests?.interests.find(
+    personalInterests: inductionDto.personalSkillsAndInterests?.interests?.map(interest => interest.interestType) || [],
```
nathanrussell-moj-digital, line 40: This change, and the next file, fixes a bug (effectively a NPE) where an Induction might not have the relevant data

server/routes/induction/dynamicAriaTextResolver.ts
```diff
@@ -18,6 +18,7 @@ const getDynamicBackLinkAriaText = (req: Request, backLinkUrl: string): string =
     '/plan/{PRISON_NUMBER}/view/education-and-training': `Back to ${prisonerName}'s learning and work progress`,
     '/plan/{PRISON_NUMBER}/view/work-and-interests': `Back to ${prisonerName}'s learning and work progress`,
     '/prisoners/{PRISON_NUMBER}/induction/hoping-to-work-on-release': `Back to Is ${prisonerName} hoping to get work when they're released?`,
+    '/prisoners/{PRISON_NUMBER}/induction/previous-work-experience': `Back to What type of work has ${prisonerName} done before?`,
```
nathanrussell-moj-digital, line 21: This file (`dynamicAriaTextResolver`) is used to produce the aria labels for the Back links.
Basically its a lookup of URLs to aria label.

(don't shoot the messenger about whether the aria labels actually make sense / read sensibly !! the scope of our work here is to migrate the Induction journey functionally as-is from the CIAG UI. These are the aria labels that the CIAG UI currently implements, and presumably were tested and signed off by someone? 🤷‍♂️ . I'm all up for changing these labels, but not in the scope of this "as-is" migration)

server/routes/induction/update/additionalTrainingUpdateController.test.ts
```diff
@@ -21,11 +23,8 @@ describe('additionalTrainingUpdateController', () => {
     typeof toCreateOrUpdateInductionDto
   >
 
-  const inductionService = {
-    updateInduction: jest.fn(),
-  }
-
-  const controller = new AdditionalTrainingUpdateController(inductionService as unknown as InductionService)
+  const inductionService = new InductionService(null) as jest.Mocked<InductionService>
```
nathanrussell-moj-digital, line 26: Whilst I was here I've taken the opportunity to change the mocking style of the dependent service.
This is so much neater, so thank you for the suggestion in my other PR @leejacobson-moj 👍

server/routes/induction/update/affectAbilityToWorkUpdateController.test.ts
```diff
@@ -132,6 +130,62 @@ describe('affectAbilityToWorkUpdateController', () => {
       expect(req.session.affectAbilityToWorkForm).toBeUndefined()
       expect(req.session.inductionDto).toEqual(inductionDto)
     })
+
```
nathanrussell-moj-digital, line 133: Most of the changes to the "update" controller tests will follow a similar pattern.

I've added a new test for the "get" endpoint that asserts that if this is part of an update journey (updateInductionQuestionSet is on the session) then the page flow history is correctly set.

And then I've added a 2nd test for the "submit" (POST) endpoint that asserts that the service is not called to update the induction, and instead the user is redirected to the correct next screen and the DTO of the Induction that we hold in session has been updated correctly

server/routes/induction/update/affectAbilityToWorkUpdateController.ts
```diff
@@ -18,12 +20,15 @@ export default class AffectAbilityToWorkUpdateController extends AffectAbilityTo
 
   getBackLinkUrl(req: Request): string {
     const { prisonNumber } = req.params
+    const { pageFlowHistory } = req.session
```
nathanrussell-moj-digital, line 23: Most of the changes to "update" controller themselves follow a similar pattern where I've:

Updated the methods that get the back link URL and aria text to make sure it uses the page flow history if there is one

Implement the code in the submit handler to detect if this is part of an update Question Set journey (updateInductionQuestionSet is on the session) and redirect to the next screen.
Most of the redirects are simple like this one as the next page is "known" - there are a couple of exceptions where the next page is dynamically worked out based on data in the Induction

server/routes/induction/update/highestLevelOfEducationUpdateController.ts
```diff
@@ -89,8 +89,7 @@ export default class HighestLevelOfEducationUpdateController extends HighestLeve
 
       // if there is no page history (i.e. because we're not changing the main question set), start one here before commencing the add qualification mini flow
       if (!req.session.pageFlowHistory) {
-        const pageFlowHistory = this.buildPageFlowHistory(prisonNumber)
-        req.session.pageFlowHistory = pageFlowHistory
+        req.session.pageFlowHistory = this.buildPageFlowHistory(prisonNumber)
```
nathanrussell-moj-digital, line 92: I did this tidying in all of the "update" controllers

server/routes/induction/update/inPrisonWorkUpdateController.test.ts
```diff
@@ -124,7 +122,7 @@ describe('inPrisonWorkUpdateController', () => {
       expect(req.session.inductionDto).toEqual(inductionDto)
     })
 
-    it('should get the In Prison Work view given short question set journey', async () => {
+    it('should get the In Prison Work view given there is an updateInductionQuestionSet on the session', async () => {
```
nathanrussell-moj-digital, line 125: Renamed quite a few of the tests like this - the scenario is not really about whether there is a short or long question set journey; the important factor in the tests is actually if we are doing an update of the question set, as identified by `updateInductionQuestionSet` being on the session

server/routes/induction/update/qualificationsListUpdateController.ts
```diff
@@ -126,3 +128,13 @@ const inductionWithRemovedQualification = (
     },
   }
 }
+
+const updatedInductionDtoWithDefaultedHighestLevelOfEducation = (inductionDto: InductionDto): InductionDto => {
```
nathanrussell-moj-digital, line 132: There is a horrible edge case with the current Induction question sets
If you originally do a short question set, you are asked if you want to add qualifications, but not what your highest level of education was. It's simply not asked! 🤷‍♂️ 
The long question set does answer it though.

So, in the case of what we are doing here - updating a short question set to a long question set - you can arrive on the educational qualifications screens with some missing data, specifically highest level of education.

One of the options on the highest level of education screens is "Not sure", so our safest option is to default it to Not Sure if we dont have the answer.

It needs to be handled like this, but it is a real edge case. It only happens if the original Induction was a short question set, AND they choose to add at least qualification (they dont need to in the short question set); AND they then change the "do you want to work" to YES which then kicks off this long question set journey

server/routes/induction/update/workInterestTypesUpdateController.test.ts
```diff
@@ -1,30 +1,29 @@
...
 import aValidPrisonerSummary from '../../../testsupport/prisonerSummaryTestDataBuilder'
 import { aLongQuestionSetInductionDto } from '../../../testsupport/inductionDtoTestDataBuilder'
 import validateWorkInterestTypesForm from './workInterestTypesFormValidator'
 import toCreateOrUpdateInductionDto from '../../../data/mappers/createOrUpdateInductionDtoMapper'
-import { InductionService } from '../../../services'
+import InductionService from '../../../services/inductionService'
 import { aLongQuestionSetUpdateInductionRequest } from '../../../testsupport/updateInductionRequestTestDataBuilder'
 import WorkInterestTypesUpdateController from './workInterestTypesUpdateController'
-import WorkInterestTypesValue from '../../../enums/workInterestTypeValue'
+import WorkInterestTypeValue from '../../../enums/workInterestTypeValue'
```
nathanrussell-moj-digital, line 13: Whoops! we had the wrong import name here previously!

server/views/pages/induction/checkYourAnswers/partials/_skillsAndInterests.njk
```diff
@@ -24,7 +24,7 @@
     <dd class="govuk-summary-list__value" data-qa="particularJobInterests">
       {% for item in inductionDto.futureWorkInterests.interests %}
         <p class="govuk-body"  data-qa="particularJobInterests-{{ loop.index }}">
-          {{ item.workTypeOther if item.workType === 'OTHER' else item.workType || formatJobType }}:<br>
+          {{ item.workTypeOther if item.workType === 'OTHER' else item.workType | formatJobType }}:<br>
```
nathanrussell-moj-digital, line 27: Typo in the nunjucks template with the filter operator

leejacobson-moj (approved):
Looks good. Approved 👍

server/routes/induction/dynamicAriaTextResolver.ts
nathanrussell-moj-digital: Yeah, agreed. The whole Back link thing is a bit of a mess IMHO.
In the scope of this work we are to migrate as-is, warts and all! If we make functional changes, even cosmetic/wording changes, it will be really hard during testing to determine if we have regressed something, or if the reported problem was always part of the CIAG UI

Personally I have a massive problem with Back links like this. I know they are part of GDS and all that, but technically I think they are flawed and can make the problem they try to solve worse!
From a technical point of view we need to track the user's history and previous page ... which depending on business requirements may not always be the browser page the user last saw (difficult to explain, but we have a couple of examples of it in these Induction flows, hence us having to track the state with the `pageFlowHistory` object in the session.
The 2nd problem I have is that from a browser history point of view, using these Back links does not actually move you "back" at all! They are regular anchor links, so actually move you forwards and add to the browser history!
This then becomes massively confusing for the user when they use a combination of the application Back link, and then the browser Back and Forward buttons (or worse, gestures on their laptop trackpad or magic mouse!)
It's a nightmare! 😱  

I think once we have migrated these screens (warts and all), we'll have an opportunity to re-address some of this stuff. I know there is some work coming our way to change the question set. This will involve simplifying the flow so there is no longer a "long" and "short" question set, change the order of some questions, and remove some others.
I think at that point we'll have our opportunity to ask questions about the Back links etc including text labels etc 👍

PR #566, opened 2024-04-12
RR-762: HMPPS Audit pageview events

leejacobson-moj:
## Overview

Adds support HMPPS audit pageview events using page routes.

Will raise both a "PAGE_VIEW_ATTEMPT_" and a "PAGE_VIEW_" depending on if the request is successful.

`subjectType` and `subjectId` are automatically pulled from route params and added the request. Route params & query strings are also automatically added to details object.

Ideally I would have like to have integration tests working for these, but this would require configuring LocalStack to run in CI. I did look to see if any other projects where doing that, but I couldn't find any so will require some investigation if that's what we want to do.

I also wanted to set this up so routes without an audit event will throw an error, however I couldn't find a good way to do this since we'd first need a way to know if the route exists otherwise every 404 will turn into a 500 error since unknown routes wouldn't have audit events. A potential solution here would be to throw a 404 instead, but this feels a bit hacky and potentially confusing, but can add that in a separate PR if we think it's worth doing.

The unit tests are using supertest to test with the audit middleware enabled. Again, it would have been nice to have integration tests validating each page in the service is raising an audit event, since the unit tests just check the middleware functionality is working as expected.


helm_deploy/hmpps-education-and-work-plan-ui/values.yaml
```diff
@@ -61,7 +61,9 @@ generic-service:
     UPDATE_INDUCTION_PRE_PRISON_EDUCATION_SECTION_ENABLED: "false"
     NEW_COURSE_AND_QUALIFICATION_HISTORY_ENABLED: "false"
     NEW_CREATE_GOAL_JOURNEY_ENABLED: "false"
+    AUDIT_SQS_QUEUE_URL: "sqs_queue_url"
```
nathanrussell-moj-digital, line 64: Minor, and I'm sure it works fine, but does this need to be here (line 64) as well as in the secrets (line 85)?

server/middleware/auditMiddleware.ts
```diff
@@ -0,0 +1,76 @@
...
+const pageViewEventMap: Record<string, Page> = {
+  '/': Page.PRISONER_LIST,
+  '/plan/:prisonNumber/view/overview': Page.OVERVIEW,
+  '/plan/:prisonNumber/view/support-needs': Page.SUPPORT_NEEDS,
+  '/plan/:prisonNumber/view/education-and-training': Page.EDUCATION_AND_TRAINING,
+  '/plan/:prisonNumber/view/work-and-interests': Page.WORK_AND_INTERESTS,
+  '/plan/:prisonNumber/view/timeline': Page.TIMELINE,
+  '/plan/:prisonNumber/functional-skills': Page.FUNCTIONAL_SKILLS,
+  '/plan/:prisonNumber/goals/create': Page.CREATE_GOALS,
+  '/plan/:prisonNumber/goals/:goalIndex/add-step/:stepIndex': Page.CREATE_GOAL_ADD_STEP,
```
nathanrussell-moj-digital, line 15: Could you add a todo and perhaps a bit of spacing between lines 15 and 18, because these routes will be removed soon(ish) as part of migrating to the single screen journey for creating goals. We already have a jira to remove the feature toggle and associated "old" code, so these audit routes would fall within that remit.
Perhaps something like:
```
 ...
 '/plan/:prisonNumber/functional-skills': Page.FUNCTIONAL_SKILLS,
  '/plan/:prisonNumber/goals/create': Page.CREATE_GOALS,

  // TODO - RR-769 - Remove route auditing for original Create Goal journey
  '/plan/:prisonNumber/goals/:goalIndex/add-step/:stepIndex': Page.CREATE_GOAL_ADD_STEP,
  '/plan/:prisonNumber/goals/:goalIndex/add-note': Page.CREATE_GOAL_ADD_NOTE,
  '/plan/:prisonNumber/goals/:goalIndex/create': Page.CREATE_GOAL,
  '/plan/:prisonNumber/goals/review': Page.CREATE_GOALS_REVIEW,

  '/plan/:prisonNumber/goals/:goalReference/update': Page.UPDATE_GOALS,
  '/plan/:prisonNumber/goals/:goalReference/update/review': Page.UPDATE_GOALS_REVIEW,
   ...
```

nathanrussell-moj-digital (approved):
LGTM 👍 

Re: test coverage / approach; I think your unit tests are fine. I agree it would be nice to see this functionality covered by our cypress integration tests, but agree the complexity etc of setting up localstack might not be worth it.

I have a thought though ..... (and it might not work) .... but I'm pretty sure all the AWS service calls are just HTTP requests. I wonder if we could use wiremock for this 🤔 . I wonder if we set `AUDIT_SQS_QUEUE_URL=http://localhost:9091)/hmpps-audit-sqs-queue` in `feature.env`, and then setup a simple stub.
I'm pretty sure its just a HTTP request, and probably a POST (we will know pretty quickly if we set the env var in `feature.env` but dont create the stub - when we run a cypress test we'll get an error from wiremock re: non-matched request)
Using that non-matched request error we can create a basic stub (or we can use postman and wiremock's `localhost:9091/__admin/requests/find` endpoint to get the request)
And then, using our handy wiremock extension where we can match elements of the request, we can write assertions in the cypress tests

Might be worth a little play, but I'd be happy that we merge this PR as is and look to address integration tests at some point afterwards?

nathanrussell-moj-digital (approved):
LGTM 👍

PR #570, opened 2024-04-15
Stringify audit details

leejacobson-moj:
## Overview

Audit SQS messages are being published, but not being received by the audit service. We think this could be because the details object isn't stringified.

If this is the reason an update will be needed to the template.


nathanrussell-moj-digital (approved):
LGTM 👍 
(though what do other UI services do in this respect? We're not the first UI service to send audit events are we? 😬 )

PR #571, opened 2024-04-15
RR-736 - Refactor page flow history

nathanrussell-moj-digital:
This PR does some refactoring of how we build the page flow history object for pages in the Induction flow.

Previously each controller had their own version of the method, which basically just build/initialised a new history object containing the current page path. I needed to add this same functionality to the Check Your Answers page (as a enabler for supporting the Change links), so rather than add yet another very similar copy of the method, I refactored it out into a common location.

This PR looks quite big, but its basically the same change in each controller and test

### DIFF
--- a/server/routes/pageFlowHistory.ts
+++ b/server/routes/pageFlowHistory.ts
@@ -1,9 +1,22 @@
+import { Request } from 'express'
 import type { PageFlow } from 'viewModels'
 
 /**
  * Module exposing utility functions that operate on [PageFlow] instances. This is to maintain a history of pages
  * that have been visited as part of a journey (e.g. for back links on forms).
  */
+
+/**
+ * Builds and returns a new [PageFlow] instance to be used as a Page Flow History, using the current route as the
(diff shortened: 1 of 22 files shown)


server/routes/pageFlowHistory.ts
```diff
@@ -1,10 +1,23 @@
...
 /**
  * Module exposing utility functions that operate on [PageFlow] instances. This is to maintain a history of pages
  * that have been visited as part of a journey (e.g. for back links on forms).
  */
 
+/**
+ * Builds and returns a new [PageFlow] instance to be used as a Page Flow History, using the current route as the
+ * initial page in the history.
+ */
+const buildNewPageFlowHistory = (req: Request): PageFlow => {
```
nathanrussell-moj-digital, line 13: This is the new common method, which is basically the same as all the individual ones that this PR removes 👍

leejacobson-moj (approved):
Looks good. Approved 👍

PR #572, opened 2024-04-16
RR-736 - Cypress tests for Change links for most of the Short Question Set Induction questions

nathanrussell-moj-digital:
This PR implements the code and cypress tests for (most of) the Short Question Set Induction questions.
The ones not currently implemented are the 2 to do with Education as they can be a little complicated, but I'll tackle those next.

I've also left a skeleton Cypress test for the Change links for the Long Question Set Induction, so that we can work on both Short and Long question set change links independently.

The Check Your Answers page for the Short Question Set Induction is as per the following screenshot; and I have done all of the Change links except the 2 highlighted education questions:
Screenshot 2024-04-16 at 18 17 15

### DIFF
--- a/integration_tests/e2e/induction/updateCheckYourAnswersChangeLinks.cy.ts
+++ b/integration_tests/e2e/induction/updateCheckYourAnswersChangeLinks.cy.ts
@@ -1,0 +1,188 @@
+/**
+ * Cypress tests that test the Change links on the Check Your Answers page when updating an Induction
+ */
+import Page from '../../pages/page'
+import HopingToWorkOnReleasePage from '../../pages/induction/HopingToWorkOnReleasePage'
+import ReasonsNotToGetWorkPage from '../../pages/induction/ReasonsNotToGetWorkPage'
+import QualificationsListPage from '../../pages/induction/QualificationsListPage'
+import AdditionalTrainingPage from '../../pages/induction/AdditionalTrainingPage'
+import InPrisonWorkPage from '../../pages/induction/InPrisonWorkPage'
+import InPrisonTrainingPage from '../../pages/induction/InPrisonTrainingPage'
+import CheckYourAnswersPage from '../../pages/induction/CheckYourAnswersPage'
+import HopingToGetWorkValue from '../../../server/enums/hopingToGetWorkValue'
+import ReasonNotToGetWorkValue from '../../../server/enums/reasonNotToGetWorkValue'
+import InPrisonWorkValue from '../../../server/enums/inPrisonWorkValue'
+import InPrisonTrainingValue from '../../../server/enums/inPrisonTrainingValue'
+import AdditionalTrainingValue from '../../../server/enums/additionalTrainingValue'
+import WorkedBeforePage from '../../pages/induction/WorkedBeforePage'
+import YesNoValue from '../../../server/enums/yesNoValue'
+import PreviousWorkExperienceTypesPage from '../../pages/induction/PreviousWorkExperienceTypesPage'
+import TypeOfWorkExperienceValue from '../../../server/enums/typeOfWorkExperienceValue'
+import PreviousWorkExperienceDetailPage from '../../pages/induction/PreviousWorkExperienceDetailPage'
+import FutureWorkInterestTypesPage from '../../pages/induction/FutureWorkInterestTypesPage'
+import WorkInterestTypeValue from '../../../server/enums/workInterestTypeValue'
+import FutureWorkInterestRolesPage from '../../pages/induction/FutureWorkInterestRolesPage'
+import SkillsPage from '../../pages/induction/SkillsPage'
+import SkillsValue from '../../../server/enums/skillsValue'
+import PersonalInterestsPage from '../../pages/induction/PersonalInterestsPage'
+import PersonalInterestsValue from '../../../server/enums/personalInterestsValue'
+import AffectAbilityToWorkPage from '../../pages/induction/AffectAbilityToWorkPage'
+import AbilityToWorkValue from '../../../server/enums/abilityToWorkValue'
+
+context(`Change links on the Check Your Answers page when updating an Induction`, () => {
+  const prisonNumber = 'G6115VJ'
+
+  beforeEach(() => {
+    cy.task('reset')
+    cy.task('stubSignInAsUserWithEditAuthority')
+    cy.task('stubAuthUser')
+    cy.task('stubGetHeaderComponent')
+    cy.task('stubGetFooterComponent')
+    cy.task('stubPrisonerList')
+    cy.task('stubCiagInductionList')
+    cy.task('stubActionPlansList')
+    cy.task('getPrisonerById')
+    cy.task('stubLearnerProfile')
+    cy.signIn()
+  })
+
+  it('should support all Change links on a Short Question Set Induction', () => {
+    // Given
+    const checkYourAnswersPage = checkYourAnswersPageForAShortQuestionSetInduction()
+
+    // When
+    // Change Hoping To Work On Release
+    Page.verifyOnPage(CheckYourAnswersPage)
+    checkYourAnswersPage
+      .clickHopingToWorkOnReleaseChangeLink()
+      .hasBackLinkTo(`/prisoners/${prisonNumber}/induction/check-your-answers`)
+      .selectHopingWorkOnRelease(HopingToGetWorkValue.NOT_SURE)
+      .submitPage()
+
+    // Change Reasons For Not Wanting To Work
+    Page.verifyOnPage(CheckYourAnswersPage)
+    checkYourAnswersPage
+      .clickReasonsForNotWantingToWorkChangeLink()
+      .hasBackLinkTo(`/prisoners/${prisonNumber}/induction/check-your-answers`)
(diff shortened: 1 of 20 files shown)


integration_tests/e2e/induction/updateCheckYourAnswersChangeLinks.cy.ts
```diff
@@ -0,0 +1,94 @@
...
+import ReasonNotToGetWorkValue from '../../../server/enums/reasonNotToGetWorkValue'
+import InPrisonWorkValue from '../../../server/enums/inPrisonWorkValue'
+import InPrisonTrainingValue from '../../../server/enums/inPrisonTrainingValue'
+import AdditionalTrainingValue from '../../../server/enums/additionalTrainingValue'
+
+context(`Change links on the Check Your Answers page when updating an Induction`, () => {
+  const prisonNumber = 'G6115VJ'
+
+  beforeEach(() => {
+    cy.signInAsUserWithEditAuthorityToArriveOnPrisonerListPage()
```
nathanrussell-moj-digital, line 16: I've written some custom Cypress commands (`signInAsUserWithEditAuthorityToArriveOnPrisonerListPage`, `updateShortQuestionSetInductionToArriveOnCheckYourAnswers` and `updateLongQuestionSetInductionToArriveOnCheckYourAnswers`) to reduce duplication and simplify the readability of the tests.

The implementation of the new commands is down in `integration_tests/support/commands.ts` 👇

integration_tests/e2e/induction/updateCheckYourAnswersChangeLinks.cy.ts
```diff
@@ -0,0 +1,94 @@
...
+    // Then
+    Page.verifyOnPage(CheckYourAnswersPage) //
+      .hasHopingToWorkOnRelease(HopingToGetWorkValue.NOT_SURE)
+      .hasReasonsForNotWantingToWork([ReasonNotToGetWorkValue.RETIRED, ReasonNotToGetWorkValue.NO_REASON])
+      .hasAdditionalTraining([AdditionalTrainingValue.MANUAL_HANDLING, AdditionalTrainingValue.CSCS_CARD])
+      .hasInPrisonWorkInterests([InPrisonWorkValue.PRISON_LAUNDRY, InPrisonWorkValue.PRISON_LIBRARY])
+      .hasInPrisonTrainingInterests([InPrisonTrainingValue.CATERING, InPrisonTrainingValue.NUMERACY_SKILLS])
+  })
+
+  // TODO - RR-736 Implement this test
```
nathanrussell-moj-digital, line 84: This is the Long Question Set test. Hopefully there is enough here to be able to get the gist of what needs to happen in this test 🤞

integration_tests/support/commands.ts
```diff
@@ -13,3 +40,110 @@ Cypress.Commands.add('wiremockVerify', (requestPatternBuilder: RequestPatternBui
...
+  // Factors Affecting Ability To Work is the next page. This is not asked on the short question set.
+  Page.verifyOnPage(AffectAbilityToWorkPage) //
+    .chooseAffectAbilityToWork(AbilityToWorkValue.HEALTH_ISSUES)
+    .submitPage()
+  // Arrive on Check Your Answers page
+  Page.verifyOnPage(CheckYourAnswersPage)
+})
+
+const signInWithAuthority = (authority: 'EDIT' | 'VIEW') => {
+  cy.task('reset')
```
nathanrussell-moj-digital, line 137: These 11 lines are in pretty much every Cypress test we write, so I'm hoping over time we can refactor them to start using `cy.signInAsUserWithViewAuthorityToArriveOnPrisonerListPage()` or `cy.signInAsUserWithEditAuthorityToArriveOnPrisonerListPage()`

If nothing else, other than removing noise from the top of the tests, the cypress command name makes it much clearer as to what is happening and why we are setting up these stubs etc

server/routes/induction/common/hopingToWorkOnReleaseController.ts
```diff
@@ -18,6 +18,11 @@ export default abstract class HopingToWorkOnReleaseController extends InductionC
   ): Promise<void> => {
     const { prisonerSummary, inductionDto } = req.session
 
+    if (req.session.pageFlowHistory) {
```
nathanrussell-moj-digital, line 21: Because the "Hoping to work on release" page is the first question in the Induction there was never a "previous" page, so the concept of page history for this page was not a thing.
This now changes because we have the circumstance where we DO have a history, in that the previous page might be the Check Your Answers page, hence this change

server/routes/induction/common/inductionController.ts
```diff
@@ -22,4 +22,12 @@ export default abstract class InductionController {
     const updatedPageFlowHistory = setCurrentPage(req.session.pageFlowHistory, currentPageUri)
     req.session.pageFlowHistory = updatedPageFlowHistory
   }
+
+  /**
+   * Returns `true` if the previous page in the Page Flow History was Check Your Answers
+   */
+  previousPageWasCheckYourAnswers(req: Request): boolean {
+    const { pageFlowHistory } = req.session
+    return pageFlowHistory && getPreviousPage(pageFlowHistory).endsWith('/check-your-answers')
```
nathanrussell-moj-digital, line 31: Because I've used `endsWith` I think this method will also work for checking whether the previous page was Check Your Answers for the Create journey as and when we get there (the Update and Create journeys will have different paths, but will end with the same page name/path element)

server/routes/induction/update/additionalTrainingUpdateController.test.ts
```diff
@@ -217,17 +217,17 @@ describe('additionalTrainingUpdateController', () => {
       req.session.inductionDto = inductionDto
 
       const additionalTrainingForm = {
-        additionalTraining: [AdditionalTrainingValue.FULL_UK_DRIVING_LICENCE, AdditionalTrainingValue.OTHER],
-        additionalTrainingOther: 'Beginners cookery for IT professionals',
+        additionalTraining: [AdditionalTrainingValue.HGV_LICENCE, AdditionalTrainingValue.OTHER],
```
nathanrussell-moj-digital, line 220: This test was flawed and slightly misleading before, because it didnt actually update/change anything on the page! The values that were being set in the form were the same values as in the loaded Induction, so whilst it did update the induction, it updated it with the same values.
I've made this change so that we can expect and see a changed Induction when it is sent to the service

server/routes/induction/update/hopingToWorkOnReleaseUpdateController.ts
```diff
@@ -17,12 +18,15 @@ export default class HopingToWorkOnReleaseUpdateController extends HopingToWorkO
 
   getBackLinkUrl(req: Request): string {
     const { prisonNumber } = req.params
+    const { pageFlowHistory } = req.session
```
nathanrussell-moj-digital, line 21: Similar to the changes in the superclass above (`HopingToWorkOnReleaseController`), these changes to the Back link URL and Aria Label are because previously there was never a concept of a "previous" page other than the Work And Interests tab on the PLP Overview page. That was the only page that was the "previous" page, so we could hard-code the backlink etc
Now we have the concept of a page flow history for when the previous page is Check Your Answers, so we need this change 👍

server/views/pages/induction/checkYourAnswers/partials/_educationAndTraining.njk
```diff
@@ -2,10 +2,14 @@
 
 {% if inductionDto.previousQualifications.qualifications.length %}
 
-  <h3 class="govuk-heading-s govuk-!-margin-top-2 govuk-!-margin-bottom-3">
-    Educational qualifications
-    <span><a class="gov-link govuk-!-font-weight-regular" style="float:right; color:#1d70b8" href="/prisoners/{{ prisonerSummary.prisonNumber }}/induction/qualifications" data-qa="qualificationsLink">Change <span class="govuk-visually-hidden"> there educational qualifications</span></a></span>
```
nathanrussell-moj-digital, line 7: I found another example of some nasty inline styling that we've inherited from the CIAG UI, so have corrected it in the same way that I did the other example the other week

server/views/pages/induction/checkYourAnswers/partials/_educationAndTraining.njk
```diff
@@ -50,7 +54,9 @@
     </dt>
     <dd class="govuk-summary-list__value" data-qa="additionalTraining">
       {% for item in inductionDto.previousTraining.trainingTypes %}
-        <div data-qa="additionalTraining-{{ loop.index }}">{{ item | formatAdditionalTraining }}{{ ' - ' + inductionDto.previousTraining.trainingTypeOther if item === 'OTHER' }}{{ ',' if inductionDto.previousTraining.trainingTypes.length !== loop.index }}</div>
+        <div data-qa="additionalTraining-{{ item }}">
```
nathanrussell-moj-digital, line 57: Most of the changes from here on in are essentially the same thing, in that the data-qa attribute of values being rendered on the Check Your Answers page has been changed so that it no longer uses the loop index, and instead uses the value itself.
EG. we change `data-qa=additionalTraining-1` to `data-qa=additionalTraining-HGV_LICENCE`

The reason for this is so that the cypress tests can look for the presence of a specific value using the enum rather than the rendered screen label (which might change)

leejacobson-moj (approved):
Looks good. Approved. 👍

PR #574, opened 2024-04-17
RR-734 - Enable feature toggle in all envs for new Create Goal journey

nathanrussell-moj-digital:
This PR enables the feature toggle in all envs for the new Create Goal journey. 
Now that it has been successfully tested and passed by the project team they would like us to roll it out into `prod`


leejacobson-moj (approved):
Looks good. Approved 👍

PR #576, opened 2024-04-18
RR-736 - Change links for Do You Want To Add Qualifications

nathanrussell-moj-digital:
This PR is for the Change link handling and cypress test for the educational qualifications question "Do they want to record any qualifications"

This is Change link 3 of 3 for the education questions.

There is some minor complexity in this one:
* "Do they want to record any qualifications" is only asked on the Short Question Set (I have never understood why!)
* The answer to the question is not actually saved as part of the Induction (it never was from the original CIAG UI and API that we inherited)
Therefore on the Check Your Answers page we infer the answer by looking at the qualifications - if there are qualifications then the answer is inferred as Yes (as in Yes they wanted to add qualifications). Otherwise it is inferred as No
* On the Check Your Answers page, if you change "Do they want to record any qualifications" from Yes to No we have to remove any previously entered qualifications (otherwise the inference mentioned above doesnt work)
* On the Check Your Answers page, if you change "Do they want to record any qualifications" from No to Yes we have to take you through the qualifications mini-flow in order to add some qualifications, to then take you back to Check Your Answers

### DIFF
--- a/server/routes/induction/common/wantToAddQualificationsController.ts
+++ b/server/routes/induction/common/wantToAddQualificationsController.ts
@@ -1,6 +1,7 @@
 import { NextFunction, Request, RequestHandler, Response } from 'express'
 import type { Assessment } from 'viewModels'
 import type { WantToAddQualificationsForm } from 'inductionForms'
+import type { InductionDto } from 'inductionDto'
 import InductionController from './inductionController'
 import WantToAddQualificationsView from './wantToAddQualificationsView'
 import dateComparator from '../../dateComparator'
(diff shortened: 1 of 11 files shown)


server/routes/induction/common/wantToAddQualificationsController.ts
```diff
@@ -71,8 +76,18 @@ const assessmentsGroupedByTypeSortedByDateDesc = (assessments: Array<Assessment>
   return assessmentsByType
 }
 
-const createWantToAddQualificationsForm = (): WantToAddQualificationsForm => {
+const createWantToAddQualificationsForm = (inductionDto: InductionDto): WantToAddQualificationsForm => {
```
nathanrussell-moj-digital, line 79: Needed to change this function so that the form object was correctly built, with its `wantToAddQualifications` correctly set. Because its an inferred value that we dont actually store in the Induction:
* the answer is Yes if there are already qualifications on the Induction (ie. you must have previously answered yes in order to add the qualifications)
* the answer is No if there is an empty array of qualifications on the Induction (ie. there is an array of qualifications, but it is empty. This means you have previously been asked about qualifications and you said No)
* the answer is undefined if the induction doesnt have a previousQualifications or qualification object. IE. its a new induction and you've not been asked about qualifications yet.

server/routes/pageFlowHistory.ts
```diff
@@ -83,4 +83,19 @@ const setCurrentPageIndex = (pageFlowHistory: PageFlow, currentPagePath: string)
   return pageFlowHistory
 }
 
-export { buildNewPageFlowHistory, setCurrentPage, getPreviousPage, isFirstPage, isLastPage, isPageInFlow }
+/**
+ * Return `true` if the specified page flow history contains a page URL that matches the expected page URL
+ */
+const pageFlowHistoryContains = (pageFlowHistory: PageFlow, expectedPageUrl: RegExp): boolean => {
```
nathanrussell-moj-digital, line 89: See the tests above, but I chose to use a `Regexp` object as the function arg. I think this will come in handy when we do the Create Journey , where the page route URLs will be similar.
EG, we'll be able to do:
* `/\/prisoners\/.*\/induction\/check-your-answers`
* `/\/prisoners\/.*\/create-induction\/check-your-answers`

(I guess I could have used `.endsWith`, but a regex felt more flexible 🤷‍♂️ )

leejacobson-moj (approved):
Looks good. Approved 👍

PR #577, opened 2024-04-18
RR-644 - Simplified feature toggles

nathanrussell-moj-digital:
This PR simplifies the "update induction" feature toggles.

We originally had a toggle for each "section", but now that we have pretty much completed the Update functionality it makes more sense for there to be just one feature toggle, rather than the 5 we had before

I've also added a new feature toggle for Creating and Induction, in readiness for the work we are about to do


leejacobson-moj (approved):
Looks good. Approved 👍

PR #579, opened 2024-04-19
RR-736: Long question set change answers

leejacobson-moj:
## Overview

Adds functionality for check your answers change links when updating Hoping to work on release and taking the long question set journey.

### DIFF
--- a/integration_tests/e2e/induction/updateCheckYourAnswersChangeLinks.cy.ts
+++ b/integration_tests/e2e/induction/updateCheckYourAnswersChangeLinks.cy.ts
@@ -210,7 +210,7 @@ context(`Change links on the Check Your Answers page when updating an Induction`
       .chooseAdditionalTraining(AdditionalTrainingValue.CSCS_CARD)
       .submitPage()
 
-    // Change Other Training (Yes -> No)
+    // Change Worked before (Yes -> No)
     Page.verifyOnPage(CheckYourAnswersPage)
       .hasWorkedBefore(YesNoValue.YES)
       .clickHasWorkedBeforeChangeLink()
@@ -218,8 +218,8 @@ context(`Change links on the Check Your Answers page when updating an Induction`
       .selectWorkedBefore(YesNoValue.NO)
       .submitPage()
 
-    // Change Other Training (No -> Yes)
-    // Requires entering previous work experience details
+    // Change Worked before (No -> Yes)
+    // Requires journey to enter previous work experience details
     Page.verifyOnPage(CheckYourAnswersPage)
       .hasWorkedBefore(YesNoValue.NO)
       .clickHasWorkedBeforeChangeLink()
@@ -241,7 +241,7 @@ context(`Change links on the Check Your Answers page when updating an Induction`
       .setJobDetails('Coaching and motivating customers fitness goals')
       .submitPage()
 
-    // Change type of work experience
+    // Change work experience types
     Page.verifyOnPage(CheckYourAnswersPage)
       .clickWorkExperienceTypesChangeLink()
       .hasBackLinkTo(`/prisoners/${prisonNumber}/induction/check-your-answers`)
@@ -255,7 +255,7 @@ context(`Change links on the Check Your Answers page when updating an Induction`
       .submitPage()
     Page.verifyOnPage(CheckYourAnswersPage)
 
-    // Change details of work experience
+    // Change work experience details
     Page.verifyOnPage(CheckYourAnswersPage)
       .clickWorkExperienceDetailChangeLink(TypeOfWorkExperienceValue.WAREHOUSING)
       .hasBackLinkTo(`/prisoners/${prisonNumber}/induction/check-your-answers`)
@@ -263,7 +263,7 @@ context(`Change links on the Check Your Answers page when updating an Induction`
       .setJobDetails('Organising pallets')
       .submitPage()
 
-    // Change work interested in
+    // Change work interests
     Page.verifyOnPage(CheckYourAnswersPage)
       .clickWorkInterestsChangeLink()
       .hasBackLinkTo(`/prisoners/${prisonNumber}/induction/check-your-answers`)
@@ -272,7 +272,7 @@ context(`Change links on the Check Your Answers page when updating an Induction`
       .chooseWorkInterestType(WorkInterestTypeValue.OUTDOOR)
       .submitPage()
 
-    // Change work interested in
+    // Change work interest roles
     Page.verifyOnPage(CheckYourAnswersPage)
       .clickParticularJobInterestsChangeLink()
       .hasBackLinkTo(`/prisoners/${prisonNumber}/induction/check-your-answers`)
@@ -289,7 +289,7 @@ context(`Change links on the Check Your Answers page when updating an Induction`
       .chooseSkill(SkillsValue.RESILIENCE)
       .submitPage()
 
-    // Change interests
+    // Change personal interests
     Page.verifyOnPage(CheckYourAnswersPage)
       .clickPersonalInterestsChangeLink()
       .hasBackLinkTo(`/prisoners/${prisonNumber}/induction/check-your-answers`)
@@ -298,6 +298,7 @@ context(`Change links on the Check Your Answers page when updating an Induction`
       .choosePersonalInterest(PersonalInterestsValue.DIGITAL)
       .submitPage()
 
+    // Change affecting ability to work
     Page.verifyOnPage(CheckYourAnswersPage)
       .clickFactorsAffectingAbilityToWorkChangeLink()
       .hasBackLinkTo(`/prisoners/${prisonNumber}/induction/check-your-answers`)
--- a/integration_tests/e2e/induction/updateCheckYourAnswersChangeLinks.cy.ts
+++ b/integration_tests/e2e/induction/updateCheckYourAnswersChangeLinks.cy.ts
@@ -17,6 +17,10 @@ import YesNoValue from '../../../server/enums/yesNoValue'
 import PreviousWorkExperienceTypesPage from '../../pages/induction/PreviousWorkExperienceTypesPage'
 import TypeOfWorkExperienceValue from '../../../server/enums/typeOfWorkExperienceValue'
 import PreviousWorkExperienceDetailPage from '../../pages/induction/PreviousWorkExperienceDetailPage'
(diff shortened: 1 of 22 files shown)


server/@types/express/index.d.ts
```diff
@@ -26,6 +26,7 @@ import type {
   WorkInterestRolesForm,
   WorkInterestTypesForm,
 } from 'inductionForms'
+import type { InductionDto } from 'inductionDto'
```
leejacobson-moj, line 29: This type is used on `inductionDto` session property and VSCode needed this import to resolve it.

integration_tests/e2e/induction/updateCheckYourAnswersChangeLinks.cy.ts
```diff
@@ -147,32 +154,26 @@ context(`Change links on the Check Your Answers page when updating an Induction`
       .hasEducationalQualifications(['Physics'])
   })
 
-  /* TODO - RR-736 finish the implementation of this test by clicking all Change links:
```
leejacobson-moj, line 150: Test for all of these should all be there now so have removed the comment.

server/routes/induction/update/previousWorkExperienceTypesUpdateController.ts
```diff
@@ -110,22 +110,36 @@ export default class PreviousWorkExperienceTypesUpdateController extends Previou
...
-      ),
-    ]
+    const nextPages = previousWorkExperienceTypes.map(
+      workType => `/prisoners/${prisonNumber}/induction/previous-work-experience/${workType.toLowerCase()}`,
+    )
+
+    // If not coming from check answers and on an existing page flow, append to the current flow
+    if (currentPageFlow && !fromCheckAnswers) {
+      return appendPagesFromCurrentPage(currentPageFlow, nextPages)
+    }
```
leejacobson-moj, line 137: If the user selects "yes" from the Worked Before page, then a flow is created which redirects them to this page before redirecting back to check your answers.

However, when the flow is created on the previous controller we don't know what work experience type will be selected on this page, so can't know the complete flow. This checks if there is an existing flow and if there is injects the needed work experience details page into the existing flow from the current page.

server/routes/induction/update/previousWorkExperienceTypesUpdateController.ts
```diff
@@ -110,22 +110,36 @@ export default class PreviousWorkExperienceTypesUpdateController extends Previou
...
+    const pageUrls = [`/prisoners/${prisonNumber}/induction/previous-work-experience`, ...nextPages]
+    // If coming from check answers, redirect back at end of journey flow
+    if (fromCheckAnswers) {
+      pageUrls.push(`/prisoners/${prisonNumber}/induction/check-your-answers`)
```
leejacobson-moj, line 141: If the user got here from the check answers page change link instead of the "Worked Before" page, then add redirect back to the check your answers page at the end of the flow.

server/routes/induction/update/workedBeforeUpdateController.ts
```diff
@@ -51,6 +53,20 @@ export default class WorkedBeforeUpdateController extends WorkedBeforeController
 
     const updatedInduction = this.updatedInductionDtoWithHasWorkedBefore(inductionDto, workedBeforeForm)
 
+    // If the previous page was Check Your Answers, decide whether to redirect back check answers on submission
+    if (this.previousPageWasCheckYourAnswers(req)) {
+      req.session.inductionDto = updatedInduction
+      // Not worked before is selected, redirect to Check Your Answers
+      req.session.workedBeforeForm = undefined
+      if (workedBeforeForm.hasWorkedBefore === YesNoValue.NO) {
```
leejacobson-moj, line 61: If coming from the check answers page and the user selects no here we can just redirect them straight back to the check answers page.

server/views/pages/induction/checkYourAnswers/partials/_abilityToWork.njk
```diff
@@ -7,7 +7,7 @@
     </dt>
     <dd class="govuk-summary-list__value" data-qa="abilityToWork">
       {% for item in inductionDto.workOnRelease.affectAbilityToWork %}
-        <div data-qa="skills-{{ loop.index }}">{{ item | formatAbilityToWorkConstraint }}{{ ' - ' + inductionDto.workOnRelease.affectAbilityToWorkOther if item === 'OTHER' }}{{ ',' if inductionDto.workOnRelease.affectAbilityToWork.length !== loop.index }}</div>
+        <div data-qa="affectingAbilityToWork-{{ item }}">{{ item | formatAbilityToWorkConstraint }}{{ ' - ' + inductionDto.workOnRelease.affectAbilityToWorkOther if item === 'OTHER' }}{{ ',' if inductionDto.workOnRelease.affectAbilityToWork.length !== loop.index }}</div>
```
leejacobson-moj, line 10: In line with your previous change, I've updated some of these loop indexes to use keys for testing.

integration_tests/e2e/induction/updateCheckYourAnswersChangeLinks.cy.ts
```diff
@@ -147,32 +154,26 @@ context(`Change links on the Check Your Answers page when updating an Induction`
...
-   * Factors affecting ability to work
-   */
   it('should support all Change links on a Long Question Set Induction', () => {
     // Given
     cy.updateLongQuestionSetInductionToArriveOnCheckYourAnswers(prisonNumber)
-    const checkYourAnswersPage = Page.verifyOnPage(CheckYourAnswersPage) // eslint-disable-line @typescript-eslint/no-unused-vars
+    Page.verifyOnPage(CheckYourAnswersPage)
 
     // When
+    // Change Hoping To Work On Release (Yes only)
```
nathanrussell-moj-digital, line 163: You don't need this change (changing Hoping To Work On Release to Yes) because its's already set to Yes by virtue of `cy.updateLongQuestionSetInductionToArriveOnCheckYourAnswers(prisonNumber)` on line 159. Setting it to Yes makes it a long question set Induction, but it's already a long question set induction.

We already have the tests that change this value and change the question set from short to long (and vica versa) in `updateInductionQuestionSet.cy.ts` 👍

integration_tests/e2e/induction/updateCheckYourAnswersChangeLinks.cy.ts
```diff
@@ -197,9 +199,135 @@ context(`Change links on the Check Your Answers page when updating an Induction`
...
+      .choosePreviousWorkExperience(TypeOfWorkExperienceValue.OFFICE)
+      .choosePreviousWorkExperience(TypeOfWorkExperienceValue.SPORTS)
+      .submitPage()
+    Page.verifyOnPage(PreviousWorkExperienceDetailPage)
+      .hasBackLinkTo(`/prisoners/${prisonNumber}/induction/previous-work-experience`)
+      .setJobRole('Office junior')
+      .setJobDetails('Filing and photocopying: Sept 2000 - Dec 2009')
+      .submitPage()
+    Page.verifyOnPage(PreviousWorkExperienceDetailPage)
+      .hasBackLinkTo(`/prisoners/${prisonNumber}/induction/previous-work-experience/office`)
```
nathanrussell-moj-digital, line 239: Nice 👍

server/routes/induction/update/personalInterestsUpdateController.test.ts
```diff
@@ -358,5 +358,48 @@ describe('personalInterestsUpdateController', () => {
...
+        ],
+        currentPageIndex: 1,
+      }
+
+      req.body = {
+        personalInterests: ['KNOWLEDGE_BASED', 'OTHER'],
+        personalInterestsOther: 'Writing poetry and short stories',
+      }
+
+      // The actual implementations are fine for this test
```
nathanrussell-moj-digital, line 380: Assuming I've read this correctly .... I'm not sure about this .... 🤔 

* We are testing a class (`PersonalInterestsUpdateController` in this case)
* Because we want our test isolated to the class/method, we mock the class's dependencies (in this case the mapper and the form validator)
* Then we take the mock, and replace the method/function that the code under test (controller method) will call with the real implementation 🤔 

I don't know how well this sits with me. Coming from a java/kotlin background, the frameworks in question (mockito) don't really allow you to do this (the closest you have is a Spy, but thats used for a different reason).
Maybe it's because of that background. Perhaps this type of approach is more accepted in js/typescript worlds? 🤷‍♂️ 

I'm happy to go with this as it is; I'm no expert in these matters 😁  and am all up for learning new or different ways of doing things 👍 

But I think if what we are saying here is that we are happy with the default implementation of a given function, rather than replace it inline in the mock like this, would we be better off simply not using a mock at all for the dependency, and simply using the real thing?

Something to think about moving forwards 🤔 👍

integration_tests/e2e/induction/updateCheckYourAnswersChangeLinks.cy.ts
leejacobson-moj: If there's no value in it I can remove, but I added it to check that the redirect back to check answers works correctly.

It verifies that if Yes -> Yes it redirects straight back to check your answers rather than initiating the long question set journey – that should only happen if going from No -> Yes.

nathanrussell-moj-digital: It will slow the test down a little (as it's a page rendering and submission), but if you want to keep it thats fine.
As you're discovering the whole Induction journey and the various flow changes within it are complex, so if it helps you (and others in the future) understand it then it serves its purpose well 👍 😁

server/routes/induction/update/personalInterestsUpdateController.test.ts
leejacobson-moj: Yeah, it might be a background thing. Personally I'm not keen on mocking unless there's a reason to do so and I couldn't see any reason to use a mock here. The more mocks you use the more tests are testing the implementations of mocks rather than the code, plus it's extra code to maintain mock implementations. So for that reason I prefer limiting mocks and testing against live code where possible.

The behaviour of these mapper and validator helper methods should be predictable, and if they ever did do something the controller method doesn't expect then using a mock would only reduce coverage in catching those bugs.

Additionally, there is no alternative implementation here. These mapper and validator helper methods are always explicitly imported and implemented in the controller logic. These aren't injected as dependancies where different mappers/validators could be used depending on the implementation of the controller.

That was my thinking anyway. 🤷‍♂️

PR #580, opened 2024-04-22
RR-775 - Submit "Hoping to work on release" to "Qualifications"

nathanrussell-moj-digital:
This PR continues the Create Induction (long question set) by implementing the submission of "Hoping to work on release" through to "Qualifications"


integration_tests/e2e/induction/createLongQuestionSetInduction.cy.ts
```diff
@@ -0,0 +1,43 @@
...
+import Page from '../../pages/page'
+import HopingToGetWorkValue from '../../../server/enums/hopingToGetWorkValue'
+import QualificationsListPage from '../../pages/induction/QualificationsListPage'
+
+context('Create a long question set Induction', () => {
+  beforeEach(() => {
+    cy.signInAsUserWithEditAuthorityToArriveOnPrisonerListPage()
+  })
+
+  it('should create a long question set Induction', () => {
```
nathanrussell-moj-digital, line 11: This test will be built upon with each PR. As we add each page we will update this test to form the complete journey

server/routes/induction/update/hopingToWorkOnReleaseUpdateController.test.ts
```diff
@@ -8,7 +8,7 @@ import {
   aShortQuestionSetInductionDto,
 } from '../../../testsupport/inductionDtoTestDataBuilder'
 import HopingToGetWorkValue from '../../../enums/hopingToGetWorkValue'
-import validateHopingToWorkOnReleaseForm from './hopingToWorkOnReleaseFormValidator'
+import validateHopingToWorkOnReleaseForm from '../../validators/induction/hopingToWorkOnReleaseFormValidator'
```
nathanrussell-moj-digital, line 11: This change, and the same change in the implementation class below, is because I have moved the validator function.

So far we have concentrated on the Update journey, and all the "Update" related code including the form validators live in the `routes/induction/update` folder.

We are now doing Create, so the "Create" related code lives in `routes/induction/create`

But the validators are common - the same validator is used for both the Update and Create journeys.

So, I have moved the validator(s) for this page into `routes/validators/induction` so that both the Update and Create controllers can share them.
I propose moving all the other Induction validators as and when we touch them as part of doing the Create journey

leejacobson-moj (approved):
Looks good. Approved. 👍

PR #585, opened 2024-04-24
RR-776: Reasons not to work create induction page

leejacobson-moj:
## Overview

Adds reasons not to work page to the create induction journey.


integration_tests/e2e/induction/createShortQuestionSetInduction.cy.ts
```diff
@@ -0,0 +1,38 @@
...
+import HopingToWorkOnReleasePage from '../../pages/induction/HopingToWorkOnReleasePage'
+import ReasonsNotToGetWorkPage from '../../pages/induction/ReasonsNotToGetWorkPage'
+import ReasonNotToGetWorkValue from '../../../server/enums/reasonNotToGetWorkValue'
+
+context('Create a short question set Induction', () => {
+  beforeEach(() => {
+    cy.signInAsUserWithEditAuthorityToArriveOnPrisonerListPage()
+  })
+
+  it('should create a long question set Induction', () => {
```
nathanrussell-moj-digital, line 13: Minor copy and paste typo 😁 
Should be `should create a short question set Induction`

server/routes/induction/common/reasonsNotToGetWorkController.ts
```diff
@@ -35,11 +35,25 @@ export default abstract class ReasonsNotToGetWorkController extends InductionCon
...
+      },
+    }
+  }
 }
 
 const toReasonsNotToGetWorkForm = (inductionDto: InductionDto): ReasonsNotToGetWorkForm => {
   return {
-    reasonsNotToGetWork: inductionDto.workOnRelease.notHopingToWorkReasons,
-    reasonsNotToGetWorkOther: inductionDto.workOnRelease.notHopingToWorkOtherReason,
+    reasonsNotToGetWork: inductionDto?.workOnRelease?.notHopingToWorkReasons ?? [],
```
nathanrussell-moj-digital, line 56: Minor, but I think there will always be an induction, so I think you can remove the `?` from `inductionDto`

server/routes/induction/create/reasonsNotToGetWorkCreateController.test.ts
```diff
@@ -0,0 +1,202 @@
...
+    } as unknown as Response
+  })
+
+  describe('getReasonsNotToGetWorkView', () => {
+    it('should get the reasons not to get work create view', async () => {
+      // Given
+      const prisonerSummary = aValidPrisonerSummary()
+      req.session.prisonerSummary = prisonerSummary
+
+      const inductionDto: InductionDto = undefined
```
nathanrussell-moj-digital, line 43: I think you'll always have an induction on the session, because the previous screen (Hoping to get work) will have set it up (albeit with a small handful of properties set)

This should do it I think:
```
const inductionDto: InductionDto = {
  workOnRelease: {
    hopingToWork: HopingToGetWorkValue.NO,
  },
}
```

server/routes/induction/create/reasonsNotToGetWorkCreateController.test.ts
```diff
@@ -0,0 +1,202 @@
...
+      expect(req.session.hopingToWorkOnReleaseForm).toBeUndefined()
+      expect(req.session.inductionDto).toEqual(inductionDto)
+    })
+
+    it('should get the reasons not to get work create view with form data', async () => {
+      // Given
+      const prisonerSummary = aValidPrisonerSummary()
+      req.session.prisonerSummary = prisonerSummary
+
+      const inductionDto: InductionDto = undefined
```
nathanrussell-moj-digital, line 74: As above re: setting up in the induction on the session

server/routes/induction/create/reasonsNotToGetWorkCreateController.test.ts
```diff
@@ -0,0 +1,202 @@
...
+      expect(req.session.inductionDto).toEqual(inductionDto)
+    })
+  })
+
+  describe('submitReasonsNotToGetWorkForm', () => {
+    it('should redisplay page given form is submitted with validation errors', async () => {
+      // Given
+      const prisonerSummary = aValidPrisonerSummary()
+      req.session.prisonerSummary = prisonerSummary
+      const inductionDto = { prisonNumber } as InductionDto
```
nathanrussell-moj-digital, line 105: As above, I think your induction on the session needs the properties that would have been set on the previous screen

server/routes/induction/create/reasonsNotToGetWorkCreateController.test.ts
```diff
@@ -0,0 +1,202 @@
...
+      await controller.submitReasonsNotToGetWorkForm(req, res, next)
+
+      // Then
+      expect(res.redirect).toHaveBeenCalledWith('/prisoners/A1234BC/create-induction/want-to-add-qualifications')
+      expect(req.flash).toHaveBeenCalledTimes(0)
+      expect(req.session.hopingToWorkOnReleaseForm).toEqual(undefined)
+      expect(req.session.inductionDto).toEqual(expectedInduction)
+    })
+
+    it('should update induction DTO and redirect to qualifcations page', async () => {
```
nathanrussell-moj-digital, line 154: Ah, have you copied/based this on the update journey?
I don't think this is a valid use case for the Create journey, as there will never be existing qualifications on a new induction at this point (because we've not asked the user about them yet 😁 )

server/routes/induction/create/reasonsNotToGetWorkCreateController.ts
```diff
@@ -0,0 +1,55 @@
...
+import validateReasonsNotToGetWorkForm from '../../validators/induction/reasonsNotToGetWorkFormValidator'
+import { getPreviousPage } from '../../pageFlowHistory'
+import getDynamicBackLinkAriaText from '../dynamicAriaTextResolver'
+import { asArray } from '../../../utils/utils'
+
+/**
+ * Controller for Reasons Not To Get Work after release screen of the create induction journey.
+ */
+export default class ReasonsNotToGetWorkCreateController extends ReasonsNotToGetWorkController {
+  getBackLinkUrl(req: Request): string {
```
nathanrussell-moj-digital, line 13: Ah, this is right/appropriate for the Update journey because there might not be a pageFlowHistory on the session, but for the Create journey there is a better/neater way of doing this 👍 
You could change this method here to:
```
    const { pageFlowHistory } = req.session
    return getPreviousPage(pageFlowHistory)
  }
```

And then in the superclass (`ReasonsNotToGetWorkController`) change line 22 to:
```
if (req.session.updateInductionQuestionSet || req.session.pageFlowHistory) {
```

You'll also need to change the GET tests for this controller to provide a `pageFlowHistory` up to and including the _previous_ page:
```
      req.session.pageFlowHistory = {
        pageUrls: [
          `/prisoners/${prisonNumber}/create-induction/hoping-to-work-on-release`,
        ],
        currentPageIndex: 0,
      }
``` 

And then make an assertion that it's had "this" page appended to it:
```
      expect(req.session.pageFlowHistory).toEqual({
        pageUrls: [
          '/prisoners/A1234BC/create-induction/hoping-to-work-on-release',
          '/prisoners/A1234BC/create-induction/reasons-not-to-get-work',
        ],
        currentPageIndex: 1,
      })
```

server/routes/induction/create/reasonsNotToGetWorkCreateController.test.ts
```diff
@@ -0,0 +1,202 @@
...
+
+  beforeEach(() => {
+    jest.resetAllMocks()
+    req = {
+      session: {} as SessionData,
+      body: {},
+      user: {} as Express.User,
+      params: { prisonNumber } as Record<string, string>,
+      flash: jest.fn(),
+    } as unknown as Request
```
nathanrussell-moj-digital, line 29: I think you're going to need to set the path property in the mock request for "this" page

server/utils/utils.ts
```diff
@@ -21,3 +21,9 @@ export const initialiseName = (fullName?: string): string | null => {
   const array = fullName.split(' ')
   return `${array[0][0]}. ${array.reverse()[0]}`
 }
+
```
nathanrussell-moj-digital, line 24: If it wasn't for the fact that we already have this `utils` file in a `utils` folder 🤮  (and that it's inherited from the template) I'd propose something else 🤣 
(Theres lots of discussion on the internet that ["utils" is an antipattern](https://www.google.com/search?q=utils+is+an+antipattern) )

server/utils/utils.ts
```diff
@@ -21,3 +21,9 @@ export const initialiseName = (fullName?: string): string | null => {
...
+export function asArray(value?: string | string[]): string[] {
```
nathanrussell-moj-digital, line 25: Not withstanding my comment about "utils" (that is more about naming conventions and organising code etc), this is a neat function 👍 

This is really minor, and we don't need it, but I wonder if we could make this function even better/ more useful in the future by using generics.
Would this method signature work do you think?
```
export function asArray<T>(value?: T | T[]): T[] {
```

server/routes/induction/create/reasonsNotToGetWorkCreateController.test.ts
leejacobson-moj: Looking at the user flow I think this wrong anyway because it seems for the create journey there is another page in between this page and the qualifications page.

That said, wouldn't the logic be needed if the user returns to the page from check answers? That was my thinking anyway.

nathanrussell-moj-digital: 100% the logic will need updating when we do Check Your Answers, but we'll come back to do that once we've got the straight forward journey working first 😁 
It's complicated enough without us adding in complexity we don't need yet ! 🤣

server/routes/induction/create/reasonsNotToGetWorkCreateController.ts
leejacobson-moj: When I tried to update `getBackLinkUrl()` with your suggestion it started throwing errors since there was no page flow history. I've added some code to build a flow history on this page, but because there's no previous page this didn't fix the issue, so needed added a conditional to check if the previous page first exists, otherwise it uses a default.

I tried to build the flow history on the previous page (hoping to work on release) to fix this bug, but this then broke the long question set journey which currently seems to be building a history on the qualifications list page.

What I have now seems to work, but may not be "right" way to do it?

server/routes/induction/create/index.ts
```diff
@@ -89,5 +92,12 @@ export default (router: Router, services: Services) => {
     router.get('/prisoners/:prisonNumber/create-induction/has-worked-before', [
       workedBeforeCreateController.getWorkedBeforeView,
     ])
+
+    router.get('/prisoners/:prisonNumber/create-induction/reasons-not-to-get-work', [
+      asyncMiddleware(reasonsNotToGetWorkCreateController.getReasonsNotToGetWorkView),
```
leejacobson-moj, line 97: There is a unlikely, but possible bug where if a user session expires while creating an induction and tries to log back in the DTO held in the session object is now undefined. This then causes the application to crash when trying to access properties from the now undefined DTO.

I considered adding error handling for this in the controller itself, but rare edge case exceptions like this should probably just be caught. Currently we don't catch errors that are thrown in the controller which means any unexpected error thrown will cause the application to crash. In live this would cause pods to crash and restart when unexpected errors occur.

`asyncMiddleware` catches any controller errors and allows them to be passed on to the error handler middleware which will show an error page to the user. I think we should probably use this everywhere, but just added here for now as a suggestion.

nathanrussell-moj-digital, line 97: What does the `asyncMiddleware` do? And why do we only need it here / not need it on any of the other routes/handlers?

integration_tests/e2e/induction/createShortQuestionSetInduction.cy.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    const overviewPage = Page.verifyOnPage(OverviewPage)
+
+    // When
+    overviewPage.clickMakeProgressPlan()
+    Page.verifyOnPage(HopingToWorkOnReleasePage) //
+      .hasBackLinkTo('/plan/A00001A/view/overview')
+      .selectHopingWorkOnRelease(HopingToGetWorkValue.NO)
+      .submitPage()
+    Page.verifyOnPage(ReasonsNotToGetWorkPage)
```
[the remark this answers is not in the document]
leejacobson-moj: Yeah, noticed you doing that on one of your PRs. I've added a test for that now.

server/utils/utils.ts
leejacobson-moj: Nice suggestion. I've updated.

server/routes/induction/update/reasonsNotToGetWorkUpdateController.test.ts
nathanrussell-moj-digital, line 1: I think you can revert the changes to this file - in the case of the Update controllers they do use the page flow history and because you've reverted your previous changes to the superclass etc, then these 2 assertions should still hold true (and in these scenarios we do want to ensure the PFH is undefined)

server/routes/induction/common/inductionController.ts
```diff
@@ -28,7 +28,7 @@ export default abstract class InductionController {
    */
   previousPageWasCheckYourAnswers(req: Request): boolean {
     const { pageFlowHistory } = req.session
-    return pageFlowHistory && getPreviousPage(pageFlowHistory).endsWith('/check-your-answers')
+    return pageFlowHistory && getPreviousPage(pageFlowHistory)?.endsWith('/check-your-answers')
```
nathanrussell-moj-digital, line 31: I think you should be able to revert this change now as well

server/routes/induction/create/reasonsNotToGetWorkCreateController.test.ts
```diff
@@ -0,0 +1,190 @@
...
+      } as InductionDto
+      req.session.inductionDto = inductionDto
+
+      req.session.reasonsNotToGetWorkForm = undefined
+
+      const expectedReasonsNotToGetWorkForm: ReasonsNotToGetWorkForm = {
+        reasonsNotToGetWork: [],
+      }
+
+      req.session.pageFlowHistory = {
```
nathanrussell-moj-digital, line 58: This test shouldnt need the page flow history any more - I think you can remove this from here

server/routes/induction/create/reasonsNotToGetWorkCreateController.test.ts
```diff
@@ -0,0 +1,190 @@
...
+        },
+      } as InductionDto
+      req.session.inductionDto = inductionDto
+
+      const expectedReasonsNotToGetWorkForm: ReasonsNotToGetWorkForm = {
+        reasonsNotToGetWork: [ReasonNotToGetWorkValue.HEALTH, ReasonNotToGetWorkValue.LIMIT_THEIR_ABILITY],
+      }
+      req.session.reasonsNotToGetWorkForm = expectedReasonsNotToGetWorkForm
+
+      req.session.pageFlowHistory = {
```
nathanrussell-moj-digital, line 98: As above - I think you can remove this setup for the page flow history now 👍

nathanrussell-moj-digital (approved):
LGTM 👍 
I suspect there's a couple of minor bits we can tidy, but I think this allows us to move forward with the next screens in the short question set journey 👍

PR #587, opened 2024-04-25
RR-775 - Simplified the use of (and the need for) the page flow history on the create journey

nathanrussell-moj-digital:
This PR simplifies the use of (and the need for) the page flow history on the create journey (though some of this will be re-introduced when we do the Change links from Check Your Answers)

The Page Flow History is used to track the history of pages a user has been through in order to display the correct Back link

At the moment, without the complication of Check Your Answers, we have a fairly linear page flow:

induction-process-flow-chart

For the most part we know what the previous page was, even with the complexity of the branching for the short vs. long journey.
EG: If you are on the Qualifications List page (the 2nd page of the long question set), we know the previous page in the flow would be Do You Want To Work On Release. We can hard-code that, and don't need anything complicated to track the state (until we get to do the Change links from Check Your Answers, but that's one for another day)

The only parts of the flow that are a little bit complicated at the moment are Qualifications, and Previous Work Experience; both of which are dynamic. You are presented with these pages as many times as necessary, so for these pages we do need the Page Flow History to track the state of the pages you have seen.

For all other pages we don't need this complexity, so this PR removes it, replacing the Back link with a hardcoded page URL of the page that we know (from the flow chart) was the previous page.

Note: As mentioned, we will need to come back and change some of this when we do the Change links from Check Your Answers, but I propose we do the simple thing here and now, and build upon it when we need to 👍 

---

@leejacobson-moj Re: your PR and the comments I made about Back links and Page Flow History for Reasons Not To Get Work ... I was wrong!, and looks like I sent you off on a bit of a wild goose chase! Sorry!

The page you are doing - Reasons Not To Get Work - is page 2 in the short question flow, and in respect of the Back link is exactly the same as the Qualifications List page in this PR (from the long question set)
I think what you need to do is:

* Hardcode the back link to "Do you want to work on release", exactly as I have done in this PR for Qualifications List . You do not need to worry about a page flow history in session:
```
    return `/prisoners/${prisonNumber}/create-induction/hoping-to-work-on-release`
  }
```

* Simplify the test by removing any setup or assertions on the page flow history - you dont need it and your controller should not be using it (pretty much what you had originally! :sorry!)
* Remove the OR condition in the super class (line 22) (if you added it as per my suggestion on your PR) - you don't need it:
```
export default abstract class ReasonsNotToGetWorkController extends InductionController {
  ...
  getReasonsNotToGetWorkView: RequestHandler = async (
    ...
    // Check if we are in the midst of changing the main induction question set (i.e. from long route to short route)
    if (req.session.updateInductionQuestionSet) {
      this.addCurrentPageToHistory(req)
    }
```
(and you dont need to build a new page flow history in that class either)
* Revert your change in `inductionController` re: adding the question mark to `getPreviousPage` - you don't need that either

### DIFF
--- a/server/routes/induction/create/qualificationsListCreateController.ts
+++ b/server/routes/induction/create/qualificationsListCreateController.ts
@@ -1,16 +1,12 @@
 import { NextFunction, Request, RequestHandler, Response } from 'express'
 import type { InductionDto } from 'inductionDto'
 import QualificationsListController from '../common/qualificationsListController'
-import { buildNewPageFlowHistory, getPreviousPage } from '../../pageFlowHistory'
+import { buildNewPageFlowHistory } from '../../pageFlowHistory'
 import getDynamicBackLinkAriaText from '../dynamicAriaTextResolver'
 
 export default class QualificationsListCreateController extends QualificationsListController {
   getBackLinkUrl(req: Request): string {
     const { prisonNumber } = req.params
-    const { pageFlowHistory } = req.session
-    if (pageFlowHistory) {
-      return getPreviousPage(pageFlowHistory)
-    }
     return `/prisoners/${prisonNumber}/create-induction/hoping-to-work-on-release`
   }
 
@@ -26,9 +22,7 @@ export default class QualificationsListCreateController extends QualificationsLi
     const { prisonNumber } = req.params
     const { inductionDto } = req.session
 
-    if (!req.session.pageFlowHistory) {
-      req.session.pageFlowHistory = buildNewPageFlowHistory(req)
-    }
+    req.session.pageFlowHistory = buildNewPageFlowHistory(req)
 
     if (userClickedOnButton(req, 'addQualification')) {
       return res.redirect(`/prisoners/${prisonNumber}/create-induction/qualification-level`)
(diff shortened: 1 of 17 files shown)


server/routes/induction/create/qualificationsListCreateController.ts
```diff
@@ -1,16 +1,12 @@
 import { NextFunction, Request, RequestHandler, Response } from 'express'
 import type { InductionDto } from 'inductionDto'
 import QualificationsListController from '../common/qualificationsListController'
-import { buildNewPageFlowHistory, getPreviousPage } from '../../pageFlowHistory'
+import { buildNewPageFlowHistory } from '../../pageFlowHistory'
 import getDynamicBackLinkAriaText from '../dynamicAriaTextResolver'
 
 export default class QualificationsListCreateController extends QualificationsListController {
   getBackLinkUrl(req: Request): string {
```
nathanrussell-moj-digital, line 8: @leejacobson-moj 
You can copy this `getBackLinkUrl` method and use it in your PR for Reasons Not To Get Work; as both pages (Reasons Not To Get Work and Qualifications List) are the next page from "Do you want to work on release" depending on how you answer it. IE. both pages need to go back to "Do you want to work on release" 👍

PR #589, opened 2024-04-25
Controller error handling

leejacobson-moj:
## Overview

Adds `asyncMiddleware()` to controller methods and refactors prisoner summary setup.


server/routes/index.ts
```diff
@@ -9,10 +9,15 @@ import postInductionCreation from './postInductionCreation'
 import createInduction from './induction/create'
 import updateInduction from './induction/update'
 import inPrisonCoursesAndQualifications from './inPrisonCoursesAndQualifications'
+import retrievePrisonerSummaryIfNotInSession from './routerRequestHandlers/retrievePrisonerSummaryIfNotInSession'
 
 export default function routes(services: Services): Router {
   const router = Router()
```
leejacobson-moj, line 15: Just noting as an observation, but the way we create a single application router here then pass it down into the various routing group methods is probably not the typical way to do routing in Express.

For example, all `createGoal()` routes begin with the path, `/plan/:prisonNumber/goals` and
instead of prefixing each route in `createGoal()` with `/plan/:prisonNumber/goals` we could simply assign this path to those routes with `router.use()`. Eg, `router.use('/plan/:prisonNumber/goals', createGoal())`

Then inside `createGoal()` we would return a new create goals router which would handle the requests for that path.

So instead of having routes like, `router.get('/plan/:prisonNumber/goals/review', [asyncMiddleware(createGoalController.getReviewGoalView)])`

You would create a route like, `createGoalRouter.get('/review', [asyncMiddleware(createGoalController.getReviewGoalView)])`

Not a big deal, but just makes routing a little more flexible. Possibly something we can look at in another PR if you see value in it.

server/routes/index.ts
```diff
@@ -27,3 +32,8 @@ export default function routes(services: Services): Router {
 
   return router
 }
+
+// Setup prisoner summary session for routes with prisonNumber param
+function prisonerSummarySetup(router: Router, services: Services) {
+  router.param('prisonNumber', retrievePrisonerSummaryIfNotInSession(services.prisonerSearchService))
+}
```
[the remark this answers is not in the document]
nathanrussell-moj-digital: Yes, this makes a lot of sense 👍

server/routes/index.ts
nathanrussell-moj-digital: Not sure I completely follow; perhaps a quick chat about it some time would help.

(I'm confident that what you're saying is right though 😁 . You've probably noticed I am not an express expert so have done things in ways which might not be ideal or best practice. Very much open to seeing how to do things in this space differently 👍 )

nathanrussell-moj-digital (dismissed):
LGTM 👍

nathanrussell-moj-digital (approved):
LGTM 👍

PR #591, opened 2024-04-26
RR-775 - Use relative URLs for Change links on Check Your Answers

nathanrussell-moj-digital:
This PR is the first step in getting the Change links working on the Check Your Answers page for the Create journey.

The Check Your Answers page template is used for both Update and Create, and currently the Change link urls have the full path for the Update journey in them.

Rather than add some convoluted logic etc for the template to know what journey it's being used for, I've changed the links to use relative urls.

There is already a cypress test that exercises all of the Change links for the Update journey. This test still passes, which proves this change is OK 👍


leejacobson-moj (approved):
Looks good. Approved 👍

PR #592, opened 2024-04-29
RR-775 - Change link for Factors Affecting Ability To Work

nathanrussell-moj-digital:
This PR implements the Change link from Check Your Answers for "factors affecting ability to work"

This is the last question in the long question set flow, and is the easiest to do because no matter what the journey (create or change link) the next page is always Check Your Answers.
All the other pages we do a Change link for will need a little bit of logic to go to the correct page, but we have plenty of tests to protect us 👍


leejacobson-moj (approved):
Looks good. Approved 👍

leejacobson-moj (approved):
Looks good. Approved 👍

PR #599, opened 2024-05-01
Form handling poc

leejacobson-moj:


integration_tests/e2e/goal/createGoals.cy.ts
```diff
@@ -57,8 +57,8 @@ context('Create goals', () => {
     Page.verifyOnPage(CreateGoalsPage)
     createGoalPage //
       .hasErrorCount(2)
-      .hasFieldInError('goals[0].title')
-      .hasFieldInError('goals[0].steps[0].title')
+      .hasFieldInError('goals-0-title')
```
leejacobson-moj, line 60: Because these ids are now automatically created it's easier to implement a more consistent naming convention.

server/@types/express/index.d.ts
```diff
@@ -27,9 +27,16 @@ import type {
 } from 'inductionForms'
 import type { InductionDto } from 'inductionDto'
 import type { UserDetails } from '../../services/userService'
+import type { CreateGoalsForm } from '../../routes/createGoal/validators/GoalForm'
 
 export default {}
 
+declare module 'express-serve-static-core' {
+  interface Response {
+    redirectWithSuccess?(path: string, message: string): void
```
leejacobson-moj, line 36: While we're only showing the success banner in the one place currently including the success banner in the layout template then allowing it to be shown from any controller via the `redirectWithSuccess()` method helps enforce a consist pattern and removes the need for boilerplate session code.

server/middleware/setUpWebSecurity.ts
```diff
@@ -45,6 +45,7 @@ export default function setUpWebSecurity(): Router {
           formAction: [`'self' ${config.apis.hmppsAuth.externalUrl} ${config.dpsHomeUrl}`],
         },
       },
+      referrerPolicy: { policy: 'same-origin' },
```
leejacobson-moj, line 48: This allows express to redirect "back".

integration_tests/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "compilerOptions": {
     "target": "es5",
+    "experimentalDecorators": true,
```
nathanrussell-moj-digital, line 4: Generally speaking I'm not a fan of using (and in this case being reliant on) "experimental" features.
Do you know how the wider HMPPS dev team feel about using this, and/or what we envisage the risk of TC39 doing something different? 🤔 
https://www.typescriptlang.org/tsconfig/#experimentalDecorators

nathanrussell-moj-digital: Having said that, it seems quite a few HMPPS projects are using it, so I guess we'd be in good company 😁 
https://github.com/search?q=org%3Aministryofjustice%20experimentalDecorators&type=code

package.json
```diff
@@ -103,6 +106,8 @@
     "body-parser": "^1.20.2",
     "bunyan": "^1.8.15",
     "bunyan-format": "^0.2.1",
+    "class-transformer": "^0.5.1",
+    "class-validator": "^0.14.1",
```
nathanrussell-moj-digital, line 110: Looks like only 10 other HMPPS projects are using class-transformer and class-validator, so perhaps a small'ish community of teams doing it this way?
https://github.com/search?q=org%3Aministryofjustice+%22class-validator%22+language%3AJSON&type=code

nathanrussell-moj-digital:
Blimey; there's quite a lot to get my head round with this! 🤣 I kinda get it, but I think I'd be better off checking out the branch and stepping through it locally to properly get my head round it

I totally get what we are trying to do in respect of sliming down our controllers 👍 I guess my worry is that it's quite geeky, and if only a handful of projects are doing it this way (ie. it's not the HMPPS Digital "standard"), then I worry about handing this over to another team when we roll off. How well will they be able to support and maintain it going forwards?

Having said this, I like the general move to use middleware instead of controller logic, and in particular the success response middleware. I wonder if we split that out into a PR in it's own right we can get that implemented as a low risk improvement. Which in turn would make this PR a little smaller which might make it easier for me to grok some of it and reduce some of my concerns?

What do you think? Am I worrying too much?

integration_tests/tsconfig.json
leejacobson-moj: They're fairly popular. The risk is probably less that they're unstable and more that they're not official spec so could change later down the road.

Not excusing it, but I think part of the explanation here is simply that JS devs are more comfortable using experimental features since the ecosystem tends to move much faster than official spec.

package.json
leejacobson-moj: Yeah, it's not part of the template so just a pattern some teams have adopted. It depends if we see any value doing the same I guess 🤷‍♂️

I saw it work well in A&A, but the caveat there is that they adopted from the start. Something I realised while working on this POC is that we might need to do some refactoring of existing controller logic and Nunjucks views to support it well.

Not validating the form when an action is submitted on the create goals page is an example of something we wouldn't have done on any A&A form, but to support that here we would probably need to add separate controller handlers with different validation rules for those actions. We'd probably also want to change the how the form is generated since it currently works by allowing unvalidated form data to be added to the session.

Similarly, the fact that `targetCompletionDate` can be both a data string or an option is something we would have avoided in A&A because it's not easy to validate and transform that with simple decorators so the form itself probably needs reworking to fit the pattern well.

I was running into a lot of stuff like this where had the pattern been in place from the start it would have been relatively easy to set things up in a way that would work, but here I kept running into bugs where the changes I made would break the existing controller logic.

server/routes/createGoal/createGoalsController.ts
```diff
@@ -1,188 +1,134 @@
...
+    return res.render('pages/createGoals/index', { ...view.renderArgs })
+  }
 
-    const today = startOfToday()
-    const futureGoalTargetDates = [
-      futureGoalTargetDateCalculator(today, 3),
-      futureGoalTargetDateCalculator(today, 6),
-      futureGoalTargetDateCalculator(today, 12),
-    ]
+  submitAction: RequestHandler = async (req, res) => {
```
leejacobson-moj, line 23: Added new handler for form actions.

server/@types/express/index.d.ts
```diff
@@ -39,7 +45,6 @@ declare module 'express-session' {
     supportNeeds: PrisonerSupportNeeds
     newGoal: NewGoal // A single NewGoal representing the Goal that is currently being added
     newGoals: Array<NewGoal> // An array of NewGoal representing the Goals that have been added
-    createGoalsForm: CreateGoalsForm
```
leejacobson-moj, line 42: This session variable is no longer needed

server/routes/createGoal/createGoalsController.ts
```diff
@@ -1,188 +1,134 @@
...
-    const { prisonNumber } = req.params
+  getCreateGoalsView: RequestHandler = async (req, res): Promise<void> => {
     const { prisonerSummary } = req.session
 
-    const createGoalsForm = req.session.createGoalsForm || {
-      prisonNumber,
-      goals: [emptyGoal()],
-    }
-    req.session.createGoalsForm = undefined
+    const view = new CreateGoalsView(prisonerSummary, GoalCompleteDateOptions)
```
leejacobson-moj, line 19: I've changed the date options to an enum. This means that the content for these options is now in the Nunjucks view and content cannot be provided by the controller. The advantage is that the value can now be explicitly validated against by `class-validator`.

server/routes/createGoal/validators/GoalForm.ts
```diff
@@ -0,0 +1,69 @@
...
+  goals: Array<GoalForm>
+}
+
+/* Types for partial form submits (eg, form action submissions piror to submiting the complete form) */
+export class PartialGoalsForm {
+  @ArrayMinSize(1, { message: 'At least one step required' })
+  steps: Array<unknown>
+}
+
+export class PartialCreateGoalsForm {
```
leejacobson-moj, line 65: I've added this partial form to validate that form data is in a processable state when submitting actions. It shouldn't ever be needed, but provides some safety that the request body will have the properties expected to process the request successfully.

server/routes/createGoal/createGoalsController.ts
```diff
@@ -1,188 +1,134 @@
...
+      const result = handleAddAnotherGoal(createGoalsForm)
+      if (result) {
+        updatedForm = result
+        fieldID = `#goals-${updatedForm.goals.length - 1}-title`
+      }
+    }
 
-    const view = new CreateGoalsView(prisonerSummary, createGoalsForm, futureGoalTargetDates, req.flash('errors'))
-    return res.render('pages/createGoals/index', { ...view.renderArgs })
+    req.flash('formValues', JSON.stringify(updatedForm))
```
nathanrussell-moj-digital, line 58: Using `req.flash` for the form values might be a problem. Try this:
* Start creating a new goal for a prisoner
* Enter a description for goal 1
* Choose a date for goal 1
* Enter a step title for goal 1
* Click to add another goal
* Observe that all the field values for Goal 1 are still there and that we have an empty goal 2 👍 
* Refresh the page with Command-R
* Observe that goal 2 has gone and all the field values for goal 1 are also gone 👎 

Maybe it's not a problem? Maybe people don't refresh the page and/or accept that if they do then their field values will be gone?
But it's difference behaviour to what we currently have, and I honestly don't know if thats a problem or not? (I wonder also if there's some GDS guidance re expectations of what should happen when the user refreshes a page in this way?)

server/routes/createGoal/validators/GoalForm.test.ts
```diff
@@ -0,0 +1,112 @@
...
+  it('should validate given validators return no errors', async () => {
+    // Given
+    const body = {
+      title: 'Learn Spanish',
+      targetCompletionDateOption: GoalCompleteDateOptions.THREE_MONTHS,
+      steps: [{ title: 'Book Spanish course' }],
+    }
+
+    // When
+    const goalForm = plainToInstance(GoalForm, body)
```
nathanrussell-moj-digital, line 16: I think the CIAG UI project also used `plainToInstance` and it confused me then as it does now 😁 
I think (for me) my biggest bug-bear with it is that args appear to be the wrong way round 🤣 
The function name is saying "take a plain object and turn it into an instance of some type", so on that basis I'd expect it to read: `const goalForm = plainToInstance(body, GoalForm)` 😁 

Anyway, not to worry about that, I know it's a 3rd party library, and I'm sure more people are happy with it than just me moaning about it! 🤣 

My real question in this code is whether we need it at all (in this use case, not generally).
Could we not simply do:
```
    // Given
    const goalForm: GoalForm = {
      title: 'Learn Spanish',
      targetCompletionDateOption: GoalCompleteDateOptions.THREE_MONTHS,
      steps: [{ title: 'Book Spanish course' }],
    }

    // When
    const validationErrors = await validate(goalForm)
```

Would that work? 🤔

server/routes/createGoal/validators/GoalForm.ts
nathanrussell-moj-digital: I'm not sure I fully understand 🤔 
Why do we need these partial forms and how are they different from the main forms?
Perhaps it's the word "partial" thats confusing me? 🤔

server/views/pages/createGoals/index.njk
```diff
@@ -28,11 +19,10 @@
         You can add the steps to achieve the goal on the next page.
       </p>
 
-      <form class="form" method="post" novalidate="">
+      <form class="form" method="post" novalidate="" action="create">
         <input type="hidden" name="_csrf" value="{{ csrfToken }}" />
-        <input type="hidden" name="prisonNumber" value="{{ form.prisonNumber }}" />
 
-        {% for goals in form.goals %}
+        {% for goals in formValues.goals or range(1) %}
```
nathanrussell-moj-digital, line 25: Oooh, `or range(1)` .... I take it that's how we can ensure we iterate this loop at least once, rather than adding an empty goal to the array in the form in the controller as we do currently?
Thats neat; I like that 👍

nathanrussell-moj-digital:
I think there's some really good stuff in here, as a POC it really helps bring some of these ideas to light and demonstrate how they might work for us and what the scope of the change is 👍 

I wonder if we take some of these ideas forward and implement them piecemeal as discrete changes in their own right with their own tickets?
How about we create tickets to:
* add the success middleware
* add the error message middleware (for all forms inc. induction) (assuming we can do it without changing the way the current forms or validators work (yet))
* Simplify new goal and step handling by using `range(1)` in the view rather than adding empty array elements for the form
* Simplify goal `targetCompletionDate` by using enum (if we can do it without adding the SimpleDate class (yet))

What do you think? I'm happy to create those jira tickets if you like? And any others you think would be worth pulling out of this and doing as a standalone thing?

server/routes/createGoal/validators/GoalForm.test.ts
leejacobson-moj: It might work, but it would be validating against the untransformed body values.

`validationMiddleware.ts` will first transform the request body (which is a generic Record<string, string> type) to a GoalForm, then once transformed it runs class validator against the actual `GoalForm` object. Otherwise it would be validating against the request body which would just be strings.

It might work in this case because all the validation should work against strings, but in other examples we might be expecting date strings in the request body to first to be transformed into `Date` objects before running date validation against them.

PR #600, opened 2024-05-03
RR-779: Added "wants to add qualifications" create induction page

leejacobson-moj:
## Overview

Adds the "wants to add qualifications" page to the "short" path of the create induction journey.

### Screenshots

Screenshot at May 03 14-23-08


server/routes/induction/update/wantToAddQualificationsFormValidator.test.ts
```diff
@@ -11,7 +11,7 @@ describe('wantToAddQualificationsFormValidator', () => {
       const expected: Array<Record<string, string>> = []
 
       // When
-      const actual = validateWorkedBeforeForm({ wantToAddQualifications: YesNoValue.YES }, prisonerSummary)
```
leejacobson-moj, line 14: Looks like this import was named incorrectly before.

server/routes/induction/common/wantToAddQualificationsController.ts
```diff
@@ -27,8 +27,9 @@ export default abstract class WantToAddQualificationsController extends Inductio
 
     const functionalSkills = {
       ...prisonerFunctionalSkills,
-      assessments: mostRecentAssessments(prisonerFunctionalSkills.assessments || []),
+      assessments: mostRecentAssessments(prisonerFunctionalSkills?.assessments || []),
```
nathanrussell-moj-digital, line 30: Hmmm, I _think_ you should always have a `prisonerFunctionalSkills` object (its data that comes from Curious). I _think_ the object itself should always exist, but the `assessments` array within it might not ... 🤔 

Given that you've had to make this change, I wonder if we've forgotten something in the route setup somewhere ... let me check ....

server/routes/induction/create/wantToAddQualificationsCreateController.test.ts
```diff
@@ -0,0 +1,141 @@
...
+      // When
+      await controller.submitWantToAddQualificationsForm(req, res, next)
+
+      // Then
+      expect(res.redirect).toHaveBeenCalledWith('/prisoners/A1234BC/create-induction/want-to-add-qualifications')
+      expect(req.flash).toHaveBeenCalledWith('errors', expectedErrors)
+      expect(req.session.wantToAddQualificationsForm).toEqual(invalidWantToAddQualificationsForm)
+    })
+
+    it(`should proceed to qualification level page given user wants to add qualifications`, async () => {
```
nathanrussell-moj-digital, line 109: Is it worth having a unit test here for the other scenario? Where the user answers No?
In that case we would expect the redirect to be to the additional-training page

server/routes/induction/update/wantToAddQualificationsFormValidator.test.ts
nathanrussell-moj-digital: Good spot 👍

server/routes/induction/update/wantToAddQualificationsFormValidator.test.ts
nathanrussell-moj-digital, line 1: Do we need to move this test file into `validators/induction` so it sits alongside the implementation ?

server/routes/induction/common/wantToAddQualificationsController.ts
nathanrussell-moj-digital: Yes, I was right; instead of this change we need to add a call to a middleware, exactly as we do for the page on the update journey:
```
    router.get('/prisoners/:prisonNumber/induction/want-to-add-qualifications', [
      retrieveFunctionalSkillsIfNotInSession(services.curiousService),
      asyncMiddleware(wantToAddQualificationsUpdateController.getWantToAddQualificationsView),
    ])
```

I think if you add that middleware call to the route you've created, you can undo the changes to this file 👍

nathanrussell-moj-digital (approved):
LGTM, Nice work 👍

PR #608, opened 2024-05-10
RR-784 - Middleware to check prisoner is in users caseload

nathanrussell-moj-digital:
This PR implements the middleware to:
1. retrieve the user's caseloads from the manage users API
2. check the user is authorised to view the prisoner based on caseload and other role memberships

The first middleware (retrieve the user's caseloads) is wired in and will be called for each user request once this PR is merged.

The 2nd middleware (check the user is authorised to view the prisoner) is not wired into any routes yet

The code in this PR is based very heavily on the code in [HMPPS Prisoner Profile](https://github.com/ministryofjustice/hmpps-prisoner-profile/blob/main/server/middleware/checkPrisonerInCaseloadMiddleware.ts), though a few differences were made based on the fact that our object models and approach to utility functions is slightly different. Having said this, functionally what I have implemented here, especially the logic in `checkPrisonerInCaseloadMiddleware`, is identical to that in HMPPS Prisoner Profile

### DIFF
--- a/server/@types/viewModels/index.d.ts
+++ b/server/@types/viewModels/index.d.ts
@@ -8,6 +8,8 @@ declare module 'viewModels' {
     receptionDate?: Date
     dateOfBirth?: Date
     location: string
+    restrictedPatient: boolean
+    supportingPrisonId?: string
   }
 
   export interface PrisonerSearchSummary extends PrisonerSummary {
--- a/server/middleware/checkPrisonerInCaseloadMiddleware.test.ts
+++ b/server/middleware/checkPrisonerInCaseloadMiddleware.test.ts
@@ -1,0 +1,353 @@
+import createError from 'http-errors'
+import { Request, Response } from 'express'
+import type { SessionData } from 'express-session'
+import aValidPrisonerSummary from '../testsupport/prisonerSummaryTestDataBuilder'
+import checkPrisonerInCaseload from './checkPrisonerInCaseloadMiddleware'
+
+describe('checkPrisonerInCaseloadMiddleware', () => {
+  const prisonNumber = 'A1234BC'
+  const prisonersPrisonId = 'BXI'
+
+  let req: Request
+  let res: Response
+  const next = jest.fn()
+
+  beforeEach(() => {
+    jest.resetAllMocks()
+    req = {
+      session: { prisonerSummary: aValidPrisonerSummary(prisonNumber, prisonersPrisonId) } as SessionData,
+    } as unknown as Request
+    res = {
+      locals: {
+        user: hmppsUser(),
+      },
+    } as unknown as Response
+  })
+
+  it('should return an error given no prisoner summary in the session', async () => {
+    // Given
+    const middleware = checkPrisonerInCaseload()
+    delete req.session.prisonerSummary
+
+    const expectedError = createError(500, 'CheckPrisonerInCaseloadMiddleware: No PrisonerSummary found in session')
+
+    // When
+    await middleware(req, res, next)
+
+    // Then
+    expect(next).toHaveBeenCalledWith(expectedError)
+  })
+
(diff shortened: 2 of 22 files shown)


server/@types/viewModels/index.d.ts
```diff
@@ -8,6 +8,8 @@ declare module 'viewModels' {
     receptionDate?: Date
     dateOfBirth?: Date
     location: string
+    restrictedPatient: boolean
```
nathanrussell-moj-digital, line 11: We needed to add and populate 2 new properties in our `PrisonerSummary` view model. These are the same property names that HMPPS Prisoner Profile uses, and indeed are the same property names in the returned `Prisoner` object from `prisoner-search-api` (so mapping is easy etc 👍 )

server/middleware/checkPrisonerInCaseloadMiddleware.test.ts
```diff
@@ -0,0 +1,353 @@
+import createError from 'http-errors'
+import { Request, Response } from 'express'
+import type { SessionData } from 'express-session'
+import aValidPrisonerSummary from '../testsupport/prisonerSummaryTestDataBuilder'
+import checkPrisonerInCaseload from './checkPrisonerInCaseloadMiddleware'
+
+describe('checkPrisonerInCaseloadMiddleware', () => {
```
nathanrussell-moj-digital, line 7: Whilst HMPPS Prisoner Profile has a test for `checkPrisonerInCaseloadMiddleware`, I found it difficult to read and port into our codebase.
In the end I decided to rewrite the test from scratch to cover all scenarios / code paths.

This meant I had to fully understand all of the code paths and functionality of `checkPrisonerInCaseloadMiddleware` which is a good thing. If nothing else it means I can explain the logic that we have copied from Prisoner Profile in this respect 👍

server/middleware/checkPrisonerInCaseloadMiddleware.ts
```diff
@@ -0,0 +1,91 @@
+import createError from 'http-errors'
+import { RequestHandler } from 'express'
+import { DpsRoles } from './roleBasedAccessControl'
+import { userHasRoles } from '../utils/userRoles'
+import { isActiveCaseLoad, isInUsersCaseLoad } from '../utils/userCaseLoads'
+
```
nathanrussell-moj-digital, line 6: This middleware is heavily based on the [HMPPS Prisoner Profile](https://github.com/ministryofjustice/hmpps-prisoner-profile/blob/main/server/middleware/checkPrisonerInCaseloadMiddleware.ts) version, with the changes being based around the use of our types and imports etc.

Personally I'm not sure I would have written it this way, eg: with nested functions etc, but it is what it is, and the fact that I have gone through this line by line to work out what it does in order to write the tests, is probably a good thing.

I guess on the plus side, if we do decide to refactor this, we have the unit tests above to protect us 👍  - what ever changes we make, the tests should still pass 👍

server/utils/userCaseLoads.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Module containing utility functions to do with user caseloads
+ */
+import { User } from '../data/manageUsersApiClient'
```
nathanrussell-moj-digital, line 4: I'm not entirely sure these utility functions to do with user caseloads and user roles warrant being here as re-usable functions, because they are only used by `checkPrisonerInCaseloadMiddleware`, and I'm not sure I can see a future use case where we might need them elsewhere? 🤷‍♂️ 

But it's similar to how HMPPS Prisoner Profile has done it, so I've followed suit.
("Similar" in that they have theirs in a massive `server/utils/utils.ts` file - at least I've broken ours our into smaller modules based on a theme)

leejacobson-moj (approved):
Looks good. Approved 👍

PR #609, opened 2024-05-10
RR-785 - Fix null qualifications when creating Inductions with no qualifications

nathanrussell-moj-digital:
This PR fixes a bug with the migrated Induction process, specifically for short question set inductions where the user says "no" to wanting to record educational qualifications.

Without this bug fix the data that is sent to the API to create the Induction is:
```
{
     ....
    "prisonNumber": "G1608UL",
    "workOnRelease": {
        "hopingToWork": "NO",
        "notHopingToWorkReasons": [
            "NO_RIGHT_TO_WORK",
            "OTHER"
        ],
        ....
    },
    ....
    "previousQualifications": null,
    .... 
}
```
Notice how `previousQualifications` is `null`. Whilst the induction is saved OK by the API, when it is later retrieved for display by the UI there is effectively a NPE because `previousQualifications` is `null`.

Based on what the CIAG UI does in this scenario (short question set, No to do you want to record qualifications), we need the induction data that is sent to the API to include empty qualification data, like this:
```
{
     ....
    "prisonNumber": "G1608UL",
    "workOnRelease": {
        "hopingToWork": "NO",
        "notHopingToWorkReasons": [
            "NO_RIGHT_TO_WORK",
            "OTHER"
        ],
        ....
    },
    ....
    "previousQualifications": {
        "educationLevel": "NOT_SURE",
        "qualifications": []
    },
    .... 
}
```

### DIFF
--- a/integration_tests/e2e/induction/createChangeInductionQuestionSet.cy.ts
+++ b/integration_tests/e2e/induction/createChangeInductionQuestionSet.cy.ts
@@ -230,6 +230,7 @@ context(`Change new Induction question set by updating 'Hoping to work on releas
         .withRequestBody(
           matchingJsonPath(
             "$[?(@.workOnRelease.hopingToWork == 'YES' && " +
+              "@.previousQualifications.educationLevel == 'NOT_SURE' && " +
               '@.previousQualifications.qualifications.size() == 2 && ' +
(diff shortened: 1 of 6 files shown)


leejacobson-moj (approved):
Looks good. Approved 👍

PR #610, opened 2024-05-10
RR-786 - Fix short to long journey

nathanrussell-moj-digital:
This PR fixes a problem when updating a short question set Induction with no qualifications to a long question set Induction.
The bug was that when you changed it do long question set, you are asked about qualifications and the prisoner's Highest Level Of Education. If you selected a highest level of education that did not require exams, then the Induction was saved too early (and in a partial state).

From the jira ticket:

## Steps to reproduce
* Create a Short question set induction with no qualifications and save it. 
* On the Work & Interests tab, click the Change link for “Do you want to work on release”.
* Change the answer from No (or Not Sure) to Yes, meaning it now becomes a Long question set induction
* At the screen that asks Highest Level of Education, choose any that do not require qualifications are set (eg: Not sure, primary, or secondary before exams)

## Expected behaviour
Expected behaviour is that the user is presented with the Additional Training screen (and then onto the rest of the Long question set screens through to Check Your Answers)

## Actual behaviour
The induction is saved at that point, meaning it is effectively incomplete (now being a Long question set induction but without the answers to questions after the educational qualifications screen). The user is returned to the Work & Interests tab, where some elements of the screen are missing (because the induction is partially complete)


leejacobson-moj (approved):
Looks good. Approved 👍

PR #611, opened 2024-05-13
RR-784 - DPS view for In-Prison Courses & Qualifications

nathanrussell-moj-digital:
This PR adds the route and security setup (checking prisoner is in user's caseload) for viewing the prisoners In Prison Courses & Qualifications from a DPS link as a non-PLP user.

### This page already exists for a PLP user as follows:
Screenshot 2024-05-13 at 17 05 15

### The new view of the same page for DPS users is as follows:
Screenshot 2024-05-13 at 17 04 50


cypress.config.ts
```diff
@@ -33,7 +33,7 @@ export default defineConfig({
         reset: resetStubs,
         ...auth,
 
-        stubSignIn: () => auth.stubSignIn([]),
+        stubSignIn: (roles: []) => auth.stubSignIn(roles),
```
nathanrussell-moj-digital, line 36: The array was in the wrong place, and was preventing me passing in specific roles as an argument to the task in the new tests I've written

integration_tests/mockData/prisonerByIdData.ts
```diff
@@ -80,6 +83,97 @@ const prisoners = {
         receptionDate: '2004-04-17',
         dateOfBirth: '1976-02-28',
         cellLocation: 'B-12-55',
+        restrictedPatient: false,
+      },
+    },
+  },
+  G9981UK: {
```
nathanrussell-moj-digital, line 90: I needed to add a few more stub prisoner data records in various prisons etc; and also set the `restrictedPatient` field for the existing prisoner records (forgot to do that in my previous PR)

server/middleware/auditMiddleware.ts
```diff
@@ -41,6 +41,9 @@ const pageViewEventMap: Record<string, Page> = {
   '/prisoners/:prisonNumber/induction/qualification-level': Page.INDUCTION_UPDATE_QUALIFICATION_LEVEL,
   '/prisoners/:prisonNumber/induction/qualification-details': Page.INDUCTION_UPDATE_QUALIFICATION_DETAILS,
   '/prisoners/:prisonNumber/induction/additional-training': Page.INDUCTION_UPDATE_ADDITIONAL_TRAINING,
+
+  '/prisoner/:prisonNumber/work-and-skills/in-prison-courses-and-qualifications':
```
nathanrussell-moj-digital, line 45: New route added to the audit middleware config 👍

server/middleware/checkPrisonerInCaseloadMiddleware.test.ts
```diff
@@ -72,7 +72,7 @@ describe('checkPrisonerInCaseloadMiddleware', () => {
     // Given
     const middleware = checkPrisonerInCaseload({ allowGlobal: true })
 
-    const user = hmppsUser({ activeCaseLoadId: 'LEI', caseLoadIds: ['LEI'], userRoles: ['ROLE_GLOBAL_SEARCH'] })
+    const user = hmppsUser({ activeCaseLoadId: 'LEI', caseLoadIds: ['LEI'], roles: ['ROLE_GLOBAL_SEARCH'] })
```
nathanrussell-moj-digital, line 75: In my last PR that added the `checkPrisonerInCaseloadMiddleware` stuff, where I'd copied it from DPS, I was using the wrong property name from the user in session re: their roles.
In the DPS codebase they use the property name `userRoles`, but our `User` uses the property name `roles` (`UserDetails` in `userService.ts`)

server/routes/index.ts
```diff
@@ -18,17 +18,18 @@ export default function routes(services: Services): Router {
   prisonerSummarySetup(router, services)
 
   // Application routes
-  prisonerList(router, services)
+  inPrisonCoursesAndQualifications(router, services)
```
nathanrussell-moj-digital, line 21: I had to re-order the route definition because express has some rules about the sequencing/ordering of how it resolves them and their middlewares etc.

I've not explained that very well, but previously we had `prisonerList` first because it made sense for it to be defined first as it is the first page the user sees and is the root route. The problem though is that it applies the `checkUserHasViewAuthority` middleware to the root:
```
  const prisonerListController = new PrisonerListController(services.prisonerListService)

  router.use('/', [checkUserHasViewAuthority()])
  router.get('/', [asyncMiddleware(prisonerListController.getPrisonerListView)])
}
```
This means that the user has to be a PLP user (either with our editor or viewer role) in order to see the root route .... or any route below it. That was a problem ...

The `inPrisonCoursesAndQualifications` routes adds a new route `/prisoner/:prisonNumber/work-and-skills/in-prison-courses-and-qualifications` which has different security requirements, specifically for those users who are not PLP users.

By putting the `inPrisonCoursesAndQualifications` first and (more importantly) the `prisonerList` (root) route last we fix this problem 👍

leejacobson-moj (approved):
Looks good. Approved 👍

PR #614, opened 2024-05-14
RR-777 - Updated text & colour on Overview screen

nathanrussell-moj-digital:
Small PR to update the colour (from green to standard text colour) and some text, on the Overview screen:

Screenshot 2024-05-14 at 08 45 12

### DIFF
--- a/assets/scss/components/_summary-card.scss
+++ b/assets/scss/components/_summary-card.scss
@@ -5,14 +5,6 @@
   padding: 16px 0 0 16px;
 }
 
-.app-summary-card__title-wrapper--green {
-  border-top: 4px solid govuk-colour("green");
-
-  .govuk-summary-card__title {
-    color: govuk-colour("green");
-  }
-}
-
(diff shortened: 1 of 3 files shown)


assets/scss/components/_summary-card.scss
```diff
@@ -5,14 +5,6 @@
   padding: 16px 0 0 16px;
 }
 
-.app-summary-card__title-wrapper--green {
```
nathanrussell-moj-digital, line 8: From day 1 we (devs) always questioned why this section was a different colour, and why that colour was green.
Our concern was that is was not accessible, in that if using a different colour was meant to convey some kind of meaning or significance, then using colour alone was not enough. We also questioned why green, as green might suggest "good" or "complete" or "better than something that is not green" etc.

Screenshot 2024-05-14 at 08 48 39

I'm glad this has been changed 👍

leejacobson-moj (approved):
Looks good. Approved 👍

PR #615, opened 2024-05-14
RR-790: Adds new audit events to service

leejacobson-moj:
**Overview**

This PR adds pageview audit events for the new pages in the create induction journey.

It also add a new `checkPageViewAuditted` middleware to check that the route has an associated audit event before continuing to load the route. If it's not the route will be skipped resulting in a 404 error.

Ideally we would show a error in these events, but doing this would mean all unknown routes would result in a error as they would not match any auditable routes. Express has no way to know whether a route will later match another route handler so our only option to catch unaudited routes without transforming all 404s into 500 errors is to treat them as unconfigured routes and display a 404. However, an error is logged in the console noting that the router will be skipped.


server/app.ts
```diff
@@ -58,7 +58,7 @@ export default function createApp(services: Services): express.Application {
   app.use(routes(services))
 
   app.use((req, res, next) => next(createError(404, 'Not found')))
-  app.use(errorHandler(process.env.NODE_ENV === 'production'))
+  app.use(errorHandler(services, config.production))
```
leejacobson-moj, line 61: In addition to passing services into the error handler middleware I've updated to use the `config.production` value here instead of reevaluating.

server/middleware/auditMiddleware.ts
```diff
@@ -41,14 +80,19 @@ const pageViewEventMap: Record<string, Page> = {
   '/prisoners/:prisonNumber/induction/qualification-level': Page.INDUCTION_UPDATE_QUALIFICATION_LEVEL,
   '/prisoners/:prisonNumber/induction/qualification-details': Page.INDUCTION_UPDATE_QUALIFICATION_DETAILS,
   '/prisoners/:prisonNumber/induction/additional-training': Page.INDUCTION_UPDATE_ADDITIONAL_TRAINING,
+  '/prisoners/:prisonNumber/induction/check-your-answers': Page.INDUCTION_UPDATE_CHECK_YOUR_ANSWERS,
 
-  '/prisoner/:prisonNumber/work-and-skills/in-prison-courses-and-qualifications':
-    Page.IN_PRISON_COURSES_AND_QUALIFICATIONS,
+  // Non audit routes. These routes do not raise an audit event
+  '/plan/:prisonNumber/induction-created': null,
```
leejacobson-moj, line 86: This route is just used in a redirect, so no need to raise an audit event for it.

server/middleware/auditMiddleware.ts
```diff
@@ -80,3 +124,12 @@ export default function auditMiddleware({ auditService }: Services): Router {
...
 }
+
+// Checks page view has been auditted, if no audit event has been raised router will be skipped
+export function checkPageViewAuditted(router: Router) {
+  router.get('*', (req: Request, res: Response, next: NextFunction) => {
+    if (res.locals.auditPageViewEvent || res.locals.auditPageViewEvent === null) return next()
+    logger.error(`No audit event found for route, "${req.path}". Skipping router.`)
+    return next('router')
+  })
+}
```
leejacobson-moj, line 135: This isn't an ideal solution, but because express has no way to know whether a route will later match another route our only option to catch unaudited routes without transforming all 404s into 500 errors is to treat them as unconfigured routes and display a 404. This will allow us to catch pages which are unaudited and ensure every page is audited.

server/routes/induction/create/index.ts
```diff
@@ -172,10 +172,6 @@ export default (router: Router, services: Services) => {
       asyncMiddleware(affectAbilityToWorkCreateController.submitAffectAbilityToWorkForm),
     ])
 
-    router.get('/prisoners/:prisonNumber/create-induction/skills', [
-      asyncMiddleware(skillsCreateController.getSkillsView),
-    ])
-
```
leejacobson-moj, line 178: This is a duplicate

nathanrussell-moj-digital (approved):
LGTM 👍

PR #616, opened 2024-05-14
RR-789 - Refactored CreateGoalsForm to use enum for goal target completion date

nathanrussell-moj-digital:
This PR implements an approach in [this POC](https://github.com/ministryofjustice/hmpps-education-and-work-plan-ui/pull/599) re: using an enum for the goal target completion date, rather than having the controller work out the dates to pass to the view.

The basic idea is that we have an enum that describes the target date completion options (3 months, 6 months, 12 months and another/manually entered date). The enum value set is passed to the view for use with the radio buttons, and the view is responsible for working out the dates to use as labels for the radios (because that is very much a presentation concern 👍 )
On form submission the controller uses a mapper to map for the submitted form into a DTO, and it is the job of that mapper to work out what 3/6/12 months from today is 👍 

There's actually slightly more change necessary in this PR than in the POC in this respect. This is because the POC was mainly about replacing the Create Goals validation. 
This PR is not concerned with replacing the validation approach (that will be the subject of a later story and PR), so we keep the existing validator. However, the date related aspects of the Create Goals Validator (that need refactoring due to the use of an enum) are also used by the old Create Goal (singular) validator and the Update Goal validator.

### DIFF
--- a/server/routes/createGoal/createGoalFormValidator.test.ts
+++ b/server/routes/createGoal/createGoalFormValidator.test.ts
@@ -1,27 +1,17 @@
 import type { CreateGoalForm } from 'forms'
-import validateGoalTitle from '../../validators/goalTitleValidator'
-import goalTargetCompletionDateValidator from '../../validators/goalTargetCompletionDateValidator'
 import validateCreateGoalForm from './createGoalFormValidator'
 
-jest.mock('../../validators/goalTitleValidator')
-jest.mock('../../validators/goalTargetCompletionDateValidator')
-
 describe('createGoalFormValidator', () => {
-  const mockedValidateGoalTitle = validateGoalTitle as jest.MockedFunction<typeof validateGoalTitle>
-  const mockedValidateTargetCompletionDate = goalTargetCompletionDateValidator as jest.MockedFunction<
-    typeof goalTargetCompletionDateValidator
-  >
-
   it('should validate given validators return no errors', () => {
     // Given
-    const form = {
+    const form: CreateGoalForm = {
       prisonNumber: 'A1234BC',
       title: 'Learn Spanish',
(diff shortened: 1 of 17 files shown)


server/routes/createGoal/createGoalFormValidator.test.ts
```diff
@@ -1,27 +1,17 @@
 import type { CreateGoalForm } from 'forms'
-import validateGoalTitle from '../../validators/goalTitleValidator'
-import goalTargetCompletionDateValidator from '../../validators/goalTargetCompletionDateValidator'
 import validateCreateGoalForm from './createGoalFormValidator'
 
-jest.mock('../../validators/goalTitleValidator')
-jest.mock('../../validators/goalTargetCompletionDateValidator')
-
 describe('createGoalFormValidator', () => {
-  const mockedValidateGoalTitle = validateGoalTitle as jest.MockedFunction<typeof validateGoalTitle>
```
nathanrussell-moj-digital, line 10: These validators call out to other validators. I've got rid of all the mocks and instead let the validator use the real implementations of the "child" validators.

When it comes to the assertion though, the important thing for _this_ unit test of _this_ validator is that the errors array is correctly built with the right field name. The text is largely irrelevant (because the text comes from the "child" validators, and they have their own unit tests that assert the text). This is why we used mocks before, so as to abstract this test from the implementation of the "child" validators (whose text might change over time)
Because I still dont want to couple _these_ tests with the child validator implementations, I've changed the assertion to assert any string:
EG: I've changed this:
```
expect(errors).toEqual([{ href: '#title', text: 'some-title-error' }])
```
to this:
```
expect(errors).toEqual([{ href: '#title', text: expect.any(String) }])
```

PR #618, opened 2024-05-14
RR-787: Success middleware

leejacobson-moj:
## Overview

This adds a new success middleware which configures a new response helper method, `redirectWithSuccess()`. This method will redirect to a page and show a success banner with the desired success message.

### DIFF
--- a/server/@types/express/index.d.ts
+++ b/server/@types/express/index.d.ts
@@ -30,6 +30,12 @@ import type { UserDetails } from '../../services/userService'
 
 export default {}
 
+declare module 'express-serve-static-core' {
+  interface Response {
+    redirectWithSuccess?(path: string, message: string): void
+  }
+}
+
 declare module 'express-session' {
   // Declare that the session will potentially contain these additional fields
   interface SessionData {
--- a/server/@types/express/index.d.ts
+++ b/server/@types/express/index.d.ts
@@ -30,12 +30,6 @@ import type { UserDetails } from '../../services/userService'
 
 export default {}
 
-declare module 'express-serve-static-core' {
-  interface Response {
-    redirectWithSuccess?(path: string, message: string): void
-  }
-}
-
 declare module 'express-session' {
   // Declare that the session will potentially contain these additional fields
   interface SessionData {
@@ -79,6 +73,10 @@ export declare global {
       username: string
       token: string
       authSource: string
+    }
+
+    interface Response {
+      redirectWithSuccess?(path: string, message: string): void
     }
 
     interface Request {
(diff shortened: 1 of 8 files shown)


server/middleware/setUpAuthentication.ts
```diff
@@ -12,7 +11,6 @@ export default function setUpAuth(): Router {
 
   router.use(passport.initialize())
   router.use(passport.session())
-  router.use(flash())
```
leejacobson-moj, line 15: I was running into an error because before this change `req.flash` didn't exist if called before this `setUpAuthentication` middleware.

 This feels like the wrong place to initialise flash since it's not coupled with authentication? I've moved into `server/app.ts` instead.

server/@types/express/index.d.ts
```diff
@@ -30,6 +30,12 @@ import type { UserDetails } from '../../services/userService'
 
 export default {}
 
+declare module 'express-serve-static-core' {
```
nathanrussell-moj-digital, line 33: I have no strong feelings over this, and I am certainly no typescript or express expert! 🤣 , but is there a reason why we create a new module here with the modified Response interface in it, rather than adding it to the global `Express` namespace towards the bottom of this file?

I'm happy if this is the right way to do it; it's more a question so that I might learn something about typescript/express 😁 👍

server/middleware/setUpAuthentication.ts
nathanrussell-moj-digital: Yeah, agreed 👍 
I don't know why it was here in `setUpAuthentication` - I guess we got it from the template? But agreed it feels misplaced here and where you've put it in `app.ts` is better 👍

nathanrussell-moj-digital: Having said that, in your POC you didnt need to make this change because you setup the success  middleware further down the file in `app.ts` than you are doing here. I think in the POC it was around line 61 after the `auditMiddleware`, but crucially it was after this authentication middleware.

Whilst I 100% agree that setting up flash in the authentication middleware seems weird and misplaced, I guess it depends on how far we want to deviate from the template? 🤷‍♂️ Might this cause us problems (or at least reason to pause and think) if we take an update to the template that updates this authentication middleware?

nathanrussell-moj-digital (dismissed):
LGTM 👍

server/middleware/setUpAuthentication.ts
leejacobson-moj: I decided to add the middleware higher because it felt like this was a core piece of functionality which other middlewares might want to call. Realistically though we're probably only ever going to use it in our application route handlers.

I think you're probably right... I don't like how it is, but this probably the wrong to make the change. It would be better done as a template change otherwise we're going to cause ourselves more headaches down the line...

I'll update.

nathanrussell-moj-digital (approved):
LGTM 👍

PR #621, opened 2024-05-15
RR-491 - Replace all uses of moment.js with date-fns in cypress tests

nathanrussell-moj-digital:
This PR replaces all uses of moment.js with corresponding date-fns function calls in our cypress integration tests.

There is still work to do re: replacing moment with date-fns; but its all in the main code now. All date related stuff in the integration tests is now using date-fns 👍


leejacobson-moj (approved):
Looks good. Approved 👍

PR #622, opened 2024-05-15
RR-788: Removes form initalisation from controller

leejacobson-moj:
## Overview

This change removes the form initalisation code from the controller and updates the view so that if the form is not initialised it will loop once to add the initial goal and step inputs.


nathanrussell-moj-digital (approved):
LGTM 👍

PR #623, opened 2024-05-15
RR-789 - Fix date formatting for BST dates

nathanrussell-moj-digital:
This PR fixes a date formatting issue that I inadvertently introduced in my last PR re: changing the `targetCompletionDate` field to use an enum

Basically I introduced a problem with any target completion date that would be in BST (and the current date is not in BST) - in this scenario the target completion date ended up being 1 day less (I hate dates and timezones! 🤣 )

Anyway, the problem was in the mapper that maps the CreateGoals form into the DTO that is sent to the API, in that it takes the targetCompletionDate (which is a `Date` instance) and formats it into a string.
The fix is to use `date-fns` to format the date; and the new unit test is there to prove 👍

### DIFF
--- a/server/data/mappers/createGoalMapper.test.ts
+++ b/server/data/mappers/createGoalMapper.test.ts
@@ -1,12 +1,17 @@
 import type { CreateGoalDto } from 'dto'
 import type { CreateGoalRequest, CreateStepRequest } from 'educationAndWorkPlanApiClient'
+import { parse, startOfToday } from 'date-fns'
 import { toCreateGoalRequest } from './createGoalMapper'
 import { aValidCreateGoalDtoWithMultipleSteps } from '../../testsupport/createGoalDtoTestDataBuilder'
 
 describe('createGoalMapper', () => {
-  it('should map to CreateGoalRequest given a valid DTO', () => {
+  const today = startOfToday()
+
+  it('should map to CreateGoalRequest given a valid DTO whose targetCompletionDate is not in BST', () => {
     // Given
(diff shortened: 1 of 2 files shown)


server/data/mappers/createGoalMapper.test.ts
```diff
@@ -31,4 +36,34 @@ describe('createGoalMapper', () => {
...
+    const expectedAddStepRequest2: CreateStepRequest = {
+      title: createGoalDto.steps[1].title,
+      sequenceNumber: 2,
+    }
+    const expectedCreateGoalRequest: CreateGoalRequest = {
+      prisonNumber: createGoalDto.prisonNumber,
+      title: createGoalDto.title,
+      category: 'WORK',
+      steps: [expectedAddStepRequest1, expectedAddStepRequest2],
+      targetCompletionDate: '2024-08-10',
```
nathanrussell-moj-digital, line 58: Without the change in the implementation (file below), this test fails because the date is formatted as `2024-08-09`

server/data/mappers/createGoalMapper.ts
```diff
@@ -24,8 +25,4 @@ const toAddStepRequest = (addStepDto: AddStepDto): CreateStepRequest => {
   }
 }
 
-const toDateString = (date: Date): string => {
-  return date?.toISOString().split('T')[0]
-}
```
nathanrussell-moj-digital, line 29: TBH, this looks like it was a fairly naive way of formatting dates 🤷‍♂️

leejacobson-moj (approved):
Looks good. Approved 👍

PR #625, opened 2024-05-16
RR-793: Error message middleware

leejacobson-moj:
## Overview

Adds error message middleware including new, `redirectWithErrors()` response method.

Changes in this PR include:
- Error message middleware
- Removal of error banner from page templates and including in layout template
- Updated tests to remove `req.flash()` assertions and replaced with assertion on `redirectWithErrors()`
- Removal of `errors` field from view models.


server/middleware/successMessageMiddleware.ts
leejacobson-moj, line 1: Renamed this to align with `errorMessageMiddleware`.

server/views/partials/layout.njk
```diff
@@ -63,19 +63,31 @@
       }
     }) }}
 
-    {% if successMessage %}
```
leejacobson-moj, line 66: While testing I noticed I put this in the wrong place in the other PR so have moved it as well as adding the error message banner.

server/@types/express/index.d.ts
```diff
@@ -77,14 +77,14 @@ export declare global {
...
       redirectWithSuccess?(path: string, message: string): void
+      redirectWithErrors?(path: string, message: Record<string, string>[]): void
     }
 
     interface Request {
       verified?: boolean
       id: string
       logout(done: (err: unknown) => void): void
       flash(type: string, message: Array<Record<string, string>>): number
-      flash(message: 'errors'): Array<Record<string, string>>
```
nathanrussell-moj-digital, line 87: I'm pretty sure that this line came from the original template, so I'm wondering whether it would be as well to keep it, just in case there are future updates from the template in this area?

server/middleware/errorMessageMiddleware.ts
```diff
@@ -0,0 +1,18 @@
...
+  if (req.method === 'GET') {
+    const errorMessages = req.flash('errors')[0]
+
+    if (errorMessages) {
+      res.locals.errors = JSON.parse(errorMessages)
+    }
+  }
+
+  next()
+}
```
nathanrussell-moj-digital, line 18: Assuming we kept the original flash errors declaration in `express/index.d.ts` above, I wonder if we could simplify this by not going via json representations of the array? Something like:
```
export default function errorMessageMiddleware(req: Request, res: Response, next: NextFunction) {
  res.redirectWithErrors = (path: string, errors: Record<string, string>[]): void => {
    req.flash('errors', errors)
    res.redirect(path)
  }

  if (req.method === 'GET') {
    res.locals.errors = req.flash('errors')
  }

  next()
}
```

Would that work? 🤷‍♂️

nathanrussell-moj-digital (approved):
Wow! This is a much bigger PR that I thought it would be, but I guess it has to touch every form based controller, every view class for form pages, and every nunjucks template for forms! Lots of files 😁 

Nice work!
Only one comment/suggestion, and that is about keeping and using the original flash messages method signature, but happy either way 👍 
Looks ace 👍 😁

PR #630, opened 2024-05-23
RR-801 - Fix routing when changing "Do they want to record qualifications" from Check Your Answers

nathanrussell-moj-digital:
This PR fixes a bug that Karen found with the Induction journey.
This PR fixes the bug and implements the Expected Behaviour

From the Jira ticket:

### Steps to reproduce
* Create a new Induction
* On the first question (do they want to work on release) choose No or Not Sure to make it a Short question set Induction
* When asked whether they want to add any qualifications choose Yes
* Enter at least 1 qualification
* Complete the rest of the Induction, no special or significant answers need to be provided. Answer as you feel fit
* Arrive on Check Your Answers
* Observe that the answer to “do they want to record any qualifications” is Yes and that the qualifications previously entered are correctly displayed
* Click the change link for “do they want to record any qualifications”
* On the “do they want to record any qualifications” change the previous answer of Yes to No and submit the page

### Expected behaviour
The user should arrive back on Check Your Answers where the following should be observed:
* The answer to “do they want to record any qualifications” is now No 
* The qualifications previously entered have been removed and are no longer displayed

### Actual behaviour
User is incorrectly taken through the rest of the Induction journey rather that showing Check Your Answers
If the user completes the rest of the Induction journey by submitting each page until they arrive at Check Your Answers the following is observed:
* The answer to “do they want to record any qualifications” is still Yes, even though it was apparently changed to No 
* The qualifications previously entered are still displayed

### DIFF
--- a/server/routes/induction/common/wantToAddQualificationsController.ts
+++ b/server/routes/induction/common/wantToAddQualificationsController.ts
@@ -6,6 +6,7 @@ import InductionController from './inductionController'
 import WantToAddQualificationsView from './wantToAddQualificationsView'
 import dateComparator from '../../dateComparator'
 import YesNoValue from '../../../enums/yesNoValue'
+import EducationLevelValue from '../../../enums/educationLevelValue'
 
 /**
  * Abstract controller class defining functionality common to both the Create and Update Induction journeys.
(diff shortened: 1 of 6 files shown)


server/routes/induction/common/wantToAddQualificationsController.ts
```diff
@@ -43,6 +44,34 @@ export default abstract class WantToAddQualificationsController extends Inductio
     )
     return res.render('pages/induction/prePrisonEducation/wantToAddQualifications', { ...view.renderArgs })
   }
+
+  protected formSubmittedFromCheckYourAnswersWithNoChangeMade = (
```
nathanrussell-moj-digital, line 48: These methods have been moved from the `WantToAddQualificationsUpdateController` subclass up into the superclass here, so that they can be used by the `WantToAddQualificationsCreateController` subclass

(basically the desired functional existed in the Update controller, but not the Create controller; so moving it here allows both to use it)

leejacobson-moj (approved):
Looks good. Approved 👍

PR #631, opened 2024-05-23
RR-729 - Basic route, controller and view definitions for prisoner notes list screen

nathanrussell-moj-digital:
This PR adds the basic initial route, controller and nunjucks template for the prisoner notes list screen (where it will list all of the notes for a prisoner)

At the moment it doesn't list anything as we dont have the API in place yet etc, but its a starter for 10 in respect of defining the route, controller and nunjucks

Screenshot 2024-05-23 at 14 42 50


server/routes/index.ts
```diff
@@ -33,6 +34,8 @@ export default function routes(services: Services): Router {
   updateInduction(router, services)
   postInductionCreation(router, services)
 
+  router.use('/plan/:prisonNumber/notes', notesRoutes())
```
nathanrussell-moj-digital, line 37: Re: this line, and the next file .... is this the preferred way of setting up routes in express?
Assuming this is the preferred way we can start to define routes using this pattern and I'll create a tech-debt ticket to refactor the other routes to use this pattern as well

leejacobson-moj: I guess "preferred" might depend on who you ask, but I think generally, yes, because it's more a flexible way to assign routes.

Doing it this way would allow us to define the notes routes independently and relative to whatever route the notes router has been assigned to which would make them easier to manage and move as a group. We could even assign the notes router to multiple parent routes if we wanted.

PR #632, opened 2024-05-23
RR-805 - Fix routing when clicking back from previous work experience detail page

nathanrussell-moj-digital:
This PR fixes a bug when clicking Back from a "previous work experience detail" page in the Create induction journey

From the jira ticket:

### Steps to reproduce
* Create a new Induction
* Answer “do they want to work on release” as Yes to make it a Long question set Induction
* When asked if the prisoner has worked before, answer Yes
* On the “previous work experience types” screen select one or more work types. Submit the page
* One the next screen (the work detail page for a work type) click the application Back button
* Observe that the “previous work experience types” screen is redisplayed with the previously made selections retained
* Do not change the work experience types, and simply click “Continue”

### Expected behaviour
* The work detail page should be displayed

### Actual behaviour
* A “page not found” error page is displayed

### DIFF
--- a/server/routes/induction/create/previousWorkExperienceTypesCreateController.ts
+++ b/server/routes/induction/create/previousWorkExperienceTypesCreateController.ts
@@ -52,7 +52,7 @@ export default class PreviousWorkExperienceTypesCreateController extends Previou
     req.session.inductionDto = updatedInduction
 
     // We need to show the Details page for each work experience type.
-    const pageFlowQueue = buildPageFlowQueue(inductionDto, updatedInduction, prisonNumber)
+    const pageFlowQueue = buildPageFlowQueue(updatedInduction, prisonNumber)
     req.session.pageFlowQueue = pageFlowQueue
 
     const userHasComeFromCheckYourAnswers = this.checkYourAnswersIsTheFirstPageInThePageHistory(req)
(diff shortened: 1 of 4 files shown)


server/routes/induction/create/previousWorkExperienceTypesCreateController.ts
```diff
@@ -69,29 +69,14 @@ export default class PreviousWorkExperienceTypesCreateController extends Previou
 
 /**
  * Builds and returns a Page Flow Queue to show the Details page for each work experience type. The list of pages to be
- * added to the queue is derived by the difference between the work types on the updated induction (ie. after the
```
nathanrussell-moj-digital, line 72: This was the problem area, in that it was trying to be too smart for it's own good!
What it tried to do was work out the diff between the previous work experience types on the induction between when the page was first loaded vs when it was submitted.

This worked great in a linear create journey, where you start off with no work experiences ticked, you set some, and submit. Therefore all the ones you ticked are the resultant diff

And it worked great when clicking Change from Check Your Answers, in that it would work out the new ones you'd selected and only show the detail page for those newly added work experience types.

Where it doesn't work, and the cause of this bug, is when there are no differences between when the page was loaded and the page being submitted. 
EG. if you click Change from Check Your Answers, but dont actually change anything
or (in this case), you click Back from the work experience detail page (ie. the next page after this one), then click Continue to move forward again without actually making any changes.

The simplest option here was to remove the complexity of it trying to be smart. We no longer try to work out the diff, and simply show the work experience detail page for each type submitted on the form.

leejacobson-moj (approved):
Looks good. Approved 👍

PR #643, opened 2024-05-28
RR-831 - Refactor Timeline Event contextualInfo; now returned from the API as a map

nathanrussell-moj-digital:
This PR is the UI counterpart to this [API PR](https://github.com/ministryofjustice/hmpps-education-and-work-plan-api/pull/287)

It re-imports the swagger spec to get the latest spec/types; then refactors all the uses of `contextualInfo` as it is now a map rather than a simple string

### DIFF
--- a/server/routes/timelineResolver.ts
+++ b/server/routes/timelineResolver.ts
@@ -47,6 +47,14 @@ const eventsWithMergedCreateGoalEvents = (
         goalCreatedEvents.length > 1
           ? ({
               ...goalCreatedEvents[goalCreatedEvents.length - 1],
+              contextualInfo: goalCreatedEvents // Map all the goal created events contextualInfo properties into an object containing them all with an incremental number
+                .map((event, idx) =>
+                  idx > 0 ? event.contextualInfo : { GOAL_TITLE_0: event.contextualInfo.GOAL_TITLE },
+                )
+                .reduce((previous, current, idx) => ({
+                  ...previous,
+                  [`GOAL_TITLE_${idx}`]: current.GOAL_TITLE,
+                })),
               eventType: 'MULTIPLE_GOALS_CREATED',
             } as TimelineEvent)
           : goalCreatedEvents[0],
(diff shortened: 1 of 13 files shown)


server/routes/timelineResolver.ts
```diff
@@ -47,6 +47,14 @@ const eventsWithMergedCreateGoalEvents = (
...
           ? ({
               ...goalCreatedEvents[goalCreatedEvents.length - 1],
+              contextualInfo: goalCreatedEvents // Map all the goal created events contextualInfo properties into an object containing them all with an incremental number
+                .map((event, idx) =>
+                  idx > 0 ? event.contextualInfo : { GOAL_TITLE_0: event.contextualInfo.GOAL_TITLE },
+                )
+                .reduce((previous, current, idx) => ({
+                  ...previous,
+                  [`GOAL_TITLE_${idx}`]: current.GOAL_TITLE,
+                })),
```
leejacobson-moj, line 57: I may be misunderstanding this change, but I don't understand why the .map() is needed here – I think you might be able to just use a reduce?

```suggestion
              contextualInfo: goalCreatedEvents // Map all the goal created events contextualInfo properties into an object containing them all with an incremental number
                .reduce((previous, current, idx) => ({
                  ...previous,
                  [`GOAL_TITLE_${idx}`]: current.contextualInfo.GOAL_TITLE,
                }), {}),
```

leejacobson-moj (approved):
Looks good. Approved 👍

server/routes/timelineResolver.ts
nathanrussell-moj-digital: There are a couple of reasons;
Firstly, the reduce on its own returns an instance of the `previous` and `current` types - in this case it would have been a `TimelineEvent` (because we are iterating through an array of `goalCreatedEvents`). Actually what we need here is a new `contextualInfo` object, which is a simple key value pairs (`contextualInfo: { [key: string]: string }`)

Secondly, reduce is a funny old beast 🤣 in that the first iteration is actually for the 2nd element in the array, and the `previous` arg is the first element (ie. if there are 2 elements in the array, the `reduce` lambda will only be called once)
Because `previous` is the first element, and our objective is to create a object such as:
```
{
    GOAL_TITLE_0: '1st goal title`,
    GOAL_TITLE_1: '2nd goal title`,
    GOAL_TITLE_2: '3rd goal title`,
}
```
we need to "fudge" the first element to have a property name of `GOAL_TITLE_0`; otherwise the result is:
```
{
    GOAL_TITLE: '1st goal title`,
    GOAL_TITLE_1: '2nd goal title`,
    GOAL_TITLE_2: '3rd goal title`,
}
```

PR #644, opened 2024-05-28
RR-832 - Fix qualification level options

nathanrussell-moj-digital:
This PR fixes a bug with the qualification options on the Qualification Level screen.

The problem was that the nunjucks view was copied from the CIAG UI, and the CIAG UI and API's approach to qualifications is that if you were not asked Highest Level of Education then the view property `educationLevel` was an empty string.

When we migrated this to our API and UI we decided that it would be better for this view property (and the value sent to the API) to be `NOT_SURE` if you were not asked the question.


leejacobson-moj (approved):
Looks good. Approved 👍

PR #645, opened 2024-05-29
RR-833: Added null check to user object

leejacobson-moj:
## Overview

I introduced null error bug when adding audits event for the error page since this page can be reached when a user is not logged in.

This change is to add null checks to the user objects.


nathanrussell-moj-digital (approved):
LGTM 👍

PR #650, opened 2024-05-30
RR-812 - Move the Type Of Work question screens in the Induction flow

nathanrussell-moj-digital:
This PR is the first of the Question Set changes. The target branch for this PR is our integration branch `feature/RR-810-question-set-epic`

---

This PR moves the "Type of work" screens (Work Interest Types and Work Interest Roles) within the Long question set to immediately after "Do they want to work on release" and before "Qualifications".
The new screen flow can be seen [in this miro frame](https://miro.com/app/board/uXjVKe-qkvg=/?moveToWidget=3458764590620132715&cot=14).

Whilst we are only moving 2 screens, this change impacts many controllers in the create and induction journeys as the controllers for the screens before and after need to change (to change the "next" page on the screen before, and change the back link on the screen after); plus other places in the journey where you can ultimately end up on these screens. 


leejacobson-moj (approved):
Looks good. Approved 👍

PR #655, opened 2024-06-05
RR-636 - Enabled feature toggles for the migrated Induction screens

nathanrussell-moj-digital:
This PR enables the feature toggles (2 toggles) to enable the migrated Induction screens in PLP (rather than redirecting the user to the CIAG UI screens)


leejacobson-moj (approved):
Looks good. Approved 👍

PR #656, opened 2024-06-05
RR-816 - Add Personal Skills and Interests to the short question set Induction flow

nathanrussell-moj-digital:
This PR is the 3rd of the Question Set changes. The target branch for this PR is our integration branch feature/RR-810-question-set-epic

---

This PR adds the "Personal Skills" and "Personal Interests" screens to the Short question set immediately after "Additional Training" and before "In Prison Work".
The new screen flow can be seen [in this miro frame](https://miro.com/app/board/uXjVKe-qkvg=/?moveToWidget=3458764591206842029&cot=14).

Whilst we are only adding 2 screens, this change impacts many controllers in the create and induction journeys as the controllers for the screens before and after need to change (to change the "next" page on the screen before, and change the back link on the screen after); plus other places in the journey where you can ultimately end up on these screens.


leejacobson-moj (approved):
Looks good. Approved 👍

PR #659, opened 2024-06-07
RR-820/RR-836 - Move 'Highest Level of Eduction' and 'Do You Want To Add Qualifications?'

nathanrussell-moj-digital:
This PR is the 6th of the Question Set changes. The target branch for this PR is our integration branch feature/RR-810-question-set-epic

---

This PR moves the 'Highest Level of Education' (RR-836) and 'Do you want to add qualifications?' (RR-820) screens.

This ticket implements 2 stories (RR-820 and RR-836) because they are so closely related it didnt make sense to do them separately.
The new screen flow can be seen [in this miro frame](https://miro.com/app/board/uXjVKe-qkvg=/?moveToWidget=3458764591343999122&cot=14).

Whilst we are only moving 2 screens, this change impacts many controllers in the create and induction journeys as the controllers for the screens before and after need to change (to change the "next" page on the screen before, and change the back link on the screen after); plus other places in the journey where you can ultimately end up on these screens.


leejacobson-moj (approved):
Looks good. Approved 👍

PR #666, opened 2024-06-13
RR-846: Add "not relevant" option to has worked before question

james-baker-digital-justice:
This is the same as No in terms of follow up questions asked but allows the work experience to be skipped in cases where the person might actually have work experience but the user decides not to enter it.

### DIFF
--- a/integration_tests/support/commands.ts
+++ b/integration_tests/support/commands.ts
@@ -34,6 +34,7 @@ import QualificationLevelValue from '../../server/enums/qualificationLevelValue'
 import QualificationDetailsPage from '../pages/induction/QualificationDetailsPage'
 import AdditionalTrainingValue from '../../server/enums/additionalTrainingValue'
 import WantToAddQualificationsPage from '../pages/induction/WantToAddQualificationsPage'
+import HasWorkedBeforeValue from '../../server/enums/hasWorkedBeforeValue'
 
 Cypress.Commands.add('signIn', (options = { failOnStatusCode: false }) => {
   cy.request('/')
@@ -243,7 +244,7 @@ Cypress.Commands.add(
       .submitPage()
     // Have You Worked Before page is next
     Page.verifyOnPage(WorkedBeforePage) //
-      .selectWorkedBefore(YesNoValue.YES)
+      .selectWorkedBefore(HasWorkedBeforeValue.YES)
       .submitPage()
     // Previous Work Experience Types is the next page
     Page.verifyOnPage(PreviousWorkExperienceTypesPage) //
@@ -323,7 +324,7 @@ Cypress.Commands.add(
       .submitPage()
     // Have You Worked Before page is next
     Page.verifyOnPage(WorkedBeforePage) //
-      .selectWorkedBefore(YesNoValue.YES)
+      .selectWorkedBefore(HasWorkedBeforeValue.YES)
       .submitPage()
     // Previous Work Experience Types is the next page
     Page.verifyOnPage(PreviousWorkExperienceTypesPage) //
--- a/integration_tests/support/commands.ts
+++ b/integration_tests/support/commands.ts
@@ -165,6 +165,7 @@ Cypress.Commands.add('updateLongQuestionSetInductionToArriveOnCheckYourAnswers',
   // 'Has the prisoner worked before' is the next page. This is asked on the short question set so will have answers but
   // we will make a change to exercise the screen flow
   Page.verifyOnPage(WorkedBeforePage) //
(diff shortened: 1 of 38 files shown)


server/routes/induction/create/workedBeforeCreateController.test.ts
```diff
@@ -200,68 +200,74 @@ describe('workedBeforeCreateController', () => {
...
-      )
-
-      // Then
-      const updatedInduction = req.session.inductionDto
-      expect(updatedInduction.previousWorkExperiences.hasWorkedBefore).toEqual(false)
-      expect(updatedInduction.previousWorkExperiences.experiences).toEqual([])
-      expect(res.redirect).toHaveBeenCalledWith('/prisoners/A1234BC/create-induction/check-your-answers')
-      expect(req.session.workedBeforeForm).toBeUndefined()
-    })
+    it.each([HasWorkedBeforeValue.NO, HasWorkedBeforeValue.NOT_RELEVANT])(
```
nathanrussell-moj-digital, line 206: Much love for `it.each` 😁 👍 
(In some early test code you'll find we create and loop through an array - we didn't know about `it.each` when we wrote those older tests; this is a much neater way 👍 )

server/views/pages/induction/workedBefore/index.njk
```diff
@@ -10,7 +10,7 @@
     * backLinkUrl - url of the back link
     * backLinkAriaText - the aria label for the back link
     * form - form object containing the following fields:
-      * hasWorkedBefore - value for the "has worked before" question
+    * hasWorkedBefore - value for the "has worked before" question
```
nathanrussell-moj-digital, line 13: Very (very) minor, but the previous indentation here was meant to show that `hasWorkedBefore` is a property on the `form` object.
With the change in indentation here it could now be read that the `form` object contains both a `hasWorkedBefore` property AND an `errors` property

server/testsupport/createInductionDtoTestDataBuilder.ts
```diff
@@ -35,13 +36,13 @@ const aLongQuestionSetCreateInductionDto = (
     previousWorkExperiences: {
       hasWorkedBefore:
         !options || options.hasWorkedBefore === null || options.hasWorkedBefore === undefined
-          ? true
+          ? HasWorkedBeforeValue.YES
```
nathanrussell-moj-digital, line 39: All of these changes to the test data builders (6 files in total I think) can be simplified if you want.

What you've done works, but was originally written in this fairly long hand way because the previous field was a `boolean` and because of javascript's horrible falsey-ness! 😬 

The idea with these test data builder functions is that they take an optional `options` object, which in turn contains some optional properties. If the `options` object, or a given property is not set, then we use a sensible default.

Now that your property is no longer a boolean, it means you can do this:
```
    hasWorkedBefore: options?.hasWorkedBefore || HasWorkedBeforeValue.YES
    ....
    experiences: options?.hasWorkedBefore === HasWorkedBefore.Yes
        ? .....
        : ....
```

Much simpler 😁 

I think you can make this same change to all 6 test data builder files you've changed here - entirely up to you if you want to do it (because what you have does still work 👍 )

nathanrussell-moj-digital (approved):
Fantastic! LGTM 👍

integration_tests/support/commands.ts
```diff
@@ -164,6 +165,7 @@ Cypress.Commands.add('updateLongQuestionSetInductionToArriveOnCheckYourAnswers',
   // 'Has the prisoner worked before' is the next page. This is asked on the short question set so will have answers but
   // we will make a change to exercise the screen flow
   Page.verifyOnPage(WorkedBeforePage) //
+    .selectWorkedBefore(HasWorkedBeforeValue.YES)
```
nathanrussell-moj-digital, line 168: I'm not sure this is right. Whilst it works I think it's masking the actual problem.
As per the comment this should already have an answer, so you shouldn't need to select anything (which we didn't beforehand)

I think you've missed a couple of changes to the `hasWorkedBefore` property in the wiremock stubs.

Have a look here:
https://github.com/ministryofjustice/hmpps-education-and-work-plan-ui/blob/bdbddc8ec69302b074707808ab3ba6605596d141/integration_tests/mockApis/educationAndWorkPlanApi.ts#L906C14-L906C33
and
https://github.com/ministryofjustice/hmpps-education-and-work-plan-ui/blob/bdbddc8ec69302b074707808ab3ba6605596d141/integration_tests/mockApis/educationAndWorkPlanApi.ts#L504

I think if you change both of these from the boolean value to the corresponding enum, I think the test should pass without this change 🤞

PR #667, opened 2024-06-13
RR-819 - Add "In Prison Courses & Qualifications" to "Want to add qualifications" and "Qualifications list" screens

nathanrussell-moj-digital:
This PR is the 9th of the Question Set changes. The target branch for this PR is our integration branch feature/RR-810-question-set-epic

---

This PR adds the prisoner's 'In-Prison Courses & Qualifications' data from Curious to the "Want to add qualifications" and "Qualifications list" screens (RR-819)

As discussed earlier with @james-baker-digital-justice I have also taken the opportunity to change (and simplify) the middleware that retrieves the Functional Skills from Curious so that it does not use the session and instead just puts the data on `res.locals`

#### Prisoner with both Functional Skills assessments and In-Prison Courses
Screenshot 2024-06-13 at 15 45 48

#### Prisoner with Functional Skills assessments but no In-Prison Courses
Screenshot 2024-06-13 at 15 46 04

#### Prisoner with neither Functional Skills assessments nor In-Prison Courses
Screenshot 2024-06-13 at 15 43 26


server/@types/express/index.d.ts
```diff
@@ -44,7 +38,6 @@ declare module 'express-session' {
     prisonerListSortOptions: string
     pageFlowHistory: PageFlow
     pageFlowQueue: PageFlow
-    prisonerFunctionalSkills: FunctionalSkills
```
nathanrussell-moj-digital, line 47: The middleware (below 👇 ) that retrieves functional skills from Curious now does not store the data on the session, hence we can remove it from here 👍

server/routes/induction/update/qualificationsListUpdateController.test.ts
```diff
@@ -76,11 +70,7 @@ describe('qualificationsListUpdateController', () => {
       }
 
       // When
-      await controller.getQualificationsListView(
-        req as undefined as Request,
-        res as undefined as Response,
-        next as undefined as NextFunction,
-      )
+      await controller.getQualificationsListView(req, res, next)
```
james-baker-digital-justice, line 73: nice cleanup

server/routes/routerRequestHandlers/retrieveCuriousInPrisonCourses.test.ts
```diff
@@ -1,160 +1,34 @@
...
-    res.locals = {} as Record<string, unknown>
-  })
-
-  const curiousService = {
-    getPrisonerInPrisonCourses: jest.fn(),
-  }
-
-  const requestHandler = retrieveCuriousInPrisonCourses(curiousService as unknown as CuriousService)
-
-  it('should retrieve prisoner In Prison Courses given In Prison Courses not already on res.locals', async () => {
```
james-baker-digital-justice, line 38: does this mean it was originally on res.locals and was moved to the session?

james-baker-digital-justice: actually this one was never in session, sorry my bad. Just checking but seems like these two tests did the same thing so you combined into one that just checks it is loaded?

server/routes/routerRequestHandlers/retrieveCuriousFunctionalSkills.ts
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ *  Middleware function that returns a Request handler function to look up the prisoner's functional skills from Curious
+ */
+const retrieveCuriousFunctionalSkills = (curiousService: CuriousService): RequestHandler => {
+  return asyncMiddleware(async (req: Request, res: Response, next: NextFunction) => {
+    const { prisonNumber } = req.params
+
+    // Lookup the prisoners functional skills and store in res.locals
+    res.locals.prisonerFunctionalSkills = await curiousService.getPrisonerFunctionalSkills(
```
james-baker-digital-justice, line 13: I assume if curios is unavailable we fail to load the whole page?

james-baker-digital-justice (approved):

server/routes/routerRequestHandlers/retrieveCuriousFunctionalSkills.ts
nathanrussell-moj-digital: No, actually this is something we put a lot of time and effort into preventing TBH 😁 

The thinking is that a given UI page might need to call a handful of APIs (prisoner-search to get the basic prisoner data, our API to get prisoner goals, curious to get functional skills, curious (again!) to get in-prison courses; etc etc). Each page will have it's own requirements for data from x number of APIs.

For some on the page it might be crucial that we get a response. For example the Prisoner List page, where we call prisoner-search to get the list of prisoners. If that fails, well, the whole page is essentially non-operational. (We have custom error pages, so the user is not left with a white screen of death or a pointless 500 page)

Anyway, for other pages, data from a given API call might not be crucial. It is perhaps supplementary data, and whether its there or not does not impact the operation of the page or core service. For example, on the prisoner overview page we display the prisoner goals, but also display Functional Skills and In-Prison Courses from Curious. If that data isnt there (ie. the Curious API call(s) fail) then the page itself is still operational and we shouldnt block the user from doing their task.

The way we do that is in the service layer.
Basically the service classes call the given REST API, but do not return the API's native response type/data. Instead we map it into a view model before returning it out from the service method.
The service methods (should) always return a view model object, and (should) never return the type from the REST API (I put "should" in brackets - we might have a couple of instances where that is not true! 😬 😢 )

Anyway, the view model object has a property called `problemRetrievingData` - eg:
```
  /**
   * A prisoner's Functional Skills, which is made up of a collection of Assessments.
   */
  export interface FunctionalSkills {
    assessments: Array<Assessment>
    prisonNumber: string
  }
```
And the service method has a try/catch that will map the response data from the API into this view model object, ensuring that `problemRetrievingData` is set correctly:
```
  async getPrisonerFunctionalSkills(prisonNumber: string, username: string): Promise<FunctionalSkills> {
    const systemToken = await this.hmppsAuthClient.getSystemClientToken(username)

    try {
      const learnerProfiles = await this.getLearnerProfile(prisonNumber, systemToken)
      return toFunctionalSkills(learnerProfiles, prisonNumber)
    } catch (error) {
      logger.error(`Error retrieving functional skills data from Curious: ${JSON.stringify(error)}`)
      return { problemRetrievingData: true } as FunctionalSkills
    }
  }
```

Lastly, in the nunjucks view layer, we have some simple conditional logic on the `problemRetrievingData` property. EG:
```
<div class="govuk-summary-card app-functional-skills-sidebar">
  <div class="govuk-summary-card__title-wrapper app-summary-card__title-wrapper">
    <h2 class="govuk-summary-card__title govuk-!-font-size-24">Current functional skills assessment scores</h2>
  </div>

    {% if not functionalSkills.problemRetrievingData %}
      <table class="govuk-table govuk-!-margin-bottom-0" id="functional-skills-sidebar-table">
        <caption class="govuk-table__caption govuk-visually-hidden">Current functional skills and assessment information</caption>
        <thead class="govuk-table__head govuk-visually-hidden">
          <tr class="govuk-table__row">
            <th scope="col" class="govuk-table__header">Skill and assessment date</th>
            <th scope="col" class="govuk-table__header">Assessment level</th>
          </tr>
        </thead>
          <tr class="govuk-table__row">
              {{ skill.type | formatFunctionalSkillType }}
              {% if skill.assessmentDate %}
              <br>
              {% endif %}
            </th>
            <td class="govuk-table__cell">
              {% if skill.grade %}{{ skill.grade }}{% else %}No functional skill assessment scores recorded in Curious{% endif %}
            </td>
          </tr>
          {% endfor %}
        </tbody>
      </table>
      <p class="govuk-body govuk-!-margin-top-2 govuk-!-display-none-print">
        <a class="govuk-link" href="../functional-skills" data-qa="view-all-functional-skills-button">View all <span class="govuk-visually-hidden">functional skills</span></a>
      </p>
    {% else %}
      <h3 class="govuk-heading-s" data-qa="functional-skills-sidebar-error-heading">We cannot show these details from Curious right now</h3>
      <p class="govuk-body">Reload the page or try again later. Other parts of this service may still be available.</p>
    {% endif %}
  </div>
</div>
```

server/routes/routerRequestHandlers/retrieveCuriousInPrisonCourses.test.ts
nathanrussell-moj-digital: Yeah, there were actually pointless tests.
The implementation used `res.locals`, but because data on `res.locals` is not held/cached in the same way it would be on the session, these tests are effectively pointless! 🤣

server/routes/routerRequestHandlers/retrieveCuriousInPrisonCourses.ts
```diff
@@ -10,18 +9,9 @@ const retrieveCuriousInPrisonCourses = (curiousService: CuriousService): Request
   return asyncMiddleware(async (req: Request, res: Response, next: NextFunction) => {
     const { prisonNumber } = req.params
 
-    // Lookup the prisoners In Prison Courses and store in res.locals if its either not there, or is for a different prisoner
```
nathanrussell-moj-digital, line 13: This relates to my answer to the comment above - this block of logic was completely pointless because it is using `res.locals` rather than the session!

server/routes/routerRequestHandlers/retrieveCuriousFunctionalSkills.ts
nathanrussell-moj-digital: IE. like this:
Screenshot 2024-06-14 at 08 49 43

PR #670, opened 2024-06-13
RR-846: Fix some missed has worked before booleans

james-baker-digital-justice:


wiremock/mappings/041-ciag-induction-mapping.json
```diff
@@ -21,7 +21,7 @@
       "abilityToWork": ["CARING_RESPONSIBILITIES"],
       "reasonToNotGetWork": ["LIMIT_THEIR_ABILITY"],
       "workExperience": {
-        "hasWorkedBefore": true,
```
nathanrussell-moj-digital, line 24: Blimey, I'd forgotten about the wiremock mappings.
I'm not sure we've kept these up to date of late, so perhaps we should have a chat about why they are there and for what purpose, and whether we still need them.
This change is fine, but we can talk about the future of the mappings some other time 😁

nathanrussell-moj-digital (approved):
LGTM 👍

PR #676, opened 2024-06-18
RR-844 - Basic refactoring of the Induction journeys to remove ability to change question set

nathanrussell-moj-digital:
This is PR 1 of 4 for RR-844 (Remove concept of "Long" and "Short" question sets, and create/update journeys that change between them)

This PR starts us off with the basic refactoring of the induction journeys so that it is no longer possible to update an Induction and switch it between "long" and "short" (and vice versa) via the "Hoping to get work" question.

The bulk of this change is in the controllers & tests etc; but there are a couple of changes in the integration tests (specifically deleting entire cypress tests that previously tested being able to change the question set)


server/@types/viewModels/index.d.ts
```diff
@@ -212,8 +212,4 @@ declare module 'viewModels' {
     pageUrls: Array<string>
     currentPageIndex: number
   }
-
-  export interface UpdateInductionQuestionSet {
```
nathanrussell-moj-digital, line 216: This was essentially the flag that we held on the session that would be used in the controller code to determine if we were in the middle of a "change question set" type journey

james-baker-digital-justice (approved):
LGTM

PR #677, opened 2024-06-18
RR-844 - Removed all logic from Update controllers re: going back to check-your-answers as that is no longer a page in the Update journey

nathanrussell-moj-digital:
This is PR 2 of 4 for RR-844 (Remove concept of "Long" and "Short" question sets, and create/update journeys that change between them)

This PR removes the logic in the update controllers re: being able to go back to check-your-answers, as that is no longer possible.

An update journey can now never end up on Check Your Answers because when we update a question now (bearing in mind that "update" means changing a single answer on a previously saved induction) it will always return to the PLP Education & Training or Work & Interests tab (ie. the page that the Change link was on)
There are now no journey permutations where you can change something and we have to force you through the entire journey again resulting in your ending up on Check Your Answers


james-baker-digital-justice (approved):
LGTM

PR #678, opened 2024-06-18
RR-844 - Refactored all references of 'long' and 'short' question set inductions to just 'induction' in all `server` code

nathanrussell-moj-digital:
This is PR 3 of 4 for RR-844 (Remove concept of "Long" and "Short" question sets, and create/update journeys that change between them)

This PR refactors all the references of "long" or "short" induction to just "induction" in the `server` code

Mainly this is removing the test data builders that build instances of "short" Inductions; and renaming test data builders that build instances of "long" Inductions to just "Induction"
And then updating all the references in all the tests to use the new "Induction" based test data builders 👍


james-baker-digital-justice (approved):
LGTM

PR #681, opened 2024-06-18
RR-855 - Correct routing from Hoping To Get Work as part of the Update journey

nathanrussell-moj-digital:
This PR addresses the routing as part of the Update journey where a prisoner already has an Induction and the CIAG has clicked the Change link for "Hoping to work" on the Work & Interests tab.

When they update the induction by changing this answer; if they have:
* changed the previous answer of Yes to No or Not Sure
  * then we need to clear any previously entered work interests (because they now don't want to work), call the API to save the Induction, and take them back to Work & Interests
* changed the previous answer of No or Not Sure to Yes
  * then we need to take them through the Work Interest Roles and Work Interest Details screens so they can enter the details of the work they would like to do.
  * We should not call the API to update the induction until we have successfully submitted Work Interest Details 
  * On the successful submission of Work Interest Details we need to call the API to update the induction, then take them back to Work & Interests
* not changed their answer at all
  * take them back to Work & Interests without calling the API to save the induction at all

### DIFF
--- a/integration_tests/e2e/induction/createCheckYourAnswersChangeLinks.cy.ts
+++ b/integration_tests/e2e/induction/createCheckYourAnswersChangeLinks.cy.ts
@@ -255,7 +255,7 @@ context(`Change links on the Check Your Answers page when creating an Induction`
       .submitPage()
 
     Page.verifyOnPage(FutureWorkInterestRolesPage)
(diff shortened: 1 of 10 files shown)


james-baker-digital-justice (approved):
LGTM

PR #684, opened 2024-06-19
RR-721: Actions card component

james-baker-digital-justice:
Re-usable actions card nunjucks component and model. Not currently used but designed to be placed on the overview page with add goal and view archived goal actions.

### DIFF
--- a/server/views/components/actions-card/actionsCard.test.ts
+++ b/server/views/components/actions-card/actionsCard.test.ts
@@ -1,0 +1,36 @@
+import nunjucks from 'nunjucks'
+import * as cheerio from 'cheerio'
+import type { Action } from 'viewModels'
+
+nunjucks.configure([
+  'node_modules/govuk-frontend/dist/',
+  'node_modules/@ministryofjustice/frontend/',
+  'server/views/',
+  __dirname,
+])
+describe('Tests for actions card component', () => {
+  it('Should display a link for each action', () => {
+    const actions: Action[] = [
+      { title: 'Do a thing', href: '/thing' },
+      { title: 'Do another thing', href: '/another-thing' },
+    ]
+    const content = nunjucks.render('test.njk', { actions })
+    const $ = cheerio.load(content)
+    const actionListItems = $('.actions-list-list').find('li')
+    expect(actionListItems.length).toStrictEqual(2)
+    expect($(actionListItems[0]).find('a').text()).toStrictEqual('Do a thing')
+    expect($(actionListItems[0]).find('a').attr('href')).toStrictEqual('/thing')
+    expect($(actionListItems[1]).find('a').text()).toStrictEqual('Do another thing')
+    expect($(actionListItems[1]).find('a').attr('href')).toStrictEqual('/another-thing')
+  })
+  it('Support setting data-qa field', () => {
+    const actions: Action[] = [
+      { title: 'Do a thing', href: '/thing', dataQa: 'thing-link' },
+      { title: 'Do another thing', href: '/another-thing', dataQa: 'another-thing-link' },
+    ]
+    const content = nunjucks.render('test.njk', { actions })
+    const $ = cheerio.load(content)
+    expect($('[data-qa=thing-link]').text()).toStrictEqual('Do a thing')
+    expect($('[data-qa=another-thing-link]').text()).toStrictEqual('Do another thing')
+  })
+})
--- a/server/views/components/actions-card/template.njk
+++ b/server/views/components/actions-card/template.njk
@@ -1,0 +1,8 @@
+<div class="actions-list">
+  <ul class="govuk-list actions-list-list">
+    <span class="govuk-heading-m actions-list-header">Actions</span>
+    {% for action in actions %}
+      <li><a class="govuk-link" {% if action.dataQa %} data-qa="{{ action.dataQa }}" {% endif %}href="{{ action.href }}">{{ action.title }}</a></li>
+    {% endfor %}
+  </ul>
+</div>
--- a/server/views/components/actions-card/test.njk
+++ b/server/views/components/actions-card/test.njk
@@ -1,0 +1,6 @@
+<html>
+  <body>
+      {% from "components/actions-card/macro.njk" import actionsCard %}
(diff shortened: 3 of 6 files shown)


james-baker-digital-justice:
This is lifted straight out of CCARD (another HMPPS GDS system) common components but with a white header as per Figma.

server/views/components/actions-card/actionsCard.test.ts
```diff
@@ -0,0 +1,36 @@
+import nunjucks from 'nunjucks'
+import * as cheerio from 'cheerio'
+import type { Action } from 'viewModels'
+
+nunjucks.configure([
+  'node_modules/govuk-frontend/dist/',
+  'node_modules/@ministryofjustice/frontend/',
+  'server/views/',
+  __dirname,
```
james-baker-digital-justice, line 9: Includes set up so that the import in the test should have the same path as using the real component. Hopefully makes it easy to crib off when you want to use the component.

server/views/components/actions-card/template.njk
```diff
@@ -0,0 +1,10 @@
+<div class="govuk-grid-column-one-third">
```
nathanrussell-moj-digital, line 1: I wonder if this component should not include the one-third div wrapper? 🤔 

My thinking is we might want several things in the right hand column (like we do at the moment) and/or we might want to use this component in a place where it is not a one-third column?

Thinking about whether we want other things in that column, we do currently, and this is what we do at the moment (`overviewTabContents.njk`):
```
<div class="govuk-grid-row">
  <div class="govuk-grid-column-two-thirds app-u-print-full-width">
    {% if isPostInduction %}
      {% include './overviewTabPostInduction.njk' %}
    {% else %}
      {% include './overviewTabPreInduction.njk' %}
    {% endif %}
  </div>

  <div class="govuk-grid-column-one-third app-u-print-full-width">
    {% include './functionalSkillsSidebar.njk' %}
    {% include './qualificationsAndAchievementsSidebar.njk' %}
  </div>

</div>
```
It feels right to me that the page defines the layout, and we just place components in in the relevant places?

server/views/components/actions-card/template.njk
```diff
@@ -0,0 +1,10 @@
...
+  <div class="actions-list">
+    <ul class="govuk-list actions-list-list">
+      <span class="govuk-heading-m actions-list-header">Actions</span>
```
nathanrussell-moj-digital, line 4: I'm no expert in this stuff, but I wonder if the markup isnt quite right from a semantics point of view?

The `span`, especially with a class of `govuk-heading-m`, feels like it should be a heading? Perhaps a `h2` ?
And if we're in agreement that it is a heading, then I wonder if it should be outside of the `ul` ?

I know its a different use case and is rendering different content, but this is what we have for the Functional Skills sidebar (which currently sits where the proposed action bar will sit) (`functionalSkillsSidebar.njk`):
```
<div class="govuk-summary-card app-functional-skills-sidebar">
  <div class="govuk-summary-card__title-wrapper app-summary-card__title-wrapper">
    <h2 class="govuk-summary-card__title govuk-!-font-size-24">Current functional skills assessment scores</h2>
  </div>

    {% if not functionalSkills.problemRetrievingData %}
      <table class="govuk-table govuk-!-margin-bottom-0" id="functional-skills-sidebar-table">
        <caption class="govuk-table__caption govuk-visually-hidden">Current functional skills and assessment information</caption>
        <thead class="govuk-table__head govuk-visually-hidden">
          <tr class="govuk-table__row">
            <th scope="col" class="govuk-table__header">Skill and assessment date</th>
            <th scope="col" class="govuk-table__header">Assessment level</th>
          </tr>
        </thead>
          <tr class="govuk-table__row">
              {{ skill.type | formatFunctionalSkillType }}
              {% if skill.assessmentDate %}
              <br>
              {% endif %}
            </th>
            <td class="govuk-table__cell">
              {% if skill.grade %}{{ skill.grade }}{% else %}No functional skill assessment scores recorded in Curious{% endif %}
            </td>
          </tr>
          {% endfor %}
        </tbody>
      </table>
      <p class="govuk-body govuk-!-margin-top-2 govuk-!-display-none-print">
        <a class="govuk-link" href="../functional-skills" data-qa="view-all-functional-skills-button">View all <span class="govuk-visually-hidden">functional skills</span></a>
      </p>
    {% else %}
      <h3 class="govuk-heading-s" data-qa="functional-skills-sidebar-error-heading">We cannot show these details from Curious right now</h3>
      <p class="govuk-body">Reload the page or try again later. Other parts of this service may still be available.</p>
    {% endif %}
  </div>
</div>
```

server/views/components/actions-card/template.njk
james-baker-digital-justice: Done

server/views/components/actions-card/test.njk
```diff
@@ -0,0 +1,6 @@
+<html>
+  <body>
+      {% from "components/actions-card/macro.njk" import actionsCard %}
+      {{ actionsCard(actions) }}
```
nathanrussell-moj-digital, line 4:
```
actionsCard(actions)
```
By passing in `actions` as an argument, does this mean the controller (or something?) is responsible for putting an array of actions in the data for the `res.render` call? 🤔 

I can see that working for things that the controller/routes are responsible for providing to the page (eg: prisoner data etc); but does this feel right for a list of actions that won't change? 🤔

james-baker-digital-justice: You can also inline when it's hardcoded like our list of actions will be. The reason it's like this is so the model is exercised, and thus guaranteed to be type safe.

nathanrussell-moj-digital: Oh yeah, good point 👍

nathanrussell-moj-digital (approved):
I think this is great, especially the concept of it being a discrete testable component - we need to be doing more of that in PLP, so we can definitely move towards this style/pattern 👍 

Re: the markup, we can change it as we develop the feature 👍 
I think with the combination of this and my feature toggle PR we can start placing this on the page to see what it looks like etc, and move on from there 👍 

Lets merge it 😁

PR #687, opened 2024-06-20
RR-863: Fix Work Interests last updated for old short question set

james-baker-digital-justice:
If the induction was performed on the old question set and they answered no to "hoping to work" then the last updated field was not displaying properly as future work interests are not entered. Change to use the last updated of hoping to work if future work interests are not present.


nathanrussell-moj-digital (approved):
LGTM 👍

PR #688, opened 2024-06-20
RR-862: Fix Work has worked before for old short question set induction

james-baker-digital-justice:
If the induction was performed on the old question set and they answered no to "hoping to work" then the last updated field was not displaying properly for work experience. Now defaults to has worked before last updated and you can also add the work experience if you so choose as you would be able to on a new induction.


nathanrussell-moj-digital (approved):
LGTM 👍

PR #689, opened 2024-06-20
RR-864 - Fixed bug when updating qualifications where induction has no qualifications to start with

nathanrussell-moj-digital:
This PR fixes a routing bug with the "Change" link for qualifications on existing inductions (any induction, whether that be new question set, short or long)

If the Induction has qualifications to start with, then clicking the Change link should take the user to the Qualifications List screen where they can add/remove qualifications

If the Induction has no qualifications to start with, then clicking the Change link should take the user to the Qualification Level screen where then can enter the first qualification

### DIFF
--- a/integration_tests/pages/overview/EducationAndTrainingPage.ts
+++ b/integration_tests/pages/overview/EducationAndTrainingPage.ts
@@ -5,7 +5,6 @@ import InPrisonCoursesAndQualificationsPage from '../inPrisonCoursesAndQualifica
 import InPrisonTrainingPage from '../induction/InPrisonTrainingPage'
 import HighestLevelOfEducationPage from '../induction/HighestLevelOfEducationPage'
 import AdditionalTrainingPage from '../induction/AdditionalTrainingPage'
-import QualificationsListPage from '../induction/QualificationsListPage'
 
 /**
  * Cypress page class representing the Education And Training tab of the Overview Page
@@ -111,9 +110,9 @@ export default class EducationAndTrainingPage extends Page {
     return Page.verifyOnPage(AdditionalTrainingPage)
   }
 
-  clickToChangeEducationalQualifications(): QualificationsListPage {
+  clickToChangeEducationalQualifications<T extends Page>(constructor: new () => T): T {
     this.educationalQualificationsChangeLink().click()
-    return Page.verifyOnPage(QualificationsListPage)
+    return Page.verifyOnPage(constructor)
   }
 
   doesNotHaveLinkToViewAllCourses(): EducationAndTrainingPage {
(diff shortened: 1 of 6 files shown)


integration_tests/pages/overview/EducationAndTrainingPage.ts
```diff
@@ -111,9 +110,9 @@ export default class EducationAndTrainingPage extends Page {
     return Page.verifyOnPage(AdditionalTrainingPage)
   }
 
-  clickToChangeEducationalQualifications(): QualificationsListPage {
+  clickToChangeEducationalQualifications<T extends Page>(constructor: new () => T): T {
```
nathanrussell-moj-digital, line 113: I had to make this change because clicking the Change link now wont always return the Qualifications List page as it used to.

server/routes/induction/common/qualificationLevelController.ts
```diff
@@ -14,10 +14,8 @@ export default abstract class QualificationLevelController extends InductionCont
     res: Response,
     next: NextFunction,
   ): Promise<void> => {
-    const { pageFlowHistory, prisonerSummary } = req.session
-    if (!pageFlowHistory) {
```
nathanrussell-moj-digital, line 18: This was just wrong! and didnt work for our new simplified question set journey
I think it was there to support a feature of the previous short/long journeys, but was getting in the way of what we want to do now. 
Removing it broke no tests, so I can only assume we removed the relevant tests for this when we simplified the question set but didnt remove this part of the implementation

server/routes/induction/update/qualificationsListUpdateController.ts
```diff
@@ -17,23 +17,9 @@ export default class QualificationsListUpdateController extends QualificationsLi
   getBackLinkUrl(req: Request): string {
     const { prisonNumber } = req.params
     const { pageFlowHistory } = req.session
-    if (pageFlowHistory) {
```
nathanrussell-moj-digital, line 20: This TODO was a hangover from the previous (very complicated) long and short journeys.
The replacement code is what we do on all the other pages in the new question set, and didnt break any tests
We know that we do test back link locations pretty thoroughly at the moment, so I am confident that the existing tests are correct/sufficient; and that the previous implementation worked but was overkill

server/views/pages/overview/partials/educationAndTrainingTab/_qualificationsAndEducationHistory.njk
```diff
@@ -15,7 +15,14 @@
           <h3 class="govuk-heading-s app-summary-card__change-link__heading">Educational qualifications</h3>
           {%- set educationalQualificationsChangeLinkUrl -%}
             {%- if featureToggles.induction.update.enabled -%}
-              /prisoners/{{ prisonerSummary.prisonNumber }}/induction/qualifications
+              {# The target of the change link is different depending on whether the induction has qualifications already recorded on not #}
```
nathanrussell-moj-digital, line 18: This is the crux of the actual fix

james-baker-digital-justice (approved):
LGTM

PR #691, opened 2024-06-24
RR-865 - Added question to capture reason why "has worked before" is "not relevant"

nathanrussell-moj-digital:
This PR adds the conditional "reason" field for "Has worked before" when the user answers it as Not Relevant

### DIFF
--- a/server/@types/educationAndWorkPlanApi/index.d.ts
+++ b/server/@types/educationAndWorkPlanApi/index.d.ts
@@ -5,48 +5,171 @@
 
 export interface paths {
   '/inductions/{prisonNumber}': {
+    parameters: {
+      query?: never
+      header?: never
+      path?: never
+      cookie?: never
+    }
     get: operations['getInduction']
     put: operations['updateInduction']
     post: operations['createInduction']
+    delete?: never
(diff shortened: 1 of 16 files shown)


server/@types/educationAndWorkPlanApi/index.d.ts
```diff
@@ -1821,13 +1993,19 @@ export interface operations {
...
       }
+      cookie?: never
     }
+    requestBody?: never
     responses: {
       /** @description OK */
       200: {
+        headers: {
+          [name: string]: unknown
+        }
```
james-baker-digital-justice, line 2008: i assume all this is related to some renovate/dependabot dependency update?

james-baker-digital-justice (approved):
LGTM

server/@types/educationAndWorkPlanApi/index.d.ts
nathanrussell-moj-digital: Yeah; I think there has been a recent update to `openapi-typescript` which we use to generate these files from the swagger spec; which has resulted in this extra "stuff" 🤷‍♂️

PR #693, opened 2024-06-25
RR-866: Changing "has worked before" to yes should ask for work experience

james-baker-digital-justice:
If selecting yes when answering "has worked before" then we should proceed through work experience types and details screens and update them all together.


nathanrussell-moj-digital (approved):
LGTM 👍

PR #696, opened 2024-06-27
RR-828 - Refactor timeline service and mapping

nathanrussell-moj-digital:
This PR does a little refactoring of the existing Timeline stuff in the UI, before we make the required changes to support the new Goal Archive and Goal Un-archive events.

Basically a few things were in the wrong place


server/@types/viewModels/index.d.ts
```diff
@@ -205,7 +205,7 @@ declare module 'viewModels' {
 
   export interface Prison {
     prisonId: string
-    prisonName: string
+    prisonName?: string
```
nathanrussell-moj-digital, line 208: `prisonName` wont have a value in some cases during the mapping

server/data/mappers/timelineMapper.ts
```diff
@@ -1,13 +1,23 @@
+import { parseISO } from 'date-fns'
```
nathanrussell-moj-digital, line 1: The `timelineMapper` needed some love ...

The purpose of the `timelineMapper` is to map a `TimelineResponse` which comes from the API into a `Timeline` which is a view model object.
For the most part they are field by field identical, with the exception of the event timestamp which is an ISO formatted string in the API response, and is a `Date` in the view model.
That was all fine ...

However, the `TimelineResponse` from the API includes ALL timeline event types. The API will record timeline events for lots of different events, but the UI only supports a subset of those. Therefore to correctly map from the API events to the view model events we need to filter the event types to just those that are supported.
Previously we did this in the service layer as technically it's too late (ie. how can this mapper return a valid view model instance for a given event type if the view model doesnt support that event type)

server/data/mappers/timelineMapper.test.ts
```diff
@@ -1,13 +1,14 @@
-import moment from 'moment'
+import { parseISO } from 'date-fns'
```
nathanrussell-moj-digital, line 1: Probably easier to read the comments on the next file (the implementation) to understand the rational for these changes

server/routes/overview/prepareTimelineForView.test.ts
```diff
@@ -0,0 +1,199 @@
+import { parseISO } from 'date-fns'
+import type { Timeline, TimelineEvent } from 'viewModels'
+import prepareTimelineForView from './prepareTimelineForView'
+
+describe('prepareTimelineForView', () => {
```
nathanrussell-moj-digital, line 5: This looks like a new file but is actually a rename and move of `timelineResolver.test.ts`

server/routes/overview/prepareTimelineForView.ts
```diff
@@ -1,8 +1,15 @@
 import type { Timeline, TimelineEvent } from 'viewModels'
-import moment from 'moment'
-import dateComparator from './dateComparator'
+import { subSeconds } from 'date-fns'
```
nathanrussell-moj-digital, line 2: There was a lot wrong here!
The original filename was `timelineResolver.ts` with an exported function of `filterTimelineEvents`. It didn't "resolve" anything (FWIW I don't like that term in file/function names - we inherited it from elsewhere); and the exported function didn't filter anything!

What it does do is prepare the `Timeline` (a view model type) for rendering in the nunjucks template.
Our UI designs and requirements for the timeline have some very specific requirements, especially with multiple goals that were all created at the same time

server/services/timelineService.ts
```diff
@@ -1,14 +1,9 @@
 import type { Timeline, TimelineEvent } from 'viewModels'
-import type { TimelineEventResponse } from 'educationAndWorkPlanApiClient'
 import EducationAndWorkPlanClient from '../data/educationAndWorkPlanClient'
-import { toTimeline } from '../data/mappers/timelineMapper'
+import toTimeline from '../data/mappers/timelineMapper'
```
nathanrussell-moj-digital, line 3: This is where we used to do the filtering of event types. It's now done in the mapper above ☝️

server/services/timelineService.ts
```diff
@@ -18,11 +13,13 @@ export default class TimelineService {
   async getTimeline(prisonNumber: string, token: string, username: string): Promise<Timeline> {
     try {
       const timelineResponse = await this.educationAndWorkPlanClient.getTimeline(prisonNumber, token)
-      timelineResponse.events = timelineResponse.events.filter(filterTimelineEvents)
 
       const timeline = toTimeline(timelineResponse)
-      timeline.events = await this.addPrisonNameToPrisons(timeline.events, username)
-      return timeline
+      const timelineWithPrisonNamesPopulated: Timeline = {
```
nathanrussell-moj-digital, line 18: I'm not a fan of mutating properties on objects if we can help it (I'm surprised the linter didn't moan TBH)
(it's especially a problem with sorted arrays, and that has caught us out a couple of times elsewhere)

james-baker-digital-justice (approved):
LGTM thanks for making the changes

PR #697, opened 2024-06-27
RR-825: Archive a goal UI

james-baker-digital-justice:
Add pages to select a reason to archive a goal and confirm before archiving it. You can reach the page from the goal list (if the feature toggle is on). The goal list shows archived goals at the moment which will be addressed in subsequent changes.


integration_tests/e2e/goal/archiveGoal.cy.ts
```diff
@@ -0,0 +1,142 @@
...
+    cy.task('getActionPlan')
+    cy.task('stubLearnerProfile')
+    cy.task('stubLearnerEducation')
+    cy.task('archiveGoal')
+  })
+
+  it('should be able to navigate directly to archive goal page', () => {
+    // Given
+    const prisonNumber = 'G6115VJ'
+    const goalReference = '10efc562-be8f-4675-9283-9ede0c19dade'
```
nathanrussell-moj-digital, line 31: I'm not suggesting we change this, but I'm guessing the `goalReference` comes from the stub via `cy.task('getActionPlan')` ?
Its a little bit of a magic number, but I cant think of a better way of doing it without too much effort (unless we pass the goal reference into the stub - and this is already a very large PR)

integration_tests/e2e/goal/archiveGoal.cy.ts
```diff
@@ -0,0 +1,142 @@
...
+      .enterReason('Just because...')
+      .submitPage()
+
+    // When
+    const reviewPage = Page.verifyOnPage(ReviewArchiveGoalPage)
+    reviewPage.clickNo()
+
+    // Then
+    Page.verifyOnPage(OverviewPage)
+  })
```
nathanrussell-moj-digital, line 113: Minor; perhaps we could also verify the API was NOT called in this case?
We have a cypress command that should do this, and I'm pretty sure we have a couple of cases where it's used
```
Cypress.Commands.add('wiremockVerifyNoInteractions', (requestPatternBuilder: RequestPatternBuilder) => {
  return cy.wrap(verify(0, requestPatternBuilder)).should('be.true')
})
```

server/data/mappers/archiveGoalMapper.test.ts
```diff
@@ -0,0 +1,24 @@
+import type { ArchiveGoalDto } from 'dto'
+import type { ArchiveGoalRequest } from 'educationAndWorkPlanApiClient'
+import ReasonToArchiveGoalValue from '../../enums/ReasonToArchiveGoalValue'
+import toArchiveGoalRequest from './archiveGoalMapper'
+
+describe('archiveGoalMapper', () => {
+  it('should map from DTO to request object', () => {
+    const prisonNumber = 'FOO'
```
nathanrussell-moj-digital, line 8: Minor, but we try to use test data that is representative of the use case
In this case `FOO` would not be a valid prison number. Can we change it to something that fits the recognised prison number pattern please? Something like `A1234BC` ?

server/middleware/auditMiddleware.ts
```diff
@@ -23,6 +23,10 @@ const pageViewEventMap: Record<string, Page> = {
   '/plan/:prisonNumber/goals/:goalReference/update': Page.UPDATE_GOALS,
   '/plan/:prisonNumber/goals/:goalReference/update/review': Page.UPDATE_GOALS_REVIEW,
 
+  // Archive goals
```
nathanrussell-moj-digital, line 26: Ah, I'm glad you found this. I forgot to tell you about this (its quite a recent addition that Lee added)
Without this change I suspect you would have been getting 404's when trying to request your new pages?

server/routes/archiveGoal/archiveGoalController.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    const errors = validateArchiveGoalForm(archiveGoalForm)
+    if (errors.length > 0) {
+      return res.redirectWithErrors(`/plan/${prisonNumber}/goals/${goalReference}/archive`, errors)
+    }
+
+    return res.redirect(`/plan/${prisonNumber}/goals/${goalReference}/archive/review`)
+  }
+
+  getReviewArchiveGoalView: RequestHandler = async (req, res, next): Promise<void> => {
```
nathanrussell-moj-digital, line 56: I don't think this is necessarily wrong (and I'm not requesting a change here, more a discussion point); but I'm surprised you've put the "review" form GET and PUT methods in the same class as the main form GET and PUT methods.
Like I said, I'm not saying it's wrong, and we do have cases where we've done this type of thing already, but we also have cases where we've split it out in favour of a controller per "page" - IE. a controller that just has a single GET and PUT handler method.

I don't mind either way, but it was your unit test above that made me realise that's perhaps what you'd done here, because there is a lot in it! 😬

server/routes/archiveGoal/archiveGoalController.test.ts
```diff
@@ -0,0 +1,258 @@
...
+    req.params = {} as Record<string, string>
+
+    req.params.prisonNumber = prisonNumber
+    req.params.goalReference = goalReference
+    req.session.prisonerSummary = prisonerSummary
+
+    errors = []
+  })
+  describe('getArchiveGoalView', () => {
+    it('Should load the page with default archive goal form for the reference', async () => {
```
nathanrussell-moj-digital, line 61: Very minor, but the method name including `with default archive goal form` would suggest there is a case/code path where you can get a non-default form in some way?

server/routes/archiveGoal/archiveGoalController.test.ts
```diff
@@ -0,0 +1,258 @@
...
+      // When
+      await controller.submitArchiveGoalForm(
+        req as undefined as Request,
+        res as undefined as Response,
+        next as undefined as NextFunction,
+      )
+
+      // Then
+      expect(res.redirect).toHaveBeenCalledWith(`/plan/${prisonNumber}/goals/${goalReference}/archive/review`)
+      expect(mockedValidateArchiveGoalForm).toHaveBeenCalledWith(archiveGoalForm)
```
nathanrussell-moj-digital, line 147: Do we also need to assert that the form is on the session? (ie. so it can be used by the next controller method)
Is it worth adding an assertion like this (which I took from your next test scenario)
```
expect(req.session.archiveGoalForm).toEqual(expectedArchiveGoalForm)
```

server/routes/archiveGoal/archiveGoalController.test.ts
```diff
@@ -0,0 +1,258 @@
...
+      // Then
+      expect(res.redirectWithErrors).toHaveBeenCalledWith(
+        `/plan/${prisonNumber}/goals/${goalReference}/archive`,
+        errors,
+      )
+      expect(req.session.archiveGoalForm).toEqual(expectedArchiveGoalForm)
+    })
+  })
+  describe('getReviewArchiveGoalView', () => {
+    it('should get the view with the form', async () => {
```
nathanrussell-moj-digital, line 175: This test is based around the `ArchiveGoalForm` being on the session (ie. the previous page left it there)

Do we also need a test for when the form is not on the session? (ie. the user has book marked the page / arrived on the page unexpectedly?) In this case I think I'd expect the user to be redirected back to the main Archive Goal page (ie. the page that logically comes before this one)

server/routes/archiveGoal/archiveGoalController.test.ts
```diff
@@ -0,0 +1,258 @@
...
+        res as undefined as Response,
+        next as undefined as NextFunction,
+      )
+
+      // Then
+      expect(res.render).toHaveBeenCalledWith('pages/goal/archive/review', expectedView)
+    })
+  })
+
+  describe('getArchiveGoalReviewForm', () => {
```
nathanrussell-moj-digital, line 200: Typo? Should this be `describe('submit....`) ?

server/routes/archiveGoal/archiveGoalController.test.ts
```diff
@@ -0,0 +1,258 @@
...
+        res as undefined as Response,
+        next as undefined as NextFunction,
+      )
+
+      // Then
+      expect(educationAndWorkPlanService.archiveGoal).toHaveBeenCalledWith(expectedArchiveGoalDto, 'some-token')
+      expect(next).toHaveBeenCalledWith(expectedError)
+      expect(req.session.archiveGoalForm).toBeUndefined()
+    })
+  })
```
nathanrussell-moj-digital, line 257: Do we need a test to cover the scenario where the user has clicked No, in which case we want to assert they are redirected back to the Overview page and we have not called the API ?

server/routes/archiveGoal/archiveGoalFormValidator.test.ts
```diff
@@ -0,0 +1,84 @@
...
+    // When
+    const errors = validateArchiveGoalForm(form)
+
+    // Then
+    expect(errors).toEqual([{ href: '#reasonOther', text: 'Enter the reason you are archiving the goal' }])
+  })
+
+  it('other reason should be 200 chars or less', () => {
+    // Given
+    const anInvalidReasonOther = new Array(202).join('x')
```
nathanrussell-moj-digital, line 50: I'm confused by this test.
It looks like we setup the data with the answer for the OTHER field as 202 characters in length:
```
const anInvalidReasonOther = new Array(202).join('x')
```
But then the error message suggests we entered 201 characters?
```
    expect(errors).toEqual([
      { href: '#reasonOther', text: 'The reason is limited to 200 characters but you entered 201' },
    ])
```

server/routes/archiveGoal/archiveGoalFormValidator.test.ts
```diff
@@ -0,0 +1,84 @@
...
+      title: 'Learn French',
+      reason: ReasonToArchiveGoalValue.OTHER,
+      reasonOther: aValidReasonOther,
+    }
+
+    // When
+    const errors = validateArchiveGoalForm(form)
+
+    // Then
+    expect(aValidReasonOther).toHaveLength(200)
```
nathanrussell-moj-digital, line 81: Similar comment to above in that I dont understand where we are apparently loosing a character from the length:
```
const aValidReasonOther = new Array(201).join('x')
```
and
```
expect(aValidReasonOther).toHaveLength(200)
```

But also I don't understand why we are making this assertion (on the length of the OTHER field) - we know how long it is (201) because we set it up on line 69.
There's nothing in the implementation that changes this value is there? 😕

server/routes/archiveGoal/archiveGoalFormValidator.ts
```diff
@@ -0,0 +1,34 @@
...
+}
+
+const validateReasonOther = (archiveGoalForm: ArchiveGoalForm): Array<string> => {
+  const errors: Array<string> = []
+
+  const { reason, reasonOther } = archiveGoalForm
+  if (reason === ReasonToArchiveGoalValue.OTHER && !reasonOther) {
+    errors.push('Enter the reason you are archiving the goal')
+  } else if (reason === ReasonToArchiveGoalValue.OTHER && reasonOther.length > 200) {
+    errors.push(`The reason is limited to 200 characters but you entered ${reasonOther.length}`)
```
nathanrussell-moj-digital, line 30: Where did we get these error messages from? 🤔 
I can't seem them in the ticket (though I didnt look particularly hard, so they might be there)
But I'm particularly interested in the length limit one, in that I dont think we have any other validation messages that say "the limit is x but you entered y" 🤔 
It seems unusual/different terminology to what we do elsewhere I think

Have a look at `stepTitleValidator` as an example, but this is more like our current terminology/pattern:
```
const MAX_STEP_TITLE_LENGTH = 512

export default function validateStepTitle(title?: string): Array<string> {
  const errors: Array<string> = []

  if (!title) {
    errors.push('Enter the step needed to work towards the goal')
  } else if (title.length > MAX_STEP_TITLE_LENGTH) {
    errors.push(`The step description must be ${MAX_STEP_TITLE_LENGTH} characters or less`)
  }

  return errors
}
```

My advice would be to check with the BAs (or dare I say it the designer/content designer, though you might regret that particular avenue of investigation! 🤣 )

server/routes/archiveGoal/index.ts
```diff
@@ -0,0 +1,27 @@
+import { Router } from 'express'
+import { Services } from '../../services'
+import ArchiveGoalController from './archiveGoalController'
+import { checkUserHasEditAuthority } from '../../middleware/roleBasedAccessControl'
+import asyncMiddleware from '../../middleware/asyncMiddleware'
+
+/**
+ * Route definitions for the pages relating to Updating A Goal
```
nathanrussell-moj-digital, line 8: Minor copy/paste typo

server/routes/archiveGoal/index.ts
```diff
@@ -0,0 +1,27 @@
...
+    asyncMiddleware(archiveGoalController.submitArchiveGoalForm),
+  ])
+  router.use('/plan/:prisonNumber/goals/:goalReference/archive/review', [checkUserHasEditAuthority()])
+  router.get('/plan/:prisonNumber/goals/:goalReference/archive/review', [
+    asyncMiddleware(archiveGoalController.getReviewArchiveGoalView),
+  ])
+  router.post('/plan/:prisonNumber/goals/:goalReference/archive/review', [
+    asyncMiddleware(archiveGoalController.submitReviewArchiveGoal),
+  ])
+}
```
nathanrussell-moj-digital, line 27: Interested in your thoughts here re: best practice / what you've done else where ....

As you've no doubt found, we have a lot of router setup functions like this one (in fact most are like this one).
IE. the router has already been defined by `app.ts`, and its passed into the function where we then add new route handlers etc to it 👍 

There is a different pattern that Lee suggested was more "standard", both in terms of express best practice, and also other HMPPS projects.
The pattern he suggested is where router functions like this do not take a router as an arg, but instead new one up internally, set it up as we are here, then return it. And the calling function (`app.ts`) uses it in an `app.use(....` construct
I think we only have 1 like this, for the new (currently unused) Notes page. See `routes/notes/index.ts`

Thoughts?
(FWIW I'm quite happy to stick to the pattern we already have and that you've used here, but I'm always open to new suggestions/ideas etc 🤔 )

server/routes/archiveGoal/mappers/archiveGoalFormToDtoMapper.test.ts
```diff
@@ -0,0 +1,24 @@
+import type { ArchiveGoalDto } from 'dto'
+import type { ArchiveGoalForm } from 'forms'
+import ReasonToArchiveGoalValue from '../../../enums/ReasonToArchiveGoalValue'
+import toArchiveGoalDto from './archiveGoalFormToDtoMapper'
+
+describe('archiveGoalFormToDtoMapper', () => {
+  it('should map from form to DTO object', () => {
+    const prisonNumber = 'FOO'
```
nathanrussell-moj-digital, line 8: As above re: using a prison number that is more representative of real data please

server/views/pages/goal/archive/reason.njk
```diff
@@ -0,0 +1,91 @@
...
+{% block beforeContent %}
+  {% include "../../../partials/breadcrumb.njk" %}
+{% endblock %}
+
+{% block content %}
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+
+      <h1 class="govuk-heading-l">Select a reason for archiving this goal</h1>
+      <section class="archive-goal-summary">
```
nathanrussell-moj-digital, line 15: I guess there is not a standard GDS component for this grey banner 🤷‍♂️ 

This is one of the areas where we need to be careful, as the designers just "design stuff" and it isn't always fully GDS compliant (not to say they should only ever use GDS components, more we've learnt its usually a red flag when they design stuff that does not have a standard component)

(This is more of an FYI than a comment or request to change 👍 )

server/views/pages/goal/archive/reason.test.ts
```diff
@@ -0,0 +1,105 @@
...
+
+const njkEnv = nunjucks.configure([
+  'node_modules/govuk-frontend/dist/',
+  'node_modules/@ministryofjustice/frontend/',
+  'server/views/',
+  __dirname,
+])
+njkEnv.addFilter('formatReasonToArchiveGoal', formatReasonToArchiveGoalFilter)
+njkEnv.addFilter('findError', findErrorFilter)
+
```
nathanrussell-moj-digital, line 17: Re: my comment above about white space ... this test is hard to read 😢

integration_tests/e2e/goal/archiveGoal.cy.ts
james-baker-digital-justice: Yeah, TBH I was a bit concerned with copying this UUIDs everywhere but kept it similar to update goal. It's also prevalent in the test data builders but I'm not sure what to do about it.

assets/scss/components/_archive-goals.scss
```diff
@@ -0,0 +1,9 @@
+.archive-goal-summary {
+  background-color: govuk-colour("light-grey");
+  padding: 25px;
+  margin-bottom: 25px;
+
+  h3 {
+    margin-bottom: 10px;
```
nathanrussell-moj-digital, line 7: Rather than hardcoding pixels I wonder if it's better to leverage the standard GDS styles on the component HTML elements themselves (which will be responsive with associated media queries etc where necessary)
EG. in the nunjucks, on the h3, how about `<h3 class="govuk-!-margin-bottom-2">` (or whatever the right number is)
That way you are using GDS styles, you are not hardcoding pixel dimensions, and then this sass becomes a "style overrides for things I cannot control with standard GDS styles"

server/middleware/auditMiddleware.ts
james-baker-digital-justice: yeah spent a bit too long figuring it out TBH. There was a hint in the logs but though I had the mappings wrong for a while!

server/views/pages/overview/partials/overviewTab/_goalsList.njk
```diff
@@ -12,11 +12,11 @@
           <li class="govuk-summary-card__action">
             <a class="govuk-link" href="/plan/{{ prisonerSummary.prisonNumber }}/goals/{{ goal.goalReference }}/update" data-qa="goal-{{ goal.sequenceNumber }}-update-button">Update<span class="govuk-visually-hidden"> goal {{ goal.sequenceNumber }}</span></a>
           </li>
-          {# Not MVP, Jira to created for archiving goals
```
nathanrussell-moj-digital, line 15: Blimey, I'd completely forgotten we'd provisionally added this ages ago! 🤣

server/routes/archiveGoal/archiveGoalController.ts
nathanrussell-moj-digital: Yup, thats cool 👍

integration_tests/e2e/goal/archiveGoal.cy.ts
nathanrussell-moj-digital: Yeah, dont worry about it for now
Like you say, its a problem that is prevalent in a few places, so perhaps we look at addressing it as a tech-debt thing at some point in the future 👍

server/routes/archiveGoal/archiveGoalController.test.ts
nathanrussell-moj-digital: Scrap that! 😁 
I've just seen the nunjucks for this template and can see the No button is an anchor link, so doesnt actually submit the form to the controller method at all 👍

server/routes/archiveGoal/archiveGoalFormValidator.test.ts
nathanrussell-moj-digital: Would `const anInvalidReasonOther = 'x'.repeat(201)` be clearer ? 🤔

server/routes/archiveGoal/archiveGoalController.test.ts
james-baker-digital-justice: That doesn't happen within the controller, the "no" button is just a link to overview. There is a cypress test covering that flow

server/routes/archiveGoal/archiveGoalFormValidator.test.ts
james-baker-digital-justice: It's a weird nuance of join. We are joining 202 empty elements with x so you don't get one for the first/last element, e.g., <empty>x<empty>x<empty> is a list of 3 that produce 2 x's. I've realised I can replace it with padString which is a bit simpler to read.

server/routes/archiveGoal/archiveGoalFormValidator.test.ts
james-baker-digital-justice: No I just left that assertion there because it was so weird 😂  also replaced with padEnd()

server/@types/educationAndWorkPlanApi/index.d.ts
```diff
@@ -633,11 +667,6 @@ export interface components {
        * @description An optional ISO-8601 date representing the target completion date of the Goal.
        */
       targetCompletionDate: string
-      /**
-       * @example null
-       * @enum {string}
-       */
-      status: 'ACTIVE' | 'COMPLETED' | 'ARCHIVED'
```
nathanrussell-moj-digital, line 640: Now that the `status` field is removed from the `UpdateGoalRequest` API type, I think you're also going to need to remove it from `UpdateGoalDto` and `UpdateGoalForm`, and the associated mappers and test data builders
Sorry! 🤷‍♂️

server/routes/archiveGoal/index.ts
james-baker-digital-justice: Yeah I did find this a bit clunky but it's not a big deal either way for me. Better to be consistent but maybe we do a little clean up and change them. TBH the thing I like the least is the separation of checkUserHasEditAuthority() from the controller definition, it seems like it would be easy to break in a non-obvious way.

server/routes/archiveGoal/mappers/archiveGoalFormToDtoMapper.test.ts
```diff
@@ -0,0 +1,24 @@
...
+    const goalReference = '95b18362-fe56-4234-9ad2-11ef98b974a3'
+    const reason = ReasonToArchiveGoalValue.OTHER
+    const reasonOther = 'Archive it'
+    const form: ArchiveGoalForm = { title: 'A goal', reference: goalReference, reason, reasonOther }
+    const expected: ArchiveGoalDto = { prisonNumber, goalReference, reason, reasonOther }
+    expect(toArchiveGoalDto(prisonNumber, form)).toStrictEqual(expected)
+  })
+  it('should map from DTO to request object with no other reason', () => {
```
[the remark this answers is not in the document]
james-baker-digital-justice: I've added some whitespace and extracted a variable. The tests are so trivial I didn't bother originally. TBH I almost didn't bother with the mappers at all given how trivial the mapping is and the resulting mock-pocalypse in tests of things using the mapper.

server/views/pages/goal/archive/reason.njk
james-baker-digital-justice: Yeah, I tried to use the semantic markup and GDS classes so hopefully it's not too far away from something GDS compliant but looks like figma

server/views/pages/goal/archive/reason.test.ts
james-baker-digital-justice: Added some whitespace, hope that helps

assets/scss/components/_archive-goals.scss
james-baker-digital-justice: Sure, I'll try to change it but no promises, this is not my forte!  I followed the style of the DPS search box (the other grey box I could see in the system) and that uses px for margin.

server/routes/archiveGoal/archiveGoalFormValidator.ts
james-baker-digital-justice: I've updated error messages and checked with Jo and Karen. I'll be adding the char count element as well after chatting with Karen.

assets/scss/components/_archive-goals.scss
```diff
@@ -0,0 +1,9 @@
+.archive-goal-summary {
+  background-color: govuk-colour("light-grey");
+  @include govuk-responsive-padding(4, 'all');
```
nathanrussell-moj-digital, line 3: Nice! 👍

nathanrussell-moj-digital (dismissed):
LGTM; thanks for making the changes 👍

PR #698, opened 2024-06-28
RR-825: Remove status from update goal

james-baker-digital-justice:
This is just to clean up the references to status which cannot be set via update goal APIs.


nathanrussell-moj-digital (approved):
Who doesn't love deleting code? 😁 
A thing of beauty 👍 😁

PR #705, opened 2024-07-09
RR-871 - Refactored to use the system token rather than user token for all PLP API calls

nathanrussell-moj-digital:
**DO NOT MERGE**
This PR is to support Spike [RR-871](https://dsdmoj.atlassian.net/browse/RR-871) and it's [Confluence page write up](https://dsdmoj.atlassian.net/wiki/spaces/RR/pages/4995809432/RR-871+-+Can+the+UI+use+the+system+token+when+calling+the+PLP+API+rather+that+the+UI+token), and is not intended to be merged (unless we decide it is a viable solution)

---

This PR refactors every call to the various PLP API endpoints so they they use the system token (enriched with the username) rather than the user token.

[RR-871]: https://dsdmoj.atlassian.net/browse/RR-871?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/routes/archiveGoal/archiveGoalController.test.ts
+++ b/server/routes/archiveGoal/archiveGoalController.test.ts
@@ -1,5 +1,4 @@
-import { NextFunction, Request, Response } from 'express'
-import type { SessionData } from 'express-session'
+import { Request, Response } from 'express'
 import type { ActionPlan } from 'viewModels'
 import type { ArchiveGoalForm } from 'forms'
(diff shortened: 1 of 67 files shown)


nathanrussell-moj-digital:
Closing this PR as we have now implemented this. No need to keep this as reference any more.

PR #706, opened 2024-07-10
RR-826: View archived goals

james-baker-digital-justice:


server/@types/educationAndWorkPlanApi/index.d.ts
```diff
@@ -1785,6 +1783,13 @@ export interface components {
        */
       reviewDate?: string
     }
+    GetGoalsResponse: {
+      /**
+       * @description A List of zero or more Goals.
+       * @example null
+       */
+      goals: components['schemas']['GoalResponse'][]
+    }
```
james-baker-digital-justice, line 1792: I had to manually create this as there is something wrong with the swagger being generated for this endpoint. I spent a bit of time trying to fix the swagger but can't see what's wrong with it. It is generating "object" rather than referencing the GetGoalsResponse schema.

nathanrussell-moj-digital (changes requested):
I'm really sorry, but this PR is far too big to review! I honestly can't see the wood for the trees! 😢 

From looking at the story I _think_ the deliverable is:
* a new screen that shows just archived goals
* some changes to the Overview screen (which _might_ include filtering the goals to just in progress goals? but that is not clear in the story 🤷‍♂️ )

I'm really struggling to understand what the design & approach is here; sorry!

Can I suggest breaking this down into at least 3 PRs please, where we can review and merge incrementally. I would suggest PRs to:

1. introduce the new client method and service method, and supporting code and tests.
I think this is particularly important given your comment about having to manually tweak the swagger spec. I'm most worried about that TBH. Something is wrong there and we need to get to the bottom of it, otherwise next time somebody imports from the swagger spec we'll lose your manual changes.

2. Introduce the new Archived Goals screen and the link in the action bar to get to it

3. Make the changes to the main Overview screen (if there are any?)

Sorry! 🤷‍♂️ 

(I would also suggest some kind of narrative in the PR description field, and where the PR delivers significant screen changes perhaps a screen shot to give the reviewer some kind of context)

server/@types/viewModels/index.d.ts
```diff
@@ -69,6 +69,11 @@ declare module 'viewModels' {
     problemRetrievingData: boolean
   }
 
+  export interface GoalsOrProblem {
```
nathanrussell-moj-digital, line 72: I'm not keen on this `OrProblem` terminology here. I know you're taking it from what we've started to introduce in the API codebase, but that terminology is part of the inner implementation detail of the API codebase and not exposed in the swagger spec or similar.
I don't think I want to introduce that terminology here in the UI - we've not used it elsewhere

Following the patterns of terminology already used, is it not simply `Goals`? (or have we already used that for something else?)

server/@types/educationAndWorkPlanApi/index.d.ts
nathanrussell-moj-digital: I wonder if you imported the spec from either before I made my change on Tuesday, or from your local branch where you hadn't pulled in my changes in some way?
The hosted swagger spec looks correct, so I think this should import and generate the code correctly. Am more worried that it doesnt TBH and what having made manual changes here might mean for future swagger spec imports. 
I think we should get to the bottom of this 🤔

james-baker-digital-justice: Nope because I had to re-import after you made your change from Array. It's like this if you run it up locally or inspect the JSON on dev.
I agree we should get to the bottom of it though. I spent some time on it earlier in the week and no matter what I changed I couldn't get it to put a reference instead of "object". I was hoping you'd know...

server/@types/viewModels/index.d.ts
james-baker-digital-justice: Well it's actually because it's quite literally `goals` or `problemRetrievingData` which is not terminology I introduced and writing `goals.goals` feels bad. I can rename it to Goals but it's super painful if I split the PR.

server/data/educationAndWorkPlanClient.ts
```diff
@@ -33,6 +35,12 @@ export default class EducationAndWorkPlanClient {
     })
   }
 
+  async getGoalsByStatus(prisonNumber: string, status: GoalStatusValue, token: string): Promise<GetGoalsResponse> {
+    return EducationAndWorkPlanClient.restClient(token).get<GetGoalsResponse>({
+      path: `/action-plans/${prisonNumber}/goals?status=${status}`,
+    })
```
nathanrussell-moj-digital, line 41: Is it worth using the `query` property, rather than concatenating the string? We do it in a couple of other places like this:
```
  async getLearnerEducationPage(prisonNumber: string, token: string, page = 0): Promise<LearnerEductionPagedResponse> {
    return CuriousClient.restClient(token).get<LearnerEductionPagedResponse>({
      path: `/learnerEducation/${prisonNumber}`,
      query: {
        page,
      },
    })
```

server/data/mappers/actionPlanMapper.ts
```diff
@@ -1,19 +1,32 @@
...
+import type { ActionPlanResponse, GetGoalsResponse, GoalResponse, StepResponse } from 'educationAndWorkPlanApiClient'
 import type { ActionPlan, Goal, Step } from 'viewModels'
 import dateComparator from '../../routes/dateComparator'
 
 const toActionPlan = (actionPlanResponse: ActionPlanResponse, problemRetrievingData: boolean): ActionPlan => {
-  const goalReferencesInCreationDateOrder = goalReferencesSortedByCreationDate(actionPlanResponse.goals)
+  const goals = toOrderedGoals(actionPlanResponse.goals)
   return {
     prisonNumber: actionPlanResponse.prisonNumber,
-    goals: actionPlanResponse.goals.map((goal: GoalResponse) => {
```
nathanrussell-moj-digital, line 9: Oh my days! This was some nasty code! I think I wrote that! 😬 🤐 
It totally didn't do what I thought it did! 🤣 It was only seeing your change and how that works that made me realise that! 🤣 
Nice change; thank you 👍

server/data/mappers/actionPlanMapper.ts
```diff
@@ -1,19 +1,32 @@
...
+/**
+ * Sets a goal sequence by creation date (oldest = 1) and returns them in creation order (newest first)
+ */
+function toOrderedGoals(goals: GoalResponse[]): Goal[] {
+  return [...goals]
+    .sort((left: GoalResponse, right: GoalResponse) =>
+      dateComparator(toDate(left.createdAt), toDate(right.createdAt), 'ASC'),
+    )
+    .map((goal, index) => toGoal(goal, index + 1))
+    .reverse()
```
nathanrussell-moj-digital, line 27: It was this call to `.reverse()` that made me look at the diff (over and over again! 🤣 ) to see how my original code was flawed! 🤣.
FWIW, instead of using `.reverse`, you could pass `DESC` into the dateComparator as an arg instead
I guess there's a very minor negligible performance benefit by not calling `reverse`, but I doubt we'd ever see and tangible benefit from it.
And if it wasn't for the call to `reverse` I might not have looked so hard at the code to see where my original code was shonky! 🤣

integration_tests/e2e/goal/archiveGoal.cy.ts
```diff
@@ -142,25 +195,40 @@ context('Archive a goal', () => {
...
+      .hasNumberOfGoals(2)
+      .clickArchiveButtonForFirstGoal()
+
     archiveGoalPage //
       .selectReason(ReasonToArchiveGoalValue.OTHER)
       .enterReason('Just because...')
       .submitPage()
 
-    // When
+    cy.task('getGoalsByStatus', {
```
nathanrussell-moj-digital, line 217: Do you think this adds any value to the cypress test? 🤔 
IE. you're re-setting up the `getGoalsByStatus` stub with fewer goals, so that the 2nd rendering of the Overview page has the "correct" goals, and then you can do the assertions on lines 229 and 230

I get it, but I wonder what it's adding to the cypress test 🤔 . In that, if the implementation were to change in some way and for whatever reason the call to archive the goal changed/failed, then according to the page rendering then it apparently would have worked.
I think the important assertion is the one that verifies the archive API was called with the expected data 👍 

What do you think? We dont do this "re-setting the stubs with updated data for the purpose of view" elsewhere - eg. we dont do it in the various tests that edit goals or inductions .... 🤔

integration_tests/e2e/goal/reviewUpdateGoal.cy.ts
```diff
@@ -18,6 +18,7 @@ context('Review updated goal', () => {
     cy.task('getPrisonerById')
     cy.task('stubGetInduction')
     cy.task('getActionPlan')
+    cy.task('getGoalsByStatus')
```
nathanrussell-moj-digital, line 21: Is the plan/approach that you're deprecating (removing) the calls to the `getActionPlan` stub and replacing with calls to `getGoalsByStatus`?
That appears to be what you're doing everywhere else, but not here? Does this mean that this cypress test needs both stubs setting up for some reason?

As a side note this is exactly why I'm struggling with this PR both in its sheer size, but also not knowing the design/approach you are taking here.
I'm guessing you've replaced `getActionPlan` calls on the Overview page with calls to the new `getGoalByStatus` API. But I wont know that until I get to the diff in the controller which is miles away! 🤣 
Normally we would put that kind of pertinent detail in some notes or narrative in the main PR description.

integration_tests/e2e/goal/unarchiveGoal.cy.ts
```diff
@@ -1,10 +1,17 @@
 import Page from '../../pages/page'
 import UnarchiveGoalPage from '../../pages/goal/UnarchiveGoalPage'
 import OverviewPage from '../../pages/overview/OverviewPage'
-import AuthorisationErrorPage from '../../pages/authorisationError'
-import { putRequestedFor } from '../../mockApis/wiremock/requestPatternBuilder'
+// import AuthorisationErrorPage from '../../pages/authorisationError'
```
nathanrussell-moj-digital, line 4: Unused comments?

integration_tests/e2e/goal/unarchiveGoal.cy.ts
```diff
@@ -22,6 +29,7 @@ context('Unarchive a goal', () => {
     cy.task('getPrisonerById')
     cy.task('stubGetInduction')
     cy.task('getActionPlan')
+    cy.task('getGoalsByStatus')
```
nathanrussell-moj-digital, line 32: As above, this sets up stubs for both `getActionPlan` and `getGoalsByStatus`. Is that right? 🤔

integration_tests/e2e/induction/inductionSecurityScenarios.cy.ts
```diff
@@ -13,7 +13,6 @@ context('Security tests for creating and updating Inductions', () => {
     cy.task('stubCiagInductionList')
     cy.task('stubActionPlansList')
     cy.task('getPrisonerById')
-    cy.task('getActionPlan')
```
nathanrussell-moj-digital, line 16: I'm guessing all these calls to `getActionPlan` are being removed from these cypress tests because these tests dont go via the Overview page and therefore don't need to get the action plan?

integration_tests/e2e/overview/preInductionOverview.cy.ts
```diff
@@ -46,7 +46,7 @@ context('Prisoner Overview page - Pre Induction', () => {
   it('should render prisoner Overview page listing no prisoner goals given prisoner has no goals created pre-induction', () => {
     // Given
     const prisonNumberForPrisonerWithNoGoals = 'A00001A'
-    cy.task('getActionPlan', prisonNumberForPrisonerWithNoGoals)
+    cy.task('getGoalsByStatus404', prisonNumberForPrisonerWithNoGoals)
```
nathanrussell-moj-digital, line 49: I'm not sure about returning a 404 rather than a response that contains an empty array.
I think we need to think about and understand the semantics of that.

I think this test, arguably possibly badly written (I'll take that), might highlight the problem (because it didnt fail!)
Previously, an action plan with no goals returns a 200 response with an empty goals array. And the UI rendering would not display any goals:
```
{# Template to render the prisoner's list of Goals on the Overview tab #}

{% for goal in goals.goals %}
  <div class="govuk-summary-card" data-qa="goal-summary-card">
    <div class="govuk-summary-card__title-wrapper">
      <h3 class="govuk-summary-card__title">Goal {{ goal.sequenceNumber }}
....
```
IE. the collection is empty and the loop would not loop

And we assert that (perhaps badly) in this test with the `.hasNoGoalsDisplayed` method
```
  hasNoGoalsDisplayed(): OverviewPage {
    this.goalSummaryCards().should('not.exist')
    return this
  }
...
  private goalSummaryCards = (): PageElement => cy.get('[data-qa=goal-summary-card]')
```

By returning a 404, unless you are specifically catching and handling it in your service somewhere, then I _think_ `problemRetrievingData` will be `true` and you'll be rendering the "service is down" content:
```
<section data-qa="post-induction-overview">
  {% if not goals.problemRetrievingData %}

    {% include './_actionsHeader.njk' %}
    {% include '../_goalsList.njk' %}

  {% else %}
    <h2 class="govuk-heading-m">Sorry, the service is currently unavailable.</h2>
    <p class="govuk-body">
      Retrieving goals is currently unavailable. Please try again later.
    </p>
  {% endif %}
</section>
```
I think we need to think about the semantics of what "no goals" vs a 404 means 🤔

integration_tests/e2e/goal/unarchiveGoal.cy.ts
```diff
@@ -73,8 +100,7 @@ context('Unarchive a goal', () => {
     )
   })
 
-  // TODO - enable this test when the Archived Goals page is complete
-  it.skip(`should not unarchive goal and display Archived Goals page given user clicks No`, () => {
+  it(`should not unarchive goal and display Archived Goals page given user clicks No`, () => {
```
nathanrussell-moj-digital, line 103: 👍 
Thank you for enabling this test 👍

server/data/educationAndWorkPlanClient.ts
james-baker-digital-justice: Done

server/data/mappers/actionPlanMapper.ts
james-baker-digital-justice: It's actually even more confusing because it numbers them in ASC date order but displays them in DESC date order, hence the reverse. The numbering is going with the new overview page so hopefully we can delete most of this mess then anyway.

integration_tests/e2e/goal/archiveGoal.cy.ts
james-baker-digital-justice: Yeah I tend to agree and wasn't really happy with this test. I kept it as it was what I was using to drive the changes in general and it's the only test that walks the whole flow of overview page > archive reason > confirm > overview page. Open to suggestions on how to improve it.

server/services/educationAndWorkPlanService.ts
```diff
@@ -29,6 +30,20 @@ export default class EducationAndWorkPlanService {
     }
   }
 
+  async getGoalsByStatus(prisonNumber: string, status: GoalStatusValue, token: string): Promise<Goals> {
+    try {
+      const response = await this.educationAndWorkPlanClient.getGoalsByStatus(prisonNumber, status, token)
+      return { goals: toGoals(response), problemRetrievingData: false }
+    } catch (error) {
+      if (error.status === 404) {
+        logger.debug(`No plan created yet so no goals with [${status}] for Prisoner [${prisonNumber}]`)
```
nathanrussell-moj-digital, line 39: Ah ha, this answers my questions above re: whether a 404 is right and whether a 404 would trigger the "the service is down" message. Got it 👍

integration_tests/e2e/overview/preInductionOverview.cy.ts
nathanrussell-moj-digital: Scrap that, I've just found the code in the service method that handles the 404:
```
    } catch (error) {
      if (error.status === 404) {
        logger.debug(`No plan created yet so no goals with [${status}] for Prisoner [${prisonNumber}]`)
        return { goals: undefined, problemRetrievingData: false }
```

Cool 👍

james-baker-digital-justice: The 404 is handled as not an error and initialise an empty array in the service so you don't get the service unavailable text. I made this test return a 404 as it's labeled as pre-induction so this should simulate the behaviour of the real service.

integration_tests/e2e/induction/inductionSecurityScenarios.cy.ts
james-baker-digital-justice: Yeah just attempting to clean it up in the tests so we have a better idea of what actually needs action plan/goals endpoints. I removed them all to see which tests actually needed them and added back as appropriate.

integration_tests/e2e/goal/reviewUpdateGoal.cy.ts
james-baker-digital-justice: Archive and update pages still use the action plan to load the specific goal. I think the best approach would be to add a "get goal" endpoint for those pages and remove getActionPlan entirely but that's obviously beyond the scope of this already large change.

package.json
```diff
@@ -10,7 +10,7 @@
     "compile-sass": "sass --quiet-deps --no-source-map --load-path=node_modules/govuk-frontend/dist --load-path=node_modules/@ministryofjustice/frontend --load-path=. assets/scss/application.scss:./assets/stylesheets/application.css assets/scss/print.scss:./assets/stylesheets/print.css --style compressed",
     "watch-ts": "tsc -w",
     "watch-views": "nodemon --watch server/views -e html,njk -x npm run copy-views",
-    "watch-node": "DEBUG=gov-starter-server* nodemon -r dotenv/config --watch dist/ dist/server.js | bunyan -o short",
+    "watch-node": "DEBUG=gov-starter-server* nodemon -r dotenv/config --inspect --watch dist/ dist/server.js | bunyan -o short",
```
nathanrussell-moj-digital, line 13: This looks like it could be useful/interesting .... what does it do?
(I have no problem with it, just interested what it does - I'm guessing something when we run the app locally in dev mode?)

integration_tests/e2e/goal/reviewUpdateGoal.cy.ts
nathanrussell-moj-digital: Ah, got it 👍 
Thanks 👍

package.json
james-baker-digital-justice: It allows you to attach a debugger in intellij. Just add a run config "Attach to nodejs/Chrome" and use the defaults. The annoying thing is the debugger detaches when you edit something and it restarts the process but I found it useful while trying to figure out what was going on in the sorting code...

server/data/mappers/actionPlanMapper.ts
nathanrussell-moj-digital: Yes, good point about the numbering; and yes, once we can start on the new design and remove the numbering this will be so much easier 👍

nathanrussell-moj-digital (approved):
I'm sorry this has been a slog; maybe in future we can try for smaller more iterative PRs, and to make better use of the main PR description.

Approved ✅

PR #707, opened 2024-07-11
RR-825: Fix archive goal form sticking in session

james-baker-digital-justice:
If you navigated away after submitting the form, either by error or cancelling the archive, when you returned the form was already filled. If you returned on a different goal the wrong goals information was displayed.


nathanrussell-moj-digital (approved):
Good PR description; thank you 👍 

What you've done here looks fine (and we might need to do a similar thing for the un-archive routes etc 🤔 ).

In other places for similar code/reasons we have opted to use middlewares, rather than doing it in the controller.

It's fine as it is, I'm not going to block it, as it allows us to get this feature into test soon 👍 . But we might want to consider whether this is the right approach at some point and potentially refactor at that time

LGTM though 👍

PR #710, opened 2024-07-15
RR-824: turn on archiving feature toggle in all envs

james-baker-digital-justice:
Now we've completed all the tickets for archiving we can turn it on in preparation for a release to prod. Keeping the feature toggle for now in the unlikely event we need to rollback it'll be much faster to turn it off.


nathanrussell-moj-digital (approved):
LGTM 👍

PR #711, opened 2024-07-16
RR-826: Convert goal summary card to component

james-baker-digital-justice:
Converted the goal summary card to a nunjucks component. Added a few unit tests for the bonus features of this approach such as custom classes & attributes as well as the render-if feature for the actions (similar to actions-card). It is otherwise functionally the same but makes some upcoming design changes and known issues with viewing archived goals easier.


server/views/pages/overview/partials/overviewTab/overviewTabPreInduction.njk
```diff
@@ -22,7 +22,7 @@
         html: createInductionBanner
       }) }}
 
-      {% include '../_goalsList.njk' %}
+      {% include './_goalsList.njk' %}
```
james-baker-digital-justice, line 25: I moved this back to only being shared between the overview screens. My plan is to move the sorting out of the service layer and allow the view to decide the order when iterating

server/routes/overview/viewArchivedGoalsController.test.ts
```diff
@@ -35,14 +35,11 @@ describe('viewArchivedGoalsController', () => {
   })
   it('should get view archived goals page', async () => {
     // Given
-    const expectedTab = 'view-archived-goals'
```
nathanrussell-moj-digital, line 38: The PR in general LGTM 👍  (and I will approve it 😁 ), but the removal of the expected tab from here and the view class itself has made me stop and think....

The `tab` property in the view class (and therefore the expected tab in the tests) is there so that we have a view class for each "tab" on the Overview page.
I'd assumed that the new Archived Goals page was a tab on the Overview page, hence the need for the property etc.
So seeing you remove it here made me look at the designs.
You are right, its not a tab on the Overview page, so you are right to remove the property 👍 

But .... if it's not a tab on the Overview page, and is a page in it's own right, then I think this controller and view class etc are in the wrong folder in the source tree (in that it's nothing to do with the Overview page)

Let's go with this PR as it is 👍 , and then do another one to refactor `viewArchivedGoalsController` and associated classes etc into a better folder structure, outside of the `overview` structure 👍

server/views/pages/overview/viewArchivedGoals.njk
nathanrussell-moj-digital, line 1: Same comment as above - I think this template file is in the wrong folder because the Archived Goals page is not a tab on the Overview page.
But we can move it in a different PR 👍

nathanrussell-moj-digital (approved):
LGTM 👍

server/routes/overview/viewArchivedGoalsController.test.ts
james-baker-digital-justice: ok - I left it here on the basis that we're shortly changing the overview and goals/archived/completed will be sub-tabs of the overview. No biggie either way to me.

nathanrussell-moj-digital: Ah, in that case, yes. If they will soon become subtabs of the Overview page, then yes, its all in the right place 😁 
Let's leave it for now 😁 👍

PR #712, opened 2024-07-16
RR-878 - Configure /info endpoint to include activeAgencies

nathanrussell-moj-digital:
This PR adds the `activeAgencies` property to the `/info` endpoint in order to support [Establishment Specific Rollouts](https://dsdmoj.atlassian.net/wiki/spaces/ALIGN/pages/4480959439/Adding+changing+services+on+the+common+header+footer+and+the+homepage#Establishment-specific-rollouts)

We need to do this to be able to support further rollouts where the LSA's in prisons will be given the responsibility for assigning our service role to their DPS users.


james-baker-digital-justice (approved):
LGTM;

PR #715, opened 2024-07-17
RR-826: Sort archived goals by most recently archived

james-baker-digital-justice:
Change the view archived goals page to sort goals by most recently archived which is actually just the last updated date. This is because once it's archived there can be no further updates so the last update would have been marking it as archived.

Refactored the cypress tests a little bit so that view archived goals is now grouped together, we're also asserting the correct goals are loaded for each page rather than updating the wiremock mid test to simulate it.


nathanrussell-moj-digital (approved):
LGTM 👍

PR #718, opened 2024-07-18
RR-826: Populate prison name onto Goals for summary card

james-baker-digital-justice:
Show the name of the prison alongside the name of the person who last updated or archived a goal.

The new getPrisonNamesById method is intended to replace the single prison method as the client code for that became complicated to avoid a race condition where we would submit many requests to the prison api before it is cached.

### DIFF
--- a/server/services/educationAndWorkPlanService.test.ts
+++ b/server/services/educationAndWorkPlanService.test.ts
@@ -15,14 +15,15 @@ import { aValidUpdateGoalRequestWithOneUpdatedStep } from '../testsupport/update
 import { aValidCreateGoalsRequestWithOneGoal } from '../testsupport/createGoalsRequestTestDataBuilder'
 import ReasonToArchiveGoalValue from '../enums/ReasonToArchiveGoalValue'
 import GoalStatusValue from '../enums/goalStatusValue'
+import PrisonService from './prisonService'
 
 jest.mock('../data/educationAndWorkPlanClient')
-
+jest.mock('./prisonService')
 describe('educationAndWorkPlanService', () => {
   const educationAndWorkPlanClient =
     new EducationAndWorkPlanClient() as unknown as jest.Mocked<EducationAndWorkPlanClient>
-
-  const educationAndWorkPlanService = new EducationAndWorkPlanService(educationAndWorkPlanClient)
+  const prisonService = new PrisonService(null, null, null) as unknown as jest.Mocked<PrisonService>
+  const educationAndWorkPlanService = new EducationAndWorkPlanService(educationAndWorkPlanClient, prisonService)
 
   beforeEach(() => {
     jest.resetAllMocks()
@@ -66,6 +67,7 @@ describe('educationAndWorkPlanService', () => {
       const userToken = 'a-user-token'
       const actionPlanResponse = aValidActionPlanResponseWithOneGoal()
       educationAndWorkPlanClient.getActionPlan.mockResolvedValue(actionPlanResponse)
+      prisonService.getAllPrisonNamesById.mockResolvedValue(new Map())
       const expectedActionPlan = aValidActionPlanWithOneGoal()
 
       // When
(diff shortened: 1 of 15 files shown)


server/services/educationAndWorkPlanService.test.ts
```diff
@@ -76,12 +78,27 @@ describe('educationAndWorkPlanService', () => {
       expect(actual).toEqual(expectedActionPlan)
     })
 
-    it('should not get Action Plan given educationAndWorkPlanClient returns an error', async () => {
+    it('should not get Action Plan given prisonService returns an error', async () => {
```
nathanrussell-moj-digital, line 81: I'm not sure by this scenario ... 🤔 

The service call is `getActionPlan`, and in the other test (which I guess you based this one on), if the `educationAndWorkPlanClient` returns an error then clearly we cant get the action plan, so returning an object with `problemRetrievingData` set make sense 👍 

But is the same true for if the `prisonService` returns an error? 🤔  The scenario would be that we have successfully retrieved the action plan (the core data of interest), but we cant augment/decorate it with the prison names. Is that an error (in the same way)? or do we just return the action plan with the prison IDs in it rather than the looked up names?

Based on what we've done elsewhere we would return the action plan with the prison IDs, rather than returning some error state.

What do you think?
(Oh, and FWIW I would not recommend asking the BAs, and definitely not the design team, about this! They will likely want to design some kind of "empty state" (I hate that term!) which will no doubt add delay and complexity. I think we should be able to make a decision between us about what to do in this circumstance 👍 )

server/services/educationAndWorkPlanService.test.ts
```diff
@@ -124,7 +143,23 @@ describe('educationAndWorkPlanService', () => {
       expect(educationAndWorkPlanClient.getGoalsByStatus).toHaveBeenCalledWith(prisonNumber, status, userToken)
       expect(actual).toEqual(expectedResponse)
     })
+    it('should return a problem loading the data if prison ids could not be loaded with status 404', async () => {
```
nathanrussell-moj-digital, line 146: As above, not sure this is a scenario that should stop the core data being displayed on the page 🤔 

Also, is this scenario even possible? Why would the prisonService return a 404? The prisonRegisterClient would return a 404 to the prisonService, but does (should?) the service not handle that error in some way? 🤔

server/services/educationAndWorkPlanService.ts
```diff
@@ -42,6 +47,12 @@ export default class EducationAndWorkPlanService {
       logger.error(`Error retrieving goals with status [${status}] for Prisoner [${prisonNumber}]: ${error}`)
       return { goals: undefined, problemRetrievingData: true }
     }
+    try {
+      const prisonNamesById = await this.prisonService.getAllPrisonNamesById(token)
+      return { goals: toGoals(response, prisonNamesById), problemRetrievingData: false }
+    } catch (error) {
```
nathanrussell-moj-digital, line 53: As discussed in the test, I'm not sure this is right? 🤔

server/services/prisonService.ts
```diff
@@ -38,6 +38,20 @@ export default class PrisonService {
     }
   }
 
+  /**
```
nathanrussell-moj-digital, line 41: I like this implementation and the unit tests above 👍 

Of interest is the fact that the service method does not throw/return an error. It catches any error and returns an empty map 👍 
Thats perfect, but it means the comments I made above re: the educationAndWorkPlanService and whether prisonService can/should throw errors and set an error response probably have some merit?
What do you think? 🤔

james-baker-digital-justice: ha sorry I should've explained this probably. As you mentioned, in reality it never returns a rejected promise and instead returns an empty list so you will just get the prison ID instead (made an executive decision on that one). 

I felt like I should test what happens if the promise is rejected and built up tests before I got to the actual implementation and the service should probably not be aware of the implementation details (maybe there is some javascripty way it can reject, e.g., NPE?). 

I'll change it to try-catch and default to empty list rather than returning a problem retrieving data.

james-baker-digital-justice: Updated to catch and return an empty map

nathanrussell-moj-digital (approved):
LGTM

PR #721, opened 2024-07-19
RR-826: Minor UI tweaks to view archived goals

james-baker-digital-justice:
Increase padding all around to 3 (15px) and add a margin to the subtitle so it aligns correctly with the target image.

Also removed the colon from last updated/archived on label.


nathanrussell-moj-digital (approved):
LGTM 👍

PR #723, opened 2024-07-22
RR-828 - Fix rendering of goal archived reason

nathanrussell-moj-digital:
This PR fixes the rendering of timeline events re: goal archive reasons.

It fixes this:
Screenshot 2024-07-22 at 08 53 43

to this:
Screenshot 2024-07-22 at 08 53 53

### DIFF
--- a/server/views/pages/overview/partials/timelineTab/_timelineEvent-GOAL_ARCHIVED.njk
+++ b/server/views/pages/overview/partials/timelineTab/_timelineEvent-GOAL_ARCHIVED.njk
@@ -9,11 +9,8 @@
   </p>
 
   <div class="moj-timeline__description govuk-!-display-none-print">
-    {% set reasonGoalArchived %}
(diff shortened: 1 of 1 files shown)


server/views/pages/overview/partials/timelineTab/_timelineEvent-GOAL_ARCHIVED.njk
```diff
@@ -9,11 +9,8 @@
   </p>
 
   <div class="moj-timeline__description govuk-!-display-none-print">
-    {% set reasonGoalArchived %}
```
nathanrussell-moj-digital, line 12: It seems the nunjucks implementation/behaviour with it's `set` operator causes this problem.
I know I probably could have fixed this with a `safe` filter, but we don't use it elsewhere (such as with goal and step titles; or the review page when archiving the goal; so I wanted to know why this page was a problem.

I imaging `set` is effectively a string concatenation or similar, and perhaps internally escapes the content before concatenation?

I'm not massively keen on using the `safe` filter anyway (unless we really really need to) as it feels like it could be used as a silver bullet / used unnecessarily , and therefore masks other problems. I'm certainly not keen on using it for user generated input as in this case.

Anyway, whatever `set` is doing internally and/or whether we think `safe` is OK to use in this case, I felt it was easier and cleaner to do away with the `set` and just inline the content 👍

james-baker-digital-justice (approved):
lgtm

PR #724, opened 2024-07-22
RR-883 - Cleanup use of get prison by

james-baker-digital-justice:
Simplifying the lookup of multiple prisons per controller/service by instead retrieving them all at once. This is what was happening behind the scenes anyway.

One further change is required to timeline service but that's a little more involved as it's currently passing the entire prison to the template although it only even uses name & id (and id is only used as a fallback if no name)


nathanrussell-moj-digital (approved):
LGTM 👍

PR #729, opened 2024-07-23
RR-883: Keep other context for prison transfer event

james-baker-digital-justice:
We map the prison id to a name for prison transfer events in the timeline. Currently, the only contextual info is the prison but if we were to add more it would be lost during mapping. Now it keeps any other context info as well as mapping the prison id to name.


nathanrussell-moj-digital (approved):
LGTM 👍

PR #730, opened 2024-07-23
RR-891 - Fix createGoalMapper

nathanrussell-moj-digital:
This PR fixes some incorrect property mapping in `createGoalMapper` that I saw the other day.
The nature of it being incorrect is that it was mapping extra properties that do not exist on the resultant `CreateGoalRequest` type, specifically:

* `prisonNumber`
* `category`

(`category` was an API request body field that existed in a very early iteration of the project; `prisonNumber` should not be mapped into the API request body as it is part of the URI path)

This problem was highlighted by the IDE, but was not a problem at runtime as they were simply extra properties that were being ignored by the API when being sent as part of the Create Goals API call.

This PR fixes the mapper and it's test, and corrects the calls to the mapper via the service and controller.


integration_tests/e2e/goal/createGoals.cy.ts
```diff
@@ -237,14 +237,16 @@ context('Create goals', () => {
         .withRequestBody(
           matchingJsonPath(
             '$[?(@.goals.size() == 2 && ' +
-              "@.goals[0].prisonNumber == 'G6115VJ' && " +
+              "@.goals[0].prisonId == 'MDI' && " +
```
nathanrussell-moj-digital, line 240: `prisonNumber` is no longer in the request body object (and should never have been)
`prisonId` and `targetCompletionDate` were always in the request body but were never tested here.

package.json
```diff
@@ -170,6 +171,7 @@
     "jest-junit": "^16.0.0",
     "jsonwebtoken": "^9.0.2",
     "lint-staged": "^15.2.7",
+    "lodash": "^4.17.21",
```
nathanrussell-moj-digital, line 174: I've added lodash / underscore to support a test I wanted to fix whilst I was here (the next file in the diff 👇 )
I'm not entirely happy with introducing a dependency , but at least it's a big/well maintained and supported one 🤷‍♂️

server/data/educationAndWorkPlanClient.test.ts
```diff
@@ -40,14 +41,14 @@ describe('educationAndWorkPlanClient', () => {
       const prisonNumber = 'A1234BC'
       const systemToken = 'a-system-token'
 
-      const createGoalsRequest = aValidCreateGoalsRequestWitMultipleGoals(prisonNumber)
+      const createGoalsRequest = aValidCreateGoalsRequestWitMultipleGoals()
       const expectedResponseBody = {}
       educationAndWorkPlanApi
-        .post(`/action-plans/${prisonNumber}/goals`, createGoalsRequest)
+        .post(`/action-plans/${prisonNumber}/goals`, requestBody => isEqual(requestBody, createGoalsRequest))
```
nathanrussell-moj-digital, line 47: I was in this test file anyway because I needed to correct the calls to the test data builders and the call to the `createGoals` client method (based on the main changes in this PR)

But intellij was highlighting errors on all of the calls to `nock` that use a request body. EG: this original line:
```
 educationAndWorkPlanApi
        .post(`/action-plans/${prisonNumber}/goals`, createGoalsRequest)
```
intellij highlighted an error on that line because the request body argument (`createGoalsRequest`) does not match  one of the supported method signatures:
Screenshot 2024-07-23 at 09 25 10

Clicking into the `post` method it defines the method sig as:
```
  type InterceptFunction = (
    uri: string | RegExp | { (uri: string): boolean },
    requestBody?: RequestBodyMatcher,
    interceptorOptions?: Options,
  ) => Interceptor
```
and clicking into the `RequestBodyMatcher` type it is defined as:
```
  type RequestBodyMatcher =
    | string
    | Buffer
    | RegExp
    | DataMatcherArray
    | DataMatcherMap
    | { (body: any): boolean }
```

We need to provide an arg that fulfils the last one - a function that takes the received request body as an argument and returns a boolean.
(This is why I've brought in `lodash` in order to get the `isEqual` function - happy to use something else if there is already something suitable in our dependencies)

FWIW I believe these calls to `nock` worked previously by virtue of javascript objects being truthy! 🤮

integration_tests/e2e/goal/createGoals.cy.ts
james-baker-digital-justice: I find prisonNumber to be a bad name anyway. You have prisonNumber and prisonId. One refers to a person, one refers to a place 🙃

james-baker-digital-justice (approved):
LGTM

integration_tests/e2e/goal/createGoals.cy.ts
nathanrussell-moj-digital: Interesting you should say that; we've never liked the name `prisonNumber` either! 
But AFAIK it's the "official" / standard HMPPS terminology for this data item (I got this from Mike Halma when I worked with him here on a previous project)
From the outset we wanted to use the "correct" terminology for things like this, rather than using "un-official" variants such as prisonerId, prisonerNumber, offenderId, pnc etc etc etc
🤷‍♂️

PR #734, opened 2024-07-24
RR-764 - Send audit event when creating goals

nathanrussell-moj-digital:
This PR sends an audit event to HMPPS Audit when the user creates goal(s)

The minor difference between this and the previous PRs that send audit events is that the user journey allows the user to create one or more goals at the same time. Therefore the controller code needs to iterate through each new goal and send an event for each one.
The other difference is that in all previous "goal" related audit events, the context is that of an existing goal (eg: archive, unarchive or update and existing goal). In these cases we have the unique goal reference number so can include it in the `details` object of the event message data.
When creating new goals we do not have a reference number (it is generated by the API as part of creating the goal, and is not returned by the API). Therefore we cannot include it in the data.
To make the create goal event data reasonably useful we include the goal index and goal count as part of the `details` object:
```
{
        what: 'CREATE_PRISONER_GOAL',
        correlationId: '49380145-d73d-4ad2-8460-f26b039249cc',
        details: {
          goalNumber: 1,
          ofGoalsCreatedInThisRequest: 2,
        },
        subjectId: 'A1234BC',
        subjectType: 'PRISONER_ID',
        who: 'a-dps-user',
}
```

### DIFF
--- a/server/routes/createGoal/createGoalsController.ts
+++ b/server/routes/createGoal/createGoalsController.ts
@@ -86,7 +86,8 @@ export default class CreateGoalsController {
       return next(createError(500, `Error creating goal(s) for prisoner ${prisonNumber}. Error: ${e}`))
     }
 
-    await Promise.all(
+    Promise.all(
+      // no need to wait for responses
       createGoalDtos.map((createGoalDto, idx) =>
         this.auditService.logCreateGoal(createGoalAuditData(req, idx + 1, createGoalDtos.length)),
       ),
--- a/server/routes/createGoal/createGoalsController.ts
+++ b/server/routes/createGoal/createGoalsController.ts
@@ -1,5 +1,5 @@
 import createError from 'http-errors'
-import type { RequestHandler } from 'express'
+import type { Request, RequestHandler } from 'express'
 import type { CreateGoalsForm } from 'forms'
 import logger from '../../../logger'
 import CreateGoalsView from './createGoalsView'
@@ -7,9 +7,14 @@ import validateCreateGoalsForm from './createGoalsFormValidator'
 import toCreateGoalDtos from '../../data/mappers/createGoalDtoMapper'
 import EducationAndWorkPlanService from '../../services/educationAndWorkPlanService'
 import GoalTargetCompletionDateOption from '../../enums/goalTargetCompletionDateOption'
+import { AuditService } from '../../services'
+import { BaseAuditData } from '../../services/auditService'
 
 export default class CreateGoalsController {
-  constructor(private readonly educationAndWorkPlanService: EducationAndWorkPlanService) {}
+  constructor(
+    private readonly educationAndWorkPlanService: EducationAndWorkPlanService,
+    private readonly auditService: AuditService,
+  ) {}
 
   getCreateGoalsView: RequestHandler = async (req, res, next): Promise<void> => {
     const { prisonerSummary } = req.session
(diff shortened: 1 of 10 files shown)


server/routes/createGoal/createGoalsController.ts
```diff
@@ -71,16 +76,22 @@ export default class CreateGoalsController {
...
       logger.error(`Error creating goal(s) for prisoner ${prisonNumber}`, e)
       return next(createError(500, `Error creating goal(s) for prisoner ${prisonNumber}. Error: ${e}`))
     }
+
+    await Promise.all(
+      createGoalDtos.map((createGoalDto, idx) =>
+        this.auditService.logCreateGoal(createGoalAuditData(req, idx + 1, createGoalDtos.length)),
+      ),
+    )
+    return res.redirectWithSuccess(`/plan/${prisonNumber}/view/overview`, 'Goals added')
```
james-baker-digital-justice, line 94: I probably should've asked this on the other PRs so sorry for bringing it up now.
In the case that the audit promise fails we'd show an error to the user but we would've actually created the goals. Maybe it should just log and not return an error?

nathanrussell-moj-digital: That is a good point, and you are right in that it affects the recent PRs in this story, but also the existing calls to the audit service (all page view attempts and all page views)
Leave it with me - I'll add something to address this 👍

nathanrussell-moj-digital: Sorted 👍 
As it happens `HmppsAuditClient` already has us covered for this, with it's optional `throwOnError` parameter on the `sendMessage` method (and all of this is already unit tested 👍 )
```
  async sendMessage(event: AuditEvent, throwOnError: boolean = true) {
    if (!this.enabled) return null

    const sqsMessage: SqsMessage = {
      ...event,
      details: JSON.stringify(event.details),
      service: this.serviceName,
      when: new Date().toISOString(),
    }

    try {
      const messageResponse = await this.sqsClient.send(
        new SendMessageCommand({ MessageBody: JSON.stringify(sqsMessage), QueueUrl: this.queueUrl }),
      )

      logger.info(`HMPPS Audit SQS message sent (${messageResponse.MessageId})`)

      return messageResponse
    } catch (error) {
      logger.error('Error sending HMPPS Audit SQS message, ', error)
      if (throwOnError) throw error
    }
    return null
  }
``` 

So we just need to make sure we pass `throwOnError` as `false` in our various methods in `AuditService` 👍 

I also took the opportunity to remove the `await` keywords to all the places we call the methods in `AuditService` because it's async fire'n'forget - the controllers don't need to wait for the response 👍

server/routes/createGoal/createGoalsController.ts
```diff
@@ -71,16 +76,23 @@ export default class CreateGoalsController {
...
       await this.educationAndWorkPlanService.createGoals(prisonNumber, createGoalDtos, req.user.token)
-
-      req.session.createGoalsForm = undefined
-      return res.redirectWithSuccess(`/plan/${prisonNumber}/view/overview`, 'Goals added')
     } catch (e) {
       logger.error(`Error creating goal(s) for prisoner ${prisonNumber}`, e)
       return next(createError(500, `Error creating goal(s) for prisoner ${prisonNumber}. Error: ${e}`))
     }
+
+    Promise.all(
```
james-baker-digital-justice, line 89: Does it still need to be wrapped in Promise.all?

nathanrussell-moj-digital: Good shout, probably not .... let me try ... 😁

james-baker-digital-justice (approved):
LGTM

PR #735, opened 2024-07-24
RR-882: Change step status to radio buttons

james-baker-moj:
Step statuses can be updated on the update a goal page. Currently this is done via a drop down. To improve this interaction and make it easier for users to update step statuses, we would like to change this from a drop down to radio buttons.

Screenshot 2024-07-24 at 13 28 29

### DIFF
--- a/assets/scss/components/_update-goal.scss
+++ b/assets/scss/components/_update-goal.scss
@@ -1,0 +1,7 @@
+.step-status-radios .govuk-label{
+  max-width: 100%;
+}
+
+.step-status-radios {
+  @include govuk-responsive-margin(1, "bottom");
+}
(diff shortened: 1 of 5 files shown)


nathanrussell-moj-digital (approved):
LGTM 👍

assets/scss/components/_update-goal.scss
```diff
@@ -0,0 +1,7 @@
+.step-status-radios .govuk-label{
+  max-width: 100%;
+}
```
james-baker-digital-justice, line 3: the small radios + inline radios GDS classes together caused some wrapping with "Not started" which this fixes. I tried it at a number of screen sizes and it seems to wrap correctly still.

PR #737, opened 2024-07-24
RR-894 - Remove session support for objects we no longer use/need

nathanrussell-moj-digital:
This PR helps us when it comes to us thinking about what we store on the session and how we store/key it, by removing some session objects that are no longer in use. Specifically it removes from the session:

* `PrisonerSupportNeeds`
This is an object we use, but we don't hold it on or reference it from the session
* `NewGoal` 
Both in singular and array form. This was the form object that we used with our old multi-page flow for adding goals. Now we have a single page flow process we also have a different form object type

It also removes the form definitions that were there to support the original create goal process (`NewGoal`, `CreateGoalForm`, `AddStepForm`, `AddNoteForm` and `ReviewGoalForm`). TBH these should have been removed when we switched over to the new single page create goal process 🙄 🤷‍♂️


server/@types/express/index.d.ts
```diff
@@ -28,14 +27,12 @@ declare module 'express-session' {
   interface SessionData {
     returnTo: string
     nowInMinutes: number
+    prisonerListSortOptions: string
```
nathanrussell-moj-digital, line 30: Moved this (`prisonerListSortOptions`) up in the order, away from the other objects we'll look at in due course.
`prisonerListSortOptions` is NOT prisoner centric and is used to retain the sort options on the prisoner list screen (so that when you come back to prisoner list screen after working on a given prisoner you keep the original sort order etc)

(It's the kind of thing that perhaps could/should have been implemented with a cookie, rather than holding it in the session, but I think cookies are perhaps frowned upon here and we would have to declare it in a cookie usage type page 🤷‍♂️ )

james-baker-digital-justice: Yes something to think about. Cookies are used elsewhere in moj so perhaps it's not the worst idea. Certainly if it was a modern webapp it'd be something stored client side unless it was going to persist across multiple user-sessions.

PR #740, opened 2024-07-25
RR-889: Update copy on add a qualification page

james-baker-digital-justice:
Change "Subject" to "Subject and type" and add a hint "For example, English GCSE".

We want to encourage users to capture the type of qualification as well as the subject which should in theory, make it easier to verify qualification levels.  Adding additional copy to the [subject] field and hint text should give greater clarity.


nathanrussell-moj-digital (approved):
LGTM 👍

PR #742, opened 2024-07-29
Spike into a prisoner context in the session

james-baker-digital-justice:
Avoid issues for users using multiple tabs with different users where session corruption can occur leading to errors and unexpected behaviour when editing & archiving goals.

I tried creating a custom type and using a Map but because of the session serialisation the type info is lost and you don't have access to the associated functions so I opted for a simple helper function instead.

### DIFF
--- a/server/routes/routerRequestHandlers/checkUpdateGoalFormExistsInSession.test.ts
+++ b/server/routes/routerRequestHandlers/checkUpdateGoalFormExistsInSession.test.ts
@@ -2,6 +2,7 @@ import { NextFunction, Request, Response } from 'express'
 import type { UpdateGoalForm } from 'forms'
 import { SessionData } from 'express-session'
 import checkUpdateGoalFormExistsInSession from './checkUpdateGoalFormExistsInSession'
+import getPrisonerContext from '../../data/session/prisonerContexts'
 
 describe('checkUpdateGoalFormExistsInSession', () => {
   const req = {
@@ -30,8 +31,9 @@ describe('checkUpdateGoalFormExistsInSession', () => {
   it(`should invoke next handler given update goal form exists in session`, async () => {
     // Given
     const reference = '1a2eae63-8102-4155-97cb-43d8fb739caf'
-
-    req.session.updateGoalForm = {
+    const prisonNumber = 'A1234BC'
+    req.params.prisonNumber = prisonNumber
+    getPrisonerContext(req.session, prisonNumber).updateGoalForm = {
       reference,
     } as UpdateGoalForm
 
@@ -52,7 +54,7 @@ describe('checkUpdateGoalFormExistsInSession', () => {
     const prisonNumber = 'A1234BC'
     req.params.prisonNumber = prisonNumber
 
-    req.session.updateGoalForm = undefined
+    getPrisonerContext(req.session, prisonNumber).updateGoalForm = undefined
 
     // When
     await checkUpdateGoalFormExistsInSession(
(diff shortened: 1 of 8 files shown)


server/middleware/setUpWebSession.ts
```diff
@@ -40,5 +40,13 @@ export default function setUpWebSession(): Router {
     next()
   })
 
+  // Ensure there is always a prisoner context holder
+  router.use((req, res, next) => {
+    if (!req.session.prisonerContexts) {
```
james-baker-digital-justice, line 45: this is to avoid having to perform this check every time we access the context. It could also be moved to the helper function instead.

server/routes/routerRequestHandlers/checkUpdateGoalFormExistsInSession.test.ts
```diff
@@ -20,7 +21,7 @@ describe('checkUpdateGoalFormExistsInSession', () => {
   beforeEach(() => {
     jest.resetAllMocks()
     req.user = {} as Express.User
-    req.session = {} as SessionData
+    req.session = { prisonerContexts: {} } as SessionData
```
james-baker-digital-justice, line 24: this is one annoyance with initialising it in middleware

server/utils/prisonerContexts.ts
```diff
@@ -0,0 +1,12 @@
+import type { PrisonerContext } from 'session'
+
+export default function getOrCreatePrisonerContext(
+  contexts: Record<string, PrisonerContext>,
+  prisonNumber: string,
+): PrisonerContext {
```
james-baker-digital-justice, line 6: It would be good to add this function directly into the context/session so that intellij suggests it rather than having to know it exists but I'm not sure of the best approach

james-baker-digital-justice (commented):
An alternative to this approach would be to create a "SessionService" which handle this logic but ideally we would just do it with type safety

nathanrussell-moj-digital (commented):
I like this; there is much to like here 👍 
I do have some questions though (questions of my understanding, rather than requests for change) ...

In `express/index.d.ts` you import `PrisonerContext` on line 13, but it looks like it's being imported from the `inductionForms` module? Is that right? How does that work, as it's defined in a new module `session` ?

in `session/index.d.ts` you define a new module `session`. I don't really know much about typescript modules, but how is it different (in intent/purpose) to the module `express-session`? Could the new `PrisonerContext` interface be defined in the `express-session` module, as that is where it is used? Would that work? (like I said, I don't really know much about ts modules - what you use them for and how they are supposed to work. 🤷‍♂️ )

You've defined the prisoner context helper function in `utils`. What are your thoughts on whether "utils" is an anti pattern or not?
Personally I go hot and cold on this one. I think `utils` can quickly turn into a dumping ground of unrelated stuff, and certainly if you'd have added these functions to the existing `utils/utils.ts` file I think we'd be in danger of that. FWIW I have added some util functions to this repo in the `utils` folder (much the same as you've done here), but they were copy and paste from another repo (and they were originally in `utils.ts` and I couldn't bring myself to do that 🤣 )
I guess the "utils is an anti-pattern" brigade perhaps come from a pure OO perspective where we'd normally want to model objects with behaviour, and organise our code in nice packages etc. I guess its a different mindset when it comes to TS/JS ? 🤷‍♂️ 
What are you thoughts?

nathanrussell-moj-digital (approved):
LGTM, especially the `prisonerContexts` function instead of it being in `utils` 👍 
Nice

PR #744, opened 2024-07-30
RR-860: Remove archiving feature toggle

james-baker-digital-justice:
Archiving is now live and being used so there is no need to keep the feature toggle.


nathanrussell-moj-digital (approved):
LGTM 👍

PR #745, opened 2024-07-30
RR-823: Use new goal summary card on overview

james-baker-digital-justice:
Use the new goal summary card on overview. As we no longer use the goal numbering I've also removed that logic from the mapper and moved the sorting into the template as per view archived goals.

Minor tweaks to styling and changed "last updated" to "last updated on".

I'll rename from v2 in a separate PR. As the old one is removed here it would've made the diff difficult to read.

Screenshot 2024-07-30 at 15 44 43


assets/scss/components/_summary-card.scss
```diff
@@ -54,7 +54,7 @@
     @extend .govuk-summary-card__title-wrapper;
     @include govuk-responsive-padding(0, "top");
     @include govuk-responsive-padding(0, "bottom");
-    @include govuk-responsive-margin(1, "left");
+    @include govuk-responsive-padding(4, "left");
```
james-baker-digital-justice, line 57: changed from margin to padding as there was a white border on the subheading when it had a grey background for overview page.

james-baker-digital-justice:
Still waiting for final confirmation on the ticket which is why this is "DO_NOT_MERGE"

integration_tests/e2e/overview/postInductionOverview.cy.ts
```diff
@@ -76,10 +79,55 @@ context('Prisoner Overview page - Post Induction', () => {
...
+    cy.visit(`/plan/${prisonNumber}/view/overview`)
+
+    // Then
+    const poverviewPage = Page.verifyOnPage(OverviewPage)
+    poverviewPage //
+      .isForPrisoner(prisonNumber)
+      .isPostInduction()
+      .activeTabIs('Overview')
+      .hasNumberOfGoals(3)
+      .goalSummaryCardAtPositionContains(0, aGoalThatIsDueSooner.title)
```
nathanrussell-moj-digital, line 126: Minor, but in similar methods in the Page subclasses, we normally use 1 based indexes (rather than zero based indexes), because it makes reading the tests here a little easier.
IE. here you are referring to the goal at position zero. 
Could we change the `goalSummaryCardAtPositionContains` in Overviewpage.ts` so that it is 1 indexed and is therefore consistent with similar methods?

server/data/mappers/actionPlanMapper.ts
```diff
@@ -23,15 +22,10 @@ const toGoals = (response: GetGoalsResponse, prisonNamesById: Map<string, string
  * Sets a goal sequence by creation date (oldest = 1) and returns them in creation order (newest first)
  */
 function toOrderedGoals(goals: GoalResponse[], prisonNamesById: Map<string, string>): Goal[] {
-  return [...goals]
-    .sort((left: GoalResponse, right: GoalResponse) =>
-      dateComparator(toDate(left.createdAt), toDate(right.createdAt), 'ASC'),
-    )
-    .map((goal, index) => toGoal(goal, index + 1, prisonNamesById))
-    .reverse()
+  return [...goals].map(goal => toGoal(goal, prisonNamesById))
```
nathanrussell-moj-digital, line 25: Is the function name and its corresponding doc now accurate? It's less about ordering/sequencing them, and now just a mapping operation? 
Given it's a simple mapping, could it be inlined to where `toOrderedGoals` is called (unless it is called from several places?)

nathanrussell-moj-digital (commented):
A couple of minor comments that might be worth looking at

integration_tests/e2e/overview/postInductionOverview.cy.ts
james-baker-digital-justice: I can do but it uses the cypress accessors now instead of manual numbering. So to make it 1 indexed I'd need to do a `-1` in the page object, which is fine, but feels a bit pointless. LMK if you still want me to change it and I will. 
It's already 0 indexed in the archive goals page so I'd have to update that as well to keep it consistent.

server/data/mappers/actionPlanMapper.ts
james-baker-digital-justice: Good spot, inlined.

james-baker-digital-justice:
This change is now sanctioned so removed the do not merge

integration_tests/e2e/overview/postInductionOverview.cy.ts
nathanrussell-moj-digital: I'm not sure its pointless as such; it's about readability of the tests.
But dont worry about it for now, especially as there are others that might also need changing.
Perhaps I'll create a dev tech-debt ticket to address it one day (might be good candidate ticket for any new dev as and when to introduce them to our codebase and the cypress tests 👍 )

nathanrussell-moj-digital (approved):
LGTM 👍

PR #751, opened 2024-08-21
RR-925: Add LCI to active agencies

lynnette-orr:
Added Leicester prison (LCI) to the active agencies so the PLP tile appears for it.


nathanrussell-moj-digital (approved):
LGTM 👍

PR #754, opened 2024-08-21
RR-825: Archived Goals page - no goals message

lynnette-orr:
If there are no archived goals and someone accesses the view archived goals screen, the page should be displayed with the following message in place of where the list of archived goals would otherwise be:

“Firstname Lastname has no archived goals”

### DIFF
--- a/integration_tests/e2e/goal/viewArchivedGoals.cy.ts
+++ b/integration_tests/e2e/goal/viewArchivedGoals.cy.ts
@@ -159,6 +159,6 @@ context('Unarchive a goal', () => {
 
     // Then
     archivedGoalsPage //
-      .noArchivedGoalsMessageContains('Daniel Craig has no archived goals')
+      .noArchivedGoalsMessageShouldBeVisible()
   })
 })
--- a/integration_tests/e2e/goal/viewArchivedGoals.cy.ts
+++ b/integration_tests/e2e/goal/viewArchivedGoals.cy.ts
@@ -56,6 +56,22 @@ context('Unarchive a goal', () => {
     Page.verifyOnPage(ViewArchivedGoalsPage)
   })
 
+  it('should not show the no archived goals message if there are archived goals', () => {
+    // Given
+    cy.signIn()
+    cy.visit(`/plan/${prisonNumber}/view/overview`)
+    const archivedGoal = { ...aValidGoalResponse(), status: GoalStatusValue.ARCHIVED, goalReference }
+    cy.task('getGoalsByStatus', { status: GoalStatusValue.ARCHIVED, goals: [archivedGoal] })
+    const overviewPage = Page.verifyOnPage(OverviewPage)
+
+    // When
+    const archivedGoalsPage = overviewPage.clickViewArchivedGoalsButton()
+
+    // Then
+    archivedGoalsPage //
+      .noArchivedGoalsMessageShouldNotBeVisible()
+  })
+
   it('should order goals by most recently archived', () => {
     // Given
     cy.signIn()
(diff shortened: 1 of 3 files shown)


server/views/pages/overview/viewArchivedGoals.njk
```diff
@@ -31,26 +31,30 @@
...
+                    attributes: {
+                      'data-qa': 'goal-'+ loop.index + '-unarchive-button'
+                    },
+                    'render-if': hasEditAuthority
+                  }
+                ]
+              }) }}
+            {% endfor %}
+          {% else %}
+           <p data-qa="no-archived-goals-message">{{ prisonerSummary.firstName + ' ' + prisonerSummary.lastName }} has no archived goals</p>
```
nathanrussell-moj-digital, line 56: When the page renders, does this paragraph tag display with the correct govuk styles/font?
I never know whether the styles inherit and cascade etc (eg: I dont know if classes used on parent elements such as `govuk-grid-column-two-thirds` set the font etc for `p` elements etc)
So often times I explicitly set the `govuk-body` class on the `p` tag:

```
<p class="govuk-body" data-qa=......
``` 

I don't know if we need to do that or not; like I said its mostly based on my lack of understanding about how the govuk classes and styles are setup. 
Would probably be worth checking it displays OK though 👍

integration_tests/e2e/goal/viewArchivedGoals.cy.ts
```diff
@@ -130,4 +146,19 @@ context('Unarchive a goal', () => {
...
+    cy.task('getGoalsByStatus', { status: GoalStatusValue.ARCHIVED, goals: [] })
+
+    // When
+    cy.visit(`/plan/${prisonNumber}/view/archived-goals`)
+    Page.verifyOnPage(ViewArchivedGoalsPage)
+    const archivedGoalsPage = Page.verifyOnPage(ViewArchivedGoalsPage)
+
+    // Then
+    archivedGoalsPage //
+      .noArchivedGoalsMessageContains('Daniel Craig has no archived goals')
```
nathanrussell-moj-digital, line 162: No doubt you've found some examples of where we do this type of thing, but generally we are trying not to have assertions like this that assert specific text.
The problem is that it makes the test quite fragile, because if (when! 🙄 ) the business change their mind about the wording it will break the test. EG they change it to `<prisoner name> doesn't have any goals that have been archived yet`

Some might argue that making a change to the code should break a test, and for the most part I would agree. But probably not for wording/content changes? 🤔 

I think a better approach here, and something we are trying to do elsewhere in similar tests, is to test that the element exists (pretty much the inverse of your assertion that tests it doesnt exist)
Functionally thats all we really care about - we have logic in the nunjucks template that conditionally displays the `<p>` element, so its right to test it. But I think all we need to do is test the element exists when it is supposed to, and does not exists when it is not suppposed to.

What do you think?

nathanrussell-moj-digital (commented):
Nice work 👍 
I've left a couple of comments that you might want to have a think about, but I'm really pleased that you have found your way through our codebase and found the relevant files to change, and especially that you've found and feel comfortable making changes to the cypress tests 👍 🎉 

Have a look at the comments and see what you think - I'm happy to approve it as it is, just interested in your thoughts on my comments and/or whether you decide to make the suggested changes.

server/views/pages/overview/viewArchivedGoals.njk
lynnette-orr: It did display correctly so I forgot to add the class!

integration_tests/e2e/goal/viewArchivedGoals.cy.ts
lynnette-orr: This makes sense, have amended.

nathanrussell-moj-digital (approved):
LGTM; great work 👍

PR #755, opened 2024-08-22
RR-901 - Cypress tests to prove mapping changes

nathanrussell-moj-digital:
This PR updates the cypress tests that:
1. create new qualifications as part of a new induction
2. updates an induction by adding & removing qualifications

to assert that the qualification reference number is sent in the relevant API request.

The implement for this was the last PR that introduced the mapping changes (and arguably the changes to these tests should have been in that PR 😬 ), but either way the tests in this PR prove that the qualification reference number is sent in the "update" API requests, and is null when the request involves adding new qualifications.

### DIFF
--- a/integration_tests/e2e/induction/updateEducationalQualifications.cy.ts
+++ b/integration_tests/e2e/induction/updateEducationalQualifications.cy.ts
@@ -36,10 +36,10 @@ context('Update educational qualifications within an Induction', () => {
 
     // Then
     /* Induction has highest level of education of UNDERGRADUATE_DEGREE_AT_UNIVERSITY with the following qualifications:
-         French, grade C, LEVEL_3
-         Maths, grade A, level LEVEL_3
-         Maths, grade 1st, level LEVEL_6
-         English, grade A, level LEVEL_3
+         French, grade C, LEVEL_3, 45b969cc-00c8-41c9-8a79-133d2ccf6326
+         Maths, grade A, level LEVEL_3, 99b56087-3e92-45b1-9082-c609976bbedb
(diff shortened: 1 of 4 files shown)


integration_tests/e2e/induction/updateEducationalQualifications.cy.ts
```diff
@@ -36,10 +36,10 @@ context('Update educational qualifications within an Induction', () => {
 
     // Then
     /* Induction has highest level of education of UNDERGRADUATE_DEGREE_AT_UNIVERSITY with the following qualifications:
-         French, grade C, LEVEL_3
-         Maths, grade A, level LEVEL_3
-         Maths, grade 1st, level LEVEL_6
-         English, grade A, level LEVEL_3
+         French, grade C, LEVEL_3, 45b969cc-00c8-41c9-8a79-133d2ccf6326
```
nathanrussell-moj-digital, line 39: These UUIDs are the reference numbers for the qualifications used in these tests.
They come from the prisoner Induction that is returned by the API stub that is setup in `educationAndWorkPlanApi.ts` (see the changes to it below 👇 )

integration_tests/mockApis/educationAndWorkPlanApi.ts
```diff
@@ -320,24 +320,36 @@ const stubGetInduction = (options?: {
             options.hasQualifications === true
               ? [
                   {
+                    reference: '45b969cc-00c8-41c9-8a79-133d2ccf6326',
```
nathanrussell-moj-digital, line 323: The response from the API when retrieving an Induction now includes the reference and audit (who/when) fields for each qualification

PR #757, opened 2024-08-23
RR-916: Cypress tests - OverviewPage methods to use 1 based indexes

lynnette-orr:
Changing the method goalSummaryCardAtPositionContains in OverviewPage.ts so that it is 1 indexed.


nathanrussell-moj-digital (approved):
LGTM 👍

PR #762, opened 2024-09-03
RR-936: UI-add feature toggle for qualifications

lynnette-orr:
Add feature toggle for allowing qualifications to be managed outside of an Induction.

### DIFF
--- a/helm_deploy/values-dev.yaml
+++ b/helm_deploy/values-dev.yaml
@@ -12,20 +12,21 @@ generic-service:
     shutdown: '00 21 * * 1-5' # Stop at 9.00pm UTC Monday-Friday
 
   env:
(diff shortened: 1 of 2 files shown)


nathanrussell-moj-digital (approved):
LGTM 👍 

One thing we normally do with feature toggles is to document it in the README in the section at the bottom:

```
## Feature Toggles
Features can be toggled by setting the relevant environment variable.

| Name                     | Default Value | Type    | Description                                                                                                |
|--------------------------|---------------|---------|------------------------------------------------------------------------------------------------------------|
| SOME_TOGGLE_ENABLED      | false         | Boolean | Example feature toggle, for demonstration purposes.                                                        |
```

But I probably didnt explain that very well when I asked you to do all this stuff, so lets not worry about it at the moment. We can add it to the README in a subsequent PR as part of this epic 👍

PR #763, opened 2024-09-03
RR-937: Service and client methods to call new education endpoint

lynnette-orr:
In order for the new UI screens and journey to know whether a prisoner has qualifications outside of the context of an Induction, it will need to be able to call the new GET /person/{prisonNumber}/education API endpoint.

This story is to implement the client method in EducationAndWorkPlanClient and a service method in EducationAndWorkPlanService

Subsequent stories will make use of the new service method. This story is not expected to implement any code that actually calls the service method yet.


server/testsupport/educationResponseTestDataBuilder.ts
```diff
@@ -0,0 +1,13 @@
...
+import EducationLevelValue from '../enums/educationLevelValue'
+import QualificationLevelValue from '../enums/qualificationLevelValue'
+
+const aValidEducationResponse = (): EducationResponse => ({
+  reference: 'dea24acc-fde5-4ead-a9eb-e1757de2542c',
+  educationLevel: EducationLevelValue.SECONDARY_SCHOOL_TOOK_EXAMS,
+  qualifications: [{ subject: 'Pottery', grade: 'C', level: QualificationLevelValue.LEVEL_4 }],
+  updatedAt: new Date('2023-06-19T00:00:00Z').toISOString(),
+  updatedByDisplayName: 'Alex Smith',
+})
```
nathanrussell-moj-digital, line 11: I think this might be missing a few fields 🤔 
According to our swagger spec (and hopefully the `EducationResponse` type that is generated from the swagger spec), it should look a bit more like this:
```
{
  "createdBy": "asmith_gen",
  "createdByDisplayName": "Alex Smith",
  "createdAt": "2023-06-19T09:39:44Z",
  "createdAtPrison": "BXI",
  "updatedBy": "asmith_gen",
  "updatedByDisplayName": "Alex Smith",
  "updatedAt": "2023-06-19T09:39:44Z",
  "updatedAtPrison": "BXI",
  "reference": "814ade0a-a3b2-46a3-862f-79211ba13f7b",
  "educationLevel": "PRIMARY_SCHOOL",
  "qualifications": [
    {
      "reference": "814ade0a-a3b2-46a3-862f-79211ba13f7b",
      "subject": "Maths GCSE",
      "level": "ENTRY_LEVEL",
      "grade": "Distinction",
      "createdBy": "asmith_gen",
      "createdAt": "2023-06-19T09:39:44Z",
      "updatedBy": "asmith_gen",
      "updatedAt": "2023-06-19T09:39:44Z"
    }
  ]
}
```

server/@types/express/index.d.ts
```diff
@@ -53,6 +53,7 @@ declare module 'express-session' {
     qualificationDetailsForm: QualificationDetailsForm
     additionalTrainingForm: AdditionalTrainingForm
     prisonerContexts: PrisonerContexts
+    educationDto: EducationDto
```
nathanrussell-moj-digital, line 56: Do we need this on the session? 
I'm not sure we do, do we? Because your middleware function adds it to `res.locals` (which I think is the right approach)

server/data/educationAndWorkPlanClient.ts
```diff
@@ -114,4 +117,17 @@ export default class EducationAndWorkPlanClient {
...
     })
   }
+
+  async getEducationResponse(prisonNumber: string, token: string): Promise<EducationResponse> {
+    return EducationAndWorkPlanClient.restClient(token).get<EducationResponse>({
+      path: `/person/${prisonNumber}/education`,
+    })
+  }
+
+  // Use this to test the view for education outside an induction
```
nathanrussell-moj-digital, line 127: I think there are better/different ways of doing this; but either way can we get rid of these comments/code, as we dont really want this kicking around in our main production code?
(same comment for all the similar commented out code; plus the `mockEducationData` file (that's what the test data builder file is for 😁 )

server/data/mappers/educationMapper.ts
```diff
@@ -0,0 +1,8 @@
+import type { EducationResponse } from 'educationAndWorkPlanApiClient'
+import type { EducationDto } from 'inductionDto'
+
+function toCreateEducationRequest(educationDto: EducationDto): EducationResponse {
```
nathanrussell-moj-digital, line 4: Ah ha 😁 , I got confused about this when reading the test above ☝️ 

At first I was about to say "no, we don't want to create a EducationRequest object (yet, we will do in a later story)"

But then I saw the code here and realised it was a copy and paste typo 🤣 

I think the function should be called `toEducationResponse` , and therefore change "request" to "response" in your test class above ☝️ 
👍

server/mockEducationData.ts
```diff
@@ -0,0 +1,28 @@
+// TODO delete this file once the education view has been tested
+
+import type { EducationResponse } from 'educationAndWorkPlanApiClient'
+import EducationLevelValue from './enums/educationLevelValue'
+import QualificationLevelValue from './enums/qualificationLevelValue'
+
+const today = new Date()
+
+const mockEducationData: EducationResponse = {
```
nathanrussell-moj-digital, line 9: As mentioned above, lets get rid of this, as there are better ways to do it
(essentially it's what `server/testsupport/educationResponseTestDataBuilder.ts` should be doing 👍 , so if you wanted to use test/mock data, you would call `aValidEducationResponse()` rather than  `mockEducationData`)

server/routes/routerRequestHandlers/retrieveEducation.test.ts
```diff
@@ -0,0 +1,105 @@
...
+      params: { prisonNumber },
+    } as unknown as Request
+    res = {
+      locals: {},
+    } as unknown as Response
+  })
+
+  it('should retrieve Education and store on res.locals', async () => {
+    // Given
+    const educationDto = aValidEducationResponse()
```
nathanrussell-moj-digital, line 37: Almost right 😁  .... the service method returns the DTO, not the `EducationResponse`, so you'll need to change this line to something like 
```
const educationDto = aValidEducationDto()
```
and create a suitable test data builder for it 👍

server/routes/routerRequestHandlers/retrieveEducation.ts
```diff
@@ -0,0 +1,69 @@
...
+    } catch (error) {
+      res.locals.education = { ...gracefullyHandleException(error, prisonNumber), educationDto: undefined }
+    }
+
+    next()
+  })
+}
+
+// Use this to test the view for education outside an induction
+// TODO delete once tested
```
nathanrussell-moj-digital, line 29: As above, nuke this 😁

server/services/educationAndWorkPlanService.test.ts
```diff
@@ -290,4 +292,32 @@ describe('educationAndWorkPlanService', () => {
       expect(actual).toEqual(Error('Service Unavailable'))
     })
   })
+
+  describe('getEducation', () => {
+    const prisonNumber = 'A1234BC'
+    const userToken = 'a-user-token'
+
+    it('should get prisoner education', async () => {
+      // Given
```
nathanrussell-moj-digital, line 301: Does this test need to setup some mocks and expectations? 🤔 

Specifically I think you'll want to setup the mock `educationAndWorkPlanClient` to return an `EducationResponse` (from your test data builder `aValidEducationResponse` 👍 )

The implementation then takes that `EducationResponse` and calls the mapper with it, in order to return the `EducationDto` , so you'll want to setup an expected `EducationDto` (with a test data builder that you've yet to write 👍 😁 ), and then in the `// Then` block assert that thats what you get back

server/services/educationAndWorkPlanService.ts
```diff
@@ -73,4 +76,28 @@ export default class EducationAndWorkPlanService {
       return new Map()
     }
   }
+
+  async getEducation(prisonNumber: string, token: string): Promise<EducationResponse> {
+    try {
+      const educationResponse = await this.educationAndWorkPlanClient.getEducationResponse(prisonNumber, token)
+      return toCreateEducationRequest(educationResponse)
+    } catch (error) {
```
nathanrussell-moj-digital, line 84: I'm not sure about this catch block ... I'll need to give it some thought 🤔 ....

I know it's a pattern we have in other similar places, but I'm not sure how well it plays with the catch block and `gracefullyHandleException` function in the middleware you've written 🤔  (ie. if the service method catches and handles the exception, then the exception can never been seen/handled by the middleware ....

I'll need to give this a little thought .... it might be indicative of problems we have elsewhere .... 🤔

server/@types/dto/index.d.ts
```diff
@@ -208,4 +208,9 @@ declare module 'inductionDto' {
     workTypeOther?: string
     role?: string
   }
+
+  export interface EducationDto extends ReferencedAndAuditable {
```
nathanrussell-moj-digital, line 212: I'm in two minds about whether this should be a "DTO" or a "view model"
(in other places for similar things we've used the concept of a view model, but I guess you could argue a "view model" is a DTO, and we'll be wanting a DTO much like this when we get into the Create Education stories ... 🤔 )

nathanrussell-moj-digital: Let's leave it as a DTO for now .... though it might eat me alive and we might come back to it! 🤣

server/@types/express/index.d.ts
lynnette-orr: Removed.

server/data/educationAndWorkPlanClient.ts
lynnette-orr: Yes this was a bad idea! The intention was in case it needed to be demoed locally first, then delete before merging.

server/data/mappers/educationMapper.ts
lynnette-orr: Refactored.

server/mockEducationData.ts
lynnette-orr: Yes this was another bad idea! Deleted.

server/services/educationAndWorkPlanService.ts
lynnette-orr: Have left as is for now.

server/testsupport/aValidEducationDtoTestDataBuilder.ts
```diff
@@ -0,0 +1,30 @@
+import type { EducationResponse } from 'educationAndWorkPlanApiClient'
+import EducationLevelValue from '../enums/educationLevelValue'
+import QualificationLevelValue from '../enums/qualificationLevelValue'
+
+const aValidEducationDto = (): EducationResponse => ({
+  createdBy: 'asmith_gen',
+  createdByDisplayName: 'Alex Smith',
+  createdAt: '2023-06-19T09:39:44Z',
```
nathanrussell-moj-digital, line 8: If we can change this function to return `EducationDto` you will probably find all these date properties (`createdAt` and `updatedAt`) will need to change to set a real date, rather than a string ... which in turn might mean your mapper will need some love .... 🤔  (shouldnt be too hard though)

lynnette-orr: Refactored.

nathanrussell-moj-digital (approved):
LGTM ; thanks for making the suggested changes 👍

PR #764, opened 2024-09-04
RR-938 - Added field length validation for Induction forms

nathanrussell-moj-digital:
This PR fixes a production bug in that currently our UI does dot validate field lengths for the various "other" fields in the Induction journey 😬 
See [the jira ticket](https://dsdmoj.atlassian.net/browse/RR-938) for the analysis of what fields were missing validation, but this PR addresses  the validation as agreed with Aliki

This is a big PR (sorry! 😬 ), but much of it is the same type of change in several files

### DIFF
--- a/integration_tests/e2e/induction/createInduction.cy.ts
+++ b/integration_tests/e2e/induction/createInduction.cy.ts
@@ -311,7 +311,7 @@ context('Create an Induction', () => {
               "@.personalSkillsAndInterests.interests[1].interestType == 'DIGITAL' && " +
               '@.workOnRelease.affectAbilityToWork.size() == 1 && ' +
               "@.workOnRelease.affectAbilityToWork[0] == 'NONE' && " +
-              "@.workOnRelease.affectAbilityToWorkOther == '' && " +
+              '!@.workOnRelease.affectAbilityToWorkOther && ' +
               '@.inPrisonInterests.inPrisonWorkInterests.size() == 2 && ' +
               "@.inPrisonInterests.inPrisonWorkInterests[0].workType == 'KITCHENS_AND_COOKING' && " +
               "@.inPrisonInterests.inPrisonWorkInterests[1].workType == 'PRISON_LIBRARY' && " +
@@ -448,7 +448,7 @@ context('Create an Induction', () => {
               "@.personalSkillsAndInterests.interests[0].interestType == 'COMMUNITY' && " +
               '@.workOnRelease.affectAbilityToWork.size() == 1 && ' +
               "@.workOnRelease.affectAbilityToWork[0] == 'NONE' && " +
-              "@.workOnRelease.affectAbilityToWorkOther == '' && " +
+              '!@.workOnRelease.affectAbilityToWorkOther && ' +
               '@.inPrisonInterests.inPrisonWorkInterests.size() == 1 && ' +
               "@.inPrisonInterests.inPrisonWorkInterests[0].workType == 'KITCHENS_AND_COOKING' && " +
               '@.inPrisonInterests.inPrisonTrainingInterests.size() == 1 && ' +
--- a/server/@types/express/index.d.ts
+++ b/server/@types/express/index.d.ts
@@ -15,6 +15,7 @@ import type {
   SkillsForm,
   WantToAddQualificationsForm,
   WorkedBeforeForm,
+  WorkInterestRolesForm,
   WorkInterestTypesForm,
 } from 'inductionForms'
 import type { InductionDto } from 'inductionDto'
@@ -31,7 +32,6 @@ declare module 'express-session' {
 
     prisonerSummary: PrisonerSummary
     createGoalsForm: CreateGoalsForm
-    updateGoalForm: UpdateGoalForm
     archiveGoalForm: ArchiveGoalForm
     pageFlowHistory: PageFlow
     pageFlowQueue: PageFlow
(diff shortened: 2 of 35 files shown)


integration_tests/e2e/induction/createInduction.cy.ts
```diff
@@ -311,7 +311,7 @@ context('Create an Induction', () => {
               "@.personalSkillsAndInterests.interests[1].interestType == 'DIGITAL' && " +
               '@.workOnRelease.affectAbilityToWork.size() == 1 && ' +
               "@.workOnRelease.affectAbilityToWork[0] == 'NONE' && " +
-              "@.workOnRelease.affectAbilityToWorkOther == '' && " +
+              '!@.workOnRelease.affectAbilityToWorkOther && ' +
```
nathanrussell-moj-digital, line 314: The "other" text fields are now only mapped if the corresponding answer is/includes "OTHER", hence this change

server/@types/express/index.d.ts
```diff
@@ -46,6 +46,7 @@ declare module 'express-session' {
     previousWorkExperienceTypesForm: PreviousWorkExperienceTypesForm
     previousWorkExperienceDetailForm: PreviousWorkExperienceDetailForm
     affectAbilityToWorkForm: AffectAbilityToWorkForm
+    workInterestRolesForm: WorkInterestRolesForm
```
nathanrussell-moj-digital, line 49: We need to add the `WorkInterestRolesForm` to the session - it was the only form that was not on the session previously because up until now it didnt have any validation at all 😬 so there was no need to hold it on the session.

This PR introduces validation for this form (below), hence we now need to be able to hold it on the session

server/@types/forms/index.d.ts
```diff
@@ -119,7 +119,7 @@ declare module 'inductionForms' {
   }
 
   export interface WorkInterestRolesForm {
-    workInterestRoles: Map<WorkInterestTypeValue, string>
+    workInterestRoles: [WorkInterestTypeValue, string][]
```
nathanrussell-moj-digital, line 122: Now that we hold the `WorkInterestRolesForm` on the session, we can no longer use a Map as its internal data structure, because a Map is not serializable so it does store/retrieve correctly from the session! 😢 
Changing it to an array of typed array.
(conceptually it's very similar to what we had previously with the Map, but it serializes properly)

server/routes/induction/common/additionalTrainingController.ts
```diff
@@ -43,7 +44,9 @@ export default abstract class AdditionalTrainingController extends InductionCont
       previousTraining: {
         ...inductionDto.previousTraining,
         trainingTypes: additionalTrainingForm.additionalTraining,
-        trainingTypeOther: additionalTrainingForm.additionalTrainingOther,
+        trainingTypeOther: additionalTrainingForm.additionalTraining.includes(AdditionalTrainingValue.OTHER)
```
nathanrussell-moj-digital, line 47: Only map the "other" text field if the corresponding checkbox values includes "OTHER"
We did this in most places in the induction journey, but there are a few places like this and the ones that follow where we hadn't

server/routes/induction/common/workInterestRolesController.ts
```diff
@@ -32,13 +33,15 @@ export default abstract class WorkInterestRolesController extends InductionContr
...
-        workTypeOther: interest.workTypeOther,
-        role: workInterestRolesForm.workInterestRoles?.get(interest.workType),
-      }
-    })
+    const updatedWorkInterests: Array<FutureWorkInterestDto> = inductionDto.futureWorkInterests.interests.map(
+      interest => {
+        return {
+          workType: interest.workType,
+          workTypeOther: interest.workType === WorkInterestTypeValue.OTHER ? interest.workTypeOther : undefined,
+          role: workInterestRolesForm.workInterestRoles?.find(keyValuePair => keyValuePair[0] === interest.workType)[1],
```
nathanrussell-moj-digital, line 41: Get the element from the array rather than the Map structure

server/routes/induction/create/workInterestRolesCreateController.test.ts
```diff
@@ -46,11 +47,11 @@ describe('workInterestRolesCreateController', () => {
       req.session.inductionDto = inductionDto
 
       const expectedWorkInterestRolesForm = {
-        workInterestRoles: new Map<WorkInterestTypeValue, string>([
+        workInterestRoles: [
```
nathanrussell-moj-digital, line 50: Its no longer a Map

server/routes/validators/induction/additionalTrainingFormValidator.ts
```diff
@@ -58,10 +60,14 @@ const validateAdditionalTrainingOther = (
 
   const { additionalTraining, additionalTrainingOther } = additionalTrainingForm
 
-  if (additionalTraining && additionalTraining.includes(AdditionalTrainingValue.OTHER) && !additionalTrainingOther) {
-    errors.push(
-      `Enter the type of training or vocational qualification ${prisonerSummary.firstName} ${prisonerSummary.lastName} has`,
-    )
+  if (additionalTraining && additionalTraining.includes(AdditionalTrainingValue.OTHER)) {
```
nathanrussell-moj-digital, line 63: This is where the bulk of the functional change is. We are updating the existing Induction form validators so that they not only validate for the presence of fields, but now also validate the length of the "other" text fields

server/routes/validators/induction/workInterestRolesFormValidator.ts
```diff
@@ -0,0 +1,52 @@
+import type { WorkInterestRolesForm } from 'inductionForms'
```
nathanrussell-moj-digital, line 1: New validator for `WorkInterestRolesForm` - this didnt exist before.
Now that we are validating it, this is why it has to live on the session (see above)

lynnette-orr (approved):

PR #787, opened 2024-09-11
RR-914: Change from user to system token for education

lynnette-orr:


server/services/educationAndWorkPlanService.test.ts
```diff
@@ -21,49 +21,57 @@ import toEducationDto from '../data/mappers/educationMapper'
...
     it('should create Goals', async () => {
       // Given
       const prisonNumber = 'A1234BC'
-      const userToken = 'a-user-token'
       const createGoalDto = aValidCreateGoalDtoWithOneStep()
       const createGoalsRequest = aValidCreateGoalsRequestWithOneGoal()
 
       // When
-      await educationAndWorkPlanService.createGoals(prisonNumber, [createGoalDto], userToken)
+      await educationAndWorkPlanService.createGoals(prisonNumber, [createGoalDto], systemToken)
```
nathanrussell-moj-digital, line 59: No, this isn't quite right.
It's only the `getEducation` method of `educationAndWorkPlanService` that you're changing to use the system token.
All other service methods in `educationAndWorkPlanService` are still using the user token, so you dont want to make these changes.

server/services/educationAndWorkPlanService.ts
```diff
@@ -74,9 +76,10 @@ export default class EducationAndWorkPlanService {
     return this.educationAndWorkPlanClient.unarchiveGoal(unarchiveGoalDto.prisonNumber, unarchiveGoalRequest, token)
   }
 
-  async getEducation(prisonNumber: string, token: string): Promise<EducationDto> {
+  async getEducation(prisonNumber: string, username: string): Promise<EducationDto> {
```
nathanrussell-moj-digital, line 79: This is the right change to make 👍 

But you'll also need to change your middleware `retrieveEducation` because that is the function that calls this function, so you need to change it to pass the username instead of the user token.

server/services/educationAndWorkPlanService.test.ts
lynnette-orr: Fixed.

server/routes/routerRequestHandlers/retrieveEducation.ts
```diff
@@ -9,11 +9,12 @@ import asyncMiddleware from '../../middleware/asyncMiddleware'
 const retrieveEducation = (educationAndWorkPlanService: EducationAndWorkPlanService): RequestHandler => {
   return asyncMiddleware(async (req: Request, res: Response, next: NextFunction) => {
     const { prisonNumber } = req.params
+    const { username } = req.params
```
nathanrussell-moj-digital, line 12: Almost right 😁 , except the username wont come from `req.params` (`req.params` are the path parameters on the URL - eg: `https://learning-and-work-progress-dev.hmpps.service.justice.gov.uk/prisoners/G2385GG/create-induction/affect-ability-to-work`, you can see the prison number (`G2385GG`) in the path. 
We don't have the username in any of our URL paths.

The user object is on the request, and it in turn has a username property - `req.user.username`
Have a look at what you did in your last PR in `TimelineController.ts` for an example:
```
    const timeline = await this.timelineService.getTimeline(prisonNumber, req.user.username)
```

Your test above will also need to change based on this

nathanrussell-moj-digital (approved):
LTGM 👍

PR #790, opened 2024-09-12
RR-914: Change from user to system token for action plan

lynnette-orr:


server/services/educationAndWorkPlanService.test.ts
```diff
@@ -87,48 +87,45 @@ describe('educationAndWorkPlanService', () => {
...
       const prisonNumber = 'A1234BC'
-      const userToken = 'a-user-token'
       const actionPlanResponse = aValidActionPlanResponseWithOneGoal()
       educationAndWorkPlanClient.getActionPlan.mockResolvedValue(actionPlanResponse)
       prisonService.getAllPrisonNamesById.mockResolvedValue(new Map())
       const expectedActionPlan = aValidActionPlanWithOneGoal()
 
       // When
-      const actual = await educationAndWorkPlanService.getActionPlan(prisonNumber, userToken)
+      const actual = await educationAndWorkPlanService.getActionPlan(prisonNumber, systemToken)
```
nathanrussell-moj-digital, line 96: Not quite .... the service should be called with the username (ie. lines 96, 112, 125)

You assertions of how the client is called (lines 99, 115, 128) are correct 👍 , its just the above lines for the service calls that are wrong.

You could beef up these tests by adding an extra assertion, just after the ones you've changed:
```
expect(hmppsAuthClient.getSystemClientToken).toHaveBeenCalledWith(username)
```
That way you can be sure the auth client is being called with the correct expcted value ; ie. the username 👍

nathanrussell-moj-digital: You could add another assertion as well, just to be triple sure its all working as expected 😁 
You could add
```
expect(prisonService.getAllPrisonNamesById).toHaveBeenCalledWith(systemToken)
```

👍

server/services/educationAndWorkPlanService.test.ts
```diff
@@ -87,48 +88,50 @@ describe('educationAndWorkPlanService', () => {
...
       prisonService.getAllPrisonNamesById.mockResolvedValue(new Map())
       const expectedActionPlan = aValidActionPlanWithOneGoal()
 
       // When
-      const actual = await educationAndWorkPlanService.getActionPlan(prisonNumber, userToken)
+      const actual = await educationAndWorkPlanService.getActionPlan(prisonNumber, username)
 
       // Then
-      expect(educationAndWorkPlanClient.getActionPlan).toHaveBeenCalledWith(prisonNumber, userToken)
+      expect(hmppsAuthClient.getSystemClientToken).toHaveBeenCalledWith(username)
```
nathanrussell-moj-digital, line 100: 👍

nathanrussell-moj-digital (approved):
Looks really good
Great work 👍

PR #794, opened 2024-09-17
RR-953: Create middleware for retrieving goals

lynnette-orr:
Creating new middleware to retrieve goals and store them on res.locals.

### DIFF
--- a/server/routes/overview/overviewController.test.ts
+++ b/server/routes/overview/overviewController.test.ts
@@ -9,7 +9,6 @@ import { aValidEnglishInPrisonCourse, aValidMathsInPrisonCourse } from '../../te
 import aValidPrisonerSummary from '../../testsupport/prisonerSummaryTestDataBuilder'
 
 jest.mock('../../services/curiousService')
-jest.mock('../../services/educationAndWorkPlanService')
 jest.mock('../../services/inductionService')
 
 describe('overviewController', () => {
@@ -83,7 +82,7 @@ describe('overviewController', () => {
       prisonerSummary: expectedPrisonerSummary,
       tab: expectedTab,
       prisonNumber,
-      goals: res.locals.goals.goals,
+      goals: res.locals.goals,
       functionalSkills: expectedFunctionalSkills,
       inPrisonCourses,
       isPostInduction: true,
@@ -127,7 +126,7 @@ describe('overviewController', () => {
       prisonerSummary: expectedPrisonerSummary,
       tab: expectedTab,
       prisonNumber,
-      goals: res.locals.goals.goals,
+      goals: res.locals.goals,
       functionalSkills: expectedFunctionalSkills,
       inPrisonCourses,
       isPostInduction: false,
--- a/server/routes/overview/overviewController.test.ts
+++ b/server/routes/overview/overviewController.test.ts
@@ -1,14 +1,12 @@
 import createError from 'http-errors'
 import { Request, Response } from 'express'
-import type { FunctionalSkills, Goals, InPrisonCourseRecords } from 'viewModels'
+import type { FunctionalSkills, InPrisonCourseRecords } from 'viewModels'
 import OverviewController from './overviewController'
 import CuriousService from '../../services/curiousService'
-import EducationAndWorkPlanService from '../../services/educationAndWorkPlanService'
 import InductionService from '../../services/inductionService'
 import { aValidGoal } from '../../testsupport/actionPlanTestDataBuilder'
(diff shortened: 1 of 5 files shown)


server/routes/overview/overviewController.test.ts
```diff
@@ -92,7 +83,7 @@ describe('overviewController', () => {
       prisonerSummary: expectedPrisonerSummary,
       tab: expectedTab,
       prisonNumber,
-      goals,
+      goals: res.locals.goals.goals,
```
nathanrussell-moj-digital, line 86: I think this line, and 130, should be:
```
goals: res.locals.goals,
```
IE. we want the whole object inc. the problemRetrievingData field, not just the array of goals.

server/routes/overview/overviewController.ts
```diff
@@ -24,11 +21,7 @@ export default class OverviewController {
       const allFunctionalSkills = await this.curiousService.getPrisonerFunctionalSkills(prisonNumber, req.user.username)
       const functionalSkills = mostRecentFunctionalSkills(allFunctionalSkills)
 
-      const goals = await this.educationAndWorkPlanService.getGoalsByStatus(
-        prisonNumber,
-        GoalStatusValue.ACTIVE,
-        req.user.username,
-      )
+      const { goals } = res.locals.goals || { goals: [], problemRetrievingData: true }
```
nathanrussell-moj-digital, line 24: I think this should just be
```
const { goals } = res.locals
```

IE. use the whole goals object from res.locals; as we shouldnt need to worry about null checks or similar as its a fair assumption that your new middleware will have run first 👍

server/routes/routerRequestHandlers/retrieveGoals.ts
```diff
@@ -0,0 +1,51 @@
+import { NextFunction, Request, RequestHandler, Response } from 'express'
+import { EducationAndWorkPlanService } from '../../services'
+import logger from '../../../logger'
+import asyncMiddleware from '../../middleware/asyncMiddleware'
+import GoalStatusValue from '../../enums/goalStatusValue'
+
+/**
+ *  Middleware function that returns a Request handler function to retrieve the goals by status from EducationAndWorkPlanService and store in res.locals
+ */
+const retrieveGoals = (educationAndWorkPlanService: EducationAndWorkPlanService): RequestHandler => {
```
nathanrussell-moj-digital, line 10: I think the `educationAndWorkPlanService` already handles whether the API client returned an error or not, so I dont think you need to worry about that here.
I think all you want here is essentially what was previously in the controller. Something like this:

```
const retrieveGoals = (educationAndWorkPlanService: EducationAndWorkPlanService): RequestHandler => {
    const status: GoalStatusValue = req.query.status as GoalStatusValue

    // Retrieve the goals by status and store in res.locals
    res.locals.goals = await educationAndWorkPlanService.getGoalsByStatus(prisonNumber, status, req.user.username)

    next()
  })
}
```

I think thats all you need 🤔

server/routes/routerRequestHandlers/retrieveGoals.test.ts
```diff
@@ -0,0 +1,121 @@
...
+    const expected = {
+      problemRetrievingData: false,
+      goals,
+    }
+
+    // When
+    await requestHandler(req, res, next)
+
+    // Then
+    expect(res.locals.goals).toEqual(expected)
```
nathanrussell-moj-digital, line 51: See my comment below on the actual implementation, but I think your expected is what we setup in the `educationAndWorkPlanService` mock; so I think this line wants to be
```
expect(res.locals.goals).toEqual(goals)
```
And you can remove the `expected` variable on line 40

server/routes/routerRequestHandlers/retrieveGoals.test.ts
```diff
@@ -0,0 +1,121 @@
...
+    expect(educationAndWorkPlanService.getGoalsByStatus).toHaveBeenCalledWith(
+      prisonNumber,
+      GoalStatusValue.ACTIVE,
+      username,
+    )
+    expect(next).toHaveBeenCalled()
+  })
+
+  it('should handle retrieval of Goals given Education service returns an unexpected error', async () => {
```
nathanrussell-moj-digital, line 60: I dont think you need these 2 sad path tests because that it all handled (and tested) in `educationAndWorkPlanService`

nathanrussell-moj-digital (approved):

nathanrussell-moj-digital (approved):

PR #802, opened 2024-09-18
RR-950 - Refactor qualification controllers; controller and route to update highest level of education

nathanrussell-moj-digital:
This PR is quite big (sorry! 😬 ) but has been necessary to enable updating qualification stuff (specifically in the PR just the Highest Level of Education)

Basically the new qualification controllers in `prePrisonEducation` have been refactored to be more like what we did with Inductions, in that there are some super classes in `common`, then we extend those classes with journey specific code for `create` and `update`
So, all the existing controllers that were previously there to support the create journey have been split out into an abstract class in `common` and a concrete implementation in `create`

Then, I've created the `update` folder, but only started with Highest Level of Education. (My thinking was to get that working first, then we can follow this pattern for the other aspects of qualifications)

Some of the existing integration tests have had to change as well


integration_tests/e2e/induction/createInduction.cy.ts
```diff
@@ -474,7 +474,7 @@ context('Create an Induction', () => {
     //  Highest Level of Education: SECONDARY_SCHOOL_TOOK_EXAMS
     //  Qualifications:
     //    Pottery, C, Level 4
-    cy.task('stubGetEducation', prisonNumberForPrisonerWithNoInduction)
+    cy.task('stubGetEducation', { prisonNumber: prisonNumberForPrisonerWithNoInduction })
```
nathanrussell-moj-digital, line 477: The `stubGetEducation` stub has changed very slightly , as it now takes an options object as it's arg, rather than simply the prison number. The options object is used to control whether there are any qualifications in the education record or not.

integration_tests/e2e/induction/updateHighestLevelOfEducation.cy.ts
```diff
@@ -19,103 +18,60 @@ context('Update highest level of education within an Induction', () => {
     cy.task('getPrisonerById')
     cy.task('stubLearnerProfile')
     cy.task('stubLearnerEducation')
-    cy.task('stubUpdateInduction')
     cy.task('stubGetInduction')
+    cy.task('stubGetEducation')
```
nathanrussell-moj-digital, line 22: Quite a few changes in this test, but basically its now all about the education API, rather than the Induction API

lynnette-orr (approved):

PR #808, opened 2024-09-24
RR-953: Goals tabs

lynnette-orr:
Adding a goals tab with subtabs for in progress and archived goals.


helm_deploy/hmpps-education-and-work-plan-ui/values.yaml
```diff
@@ -56,6 +56,7 @@ generic-service:
     AUDIT_SERVICE_NAME: "hmpps-education-and-work-plan"
     AUDIT_ENABLED: "true"
     QUALIFICATIONS_ENABLED: "false"
+    NEW_OVERVIEW_PAGE_ENABLED: "true"
```
nathanrussell-moj-digital, line 59: This needs to be `false`; otherwise it will be enabled in all envs (inc production! 😬 ) as soon as it is merged!
We want it false in this file, but true in the values-dev.yaml file, so that it is only enabled in dev 👍

PR #815, opened 2024-09-26
RR-953 - Add middleware

lynnette-orr:


server/routes/routerRequestHandlers/retrieveAllGoalsForPrisoner.ts
```diff
@@ -0,0 +1,21 @@
+import { NextFunction, Request, RequestHandler, Response } from 'express'
+import { EducationAndWorkPlanService } from '../../services'
+import asyncMiddleware from '../../middleware/asyncMiddleware'
+
+/**
+ *  Middleware function that returns a Request handler function to look up the prisoner's goals and store in res.locals
+ */
+const getAllGoalsForPrisoner = (educationAndWorkPlanService: EducationAndWorkPlanService): RequestHandler => {
```
nathanrussell-moj-digital, line 8: Yep, this is exactly right; but you will need to do a unit test for it.
Should be pretty simple; base it on the unit test of `retrieveCuriousInPrisonCourses` 😁

server/routes/overview/index.ts
```diff
@@ -57,5 +58,8 @@ export default (router: Router, services: Services) => {
     asyncMiddleware(viewArchivedGoalsController.viewArchivedGoals),
   ])
 
-  router.get('/plan/:prisonNumber/view/goals', [asyncMiddleware(viewGoalsController.viewGoals)])
+  router.get('/plan/:prisonNumber/view/goals', [
+    retrieveAllGoalsForPrisoner(services.educationAndWorkPlanService),
```
nathanrussell-moj-digital, line 62: This is correct 👍 
What we are doing here is calling the middleware as part of the route definition
The middleware runs first, and puts the prisoner goals on res.locals 👍 
And then the controller method is invoked.

Which means .... your controller can just assume that the goals are already on res.locals, so you dont need to inject the service into the controller, and you dont need to call the service to lookup the goals within the controller

server/routes/overview/index.ts
```diff
@@ -27,7 +28,7 @@ export default (router: Router, services: Services) => {
   const workAndInterestsController = new WorkAndInterestsController()
   const educationAndTrainingController = new EducationAndTrainingController(services.curiousService)
   const viewArchivedGoalsController = new ViewArchivedGoalsController(services.educationAndWorkPlanService)
-  const viewGoalsController = new ViewGoalsController()
+  const viewGoalsController = new ViewGoalsController(services.educationAndWorkPlanService)
```
nathanrussell-moj-digital, line 31: You dont need to do this; no need to inject the service into the controller (see my other comments)

server/routes/overview/viewGoalsController.ts
```diff
@@ -1,11 +1,29 @@
 import { RequestHandler } from 'express'
 import ViewGoalsView from './viewGoalsView'
+import EducationAndWorkPlanService from '../../services/educationAndWorkPlanService'
 
 export default class ViewGoalsController {
-  constructor() {}
+  constructor(private readonly educationAndWorkPlanService: EducationAndWorkPlanService) {}
```
nathanrussell-moj-digital, line 6: No need for the constructor and no need to inject the service into the controller

server/routes/overview/viewGoalsController.ts
```diff
@@ -1,11 +1,29 @@
...
-  viewGoals: RequestHandler = async (req, res, next): Promise<void> => {
-    const view = new ViewGoalsView(req.session.prisonerSummary)
-    res.render('pages/overview/partials/goalsTab/goalsTabContents', { ...view.renderArgs })
+  viewGoals: RequestHandler = async (req, res, next) => {
+    try {
+      const { prisonerSummary } = req.session
+      const { prisonNumber } = req.params
+      const { username } = req.user
+
+      const prisonerGoals = await this.educationAndWorkPlanService.getAllGoalsForPrisoner(prisonNumber, username)
```
nathanrussell-moj-digital, line 14: No need to call the service to get the prisoner goals, because you've already done it in the route definition via your new middleware.
At this point in your code you can safely assume that res.locals contains a `PrisonerGoals` instance (res.locals.allGoalsForPrisoner)

nathanrussell-moj-digital: I also dont think we'll need the try catch 👍

server/routes/overview/viewGoalsView.ts
```diff
@@ -1,14 +1,21 @@
...
 
 export default class ViewGoalsView {
-  constructor(private readonly prisonerSummary: PrisonerSummary) {}
+  constructor(
+    private readonly prisonerSummary: PrisonerSummary,
+    private readonly inProgressGoals: Array<Goal>,
+    private readonly archivedGoals: Array<Goal>,
+    private readonly completedGoals: Array<Goal>,
+    private readonly problemRetrievingData: boolean,
+  ) {}
```
nathanrussell-moj-digital, line 10: I think you can tidy your constructor here, which will make your controller code neater.
Basically your constructor only needs to take the prisonerSummary and allGoalsForPrisoner: PrisonerGoals
```
  constructor(
    private readonly prisonerSummary: PrisonerSummary,
    private readonly allGoalsForPrisoner: PrisonerGoals,
  ) {}
```

Then your controller code becomes really neat and concise:
```
      const view = new ViewGoalsView(
        prisonerSummary,
        res.locals.allGoalsForPrisoner,
      )

      res.render('pages/overview/partials/goalsTab/goalsTabContents', view.renderArgs)
```

Then, your renderArgs method here needs to define it's return type as being different to the constructor, and then break out the fields from the PrisonerGoals:
```
  get renderArgs(): {
    tab: string
    inProgressGoals: Array<Goal>
    archivedGoals: Array<Goal>
    completedGoals: Array<Goal>
  } {
    return {
      tab: 'goals',      
      prisonerSummary: this.prisonerSummary,
      problemRetrievingData: this.allGoalsForPrisoner.problemRetrievingData,
      inProgressGoals: this.allGoalsForPrisoner.inProgressGoals,
      archivedGoals: this.allGoalsForPrisoner.archivedGoals,
      completedGoals: this.allGoalsForPrisoner.completedGoals,
    }
```

PR #827, opened 2024-10-02
RR-895 - Corrected module name for prisoner search API client

nathanrussell-moj-digital:
This PR corrects the prisoner search API client module name
It was originally `prisonRegisterApiClient` (a copy and paste typo from the real Prison Register API Client !), and should be `prisonerSearchApiClient`

Whilst the app works with this incorrect module name, we need to fix this because a) it is wrong/misleading and b) we are about to make some improvements to how Prisoner Search works, so we need this module name to be correct

### DIFF
--- a/server/@types/prisonerSearchApiClient/index.d.ts
+++ b/server/@types/prisonerSearchApiClient/index.d.ts
@@ -1,4 +1,4 @@
-declare module 'prisonRegisterApiClient' {
+declare module 'prisonerSearchApiClient' {
   import { components } from '../prisonerSearchApi'
 
   export type PagedCollectionOfPrisoners = components['schemas']['PagePrisoner']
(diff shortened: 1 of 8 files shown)


server/@types/prisonerSearchApiClient/index.d.ts
```diff
@@ -1,4 +1,4 @@
-declare module 'prisonRegisterApiClient' {
```
nathanrussell-moj-digital, line 1: This is the core problem, where we have incorrectly copy/pasted the module name from Prison Register API Client

The data types within this module are to do with prisoners, not prisons; and the data comes from the HMPPS Prisoner Search API.  So the module name should be `prisonerSearchApiClient`

The rest of the changes in this PR are simply correcting the imports based on the new module name

lynnette-orr (approved):

PR #828, opened 2024-10-02
RR-895 - Fixed and improved the "page not found" cypress tests

nathanrussell-moj-digital:
This PR fixes and improves a couple of cypress tests re: page not found

These tests need to be changed/improved as per this PR because the changes we are about to introduce re: caching prisoner searches ends up breaking one of the tests (because they are actually flawed! 🙄 )

### DIFF
--- a/integration_tests/e2e/page-not-found.cy.ts
+++ b/integration_tests/e2e/page-not-found.cy.ts
@@ -1,6 +1,7 @@
 import Page from '../pages/page'
 import Error404Page from '../pages/error404'
 import AuthSignInPage from '../pages/authSignIn'
+import AuthorisationErrorPage from '../pages/authorisationError'
 
 context('404 Page Not Found', () => {
   beforeEach(() => {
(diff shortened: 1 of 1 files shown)


integration_tests/e2e/page-not-found.cy.ts
```diff
@@ -27,7 +29,7 @@ context('404 Page Not Found', () => {
     Page.verifyOnPage(AuthSignInPage)
   })
 
-  it('should redirect to 404 page given user does not have any authorities', () => {
+  it('should redirect to Authorisation Error page given user does not have any authorities', () => {
```
nathanrussell-moj-digital, line 32: Previously this test was going to the 404 page not because the user had no/incorrect roles, but because we had not setup the stub for getting the prisoner by ID !!!, so it was actually that (getting a prisoner without setting up the stub) that was throwing the 404! 🙄

lynnette-orr (approved):

PR #829, opened 2024-10-02
RR-895 - Cache prisoner searches with a redis backed `prisonerSearchStore`

nathanrussell-moj-digital:
This PR adds caching to the prisoner search service method
IE. the service method to lookup a prisoner from `prisoner-search-api` and return as a `PrisonerSummary`

This follows our existing patterns where we do similar things (prison lookups via prison-register-api for example), where the service method and it's method signature do not change, but the internal implementation of the method looks in the redis cache first rather than simply calling the API.

### DIFF
--- a/server/data/prisonerSearchStore/prisonerSearchStore.ts
+++ b/server/data/prisonerSearchStore/prisonerSearchStore.ts
@@ -1,0 +1,30 @@
+import type { Prisoner } from 'prisonerSearchApiClient'
+import { RedisClient } from '../redisClient'
+import logger from '../../../logger'
+
+const PRISONER = 'prisoner'
+
+export default class PrisonerSearchStore {
+  constructor(private readonly client: RedisClient) {
+    client.on('error', error => {
+      logger.error(error, `Redis error`)
+    })
+  }
+
+  private async ensureConnected() {
+    if (!this.client.isOpen) {
+      await this.client.connect()
+    }
+  }
+
+  async setPrisoner(prisonNumber: string, prisoner: Prisoner, durationDays = 1): Promise<string> {
+    await this.ensureConnected()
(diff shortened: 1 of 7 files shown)


server/data/prisonerSearchStore/prisonerSearchStore.ts
```diff
@@ -0,0 +1,30 @@
+import type { Prisoner } from 'prisonerSearchApiClient'
+import { RedisClient } from '../redisClient'
+import logger from '../../../logger'
+
+const PRISONER = 'prisoner'
+
+export default class PrisonerSearchStore {
```
nathanrussell-moj-digital, line 7: The concept of a "store" class simply takes a redis client instance, then exposes methods to get/retrieve objects from redis
In this case we are storing and retrieving a `Prisoner` instance in redis

server/services/prisonerSearchService.ts
```diff
@@ -1,16 +1,64 @@
...
   constructor(
     private readonly hmppsAuthClient: HmppsAuthClient,
     private readonly prisonerSearchClient: PrisonerSearchClient,
+    private readonly prisonerSearchStore: PrisonerSearchStore,
   ) {}
 
   async getPrisonerByPrisonNumber(prisonNumber: string, username: string): Promise<PrisonerSummary> {
-    const systemToken = await this.hmppsAuthClient.getSystemClientToken(username)
-    const prisoner = await this.prisonerSearchClient.getPrisonerByPrisonNumber(prisonNumber, systemToken)
+    let prisoner: Prisoner
```
nathanrussell-moj-digital, line 18: The service method to get a prisoner by prison number now looks to return it from redis first
It the prisoner is in redis (the store class) then use it
If the prisoner is not in redis then retrieve if from the API (the previous behaviour), and store it in redis

lynnette-orr (approved):

PR #839, opened 2024-10-04
RR-954-Realignment of grid items on overview page

lynnette-orr:
Adding a container around the goal counts and tags so they can be better aligned.

Screenshot 2024-10-04 at 16 04 22


nathanrussell-moj-digital (approved):
LGTM 👍

PR #845, opened 2024-10-07
RR-978: Removing neurodiversity summary card

lynnette-orr:
Removed neurodiversity summary card from Support Needs page.

Screenshot 2024-10-07 at 16 14 11


nathanrussell-moj-digital (commented):
What you've done here is fine and works 👍 
but I think we should also get rid of the code that actually gets and populates the neuro data etc.

Have a look at the view model `PrisonerSupportNeeds` in `server/@types/viewModels` 👀 
If you start there, you can remove the property `neurodiversities` and the type `Neurodiversity`
Then you can work you way forwards through the service (`CuriousService`), and the mapper (`PrisonerSupportNeedsMapper`), the controller (`SupportNeedsController`) and the test data builder (`supportNeedsTestDataBuilder`)

nathanrussell-moj-digital (approved):
Awesome work; LGTM 👍

PR #851, opened 2024-10-09
RR-980 - Support needs tab changes

lynnette-orr:
Updating the Support Needs page to show LDD data for each prison, or a message underneath the prison name if no data is present. A message will be displayed if no prisons have LDD data.

Scenario 1: No prisons have any screener data available
Screenshot 2024-10-09 at 14 16 35

Scenario 2: Some, but not all, data is available for a prison
Screenshot 2024-10-09 at 14 09 44

Scenario 3: One prison has data, the other does not have any data
Screenshot 2024-10-09 at 14 14 59


server/views/pages/overview/partials/supportNeedsTab/supportNeedsTabContents.test.ts
```diff
@@ -1,54 +1,101 @@
...
+import fallbackMessageFilter from '../../../../../filters/fallbackMessageFilter'
+import formatDateFilter from '../../../../../filters/formatDateFilter'
 
-describe('Support Needs tab view', () => {
-  const template = fs.readFileSync('server/views/pages/overview/partials/supportNeedsTab/supportNeedsTabContents.njk')
-  const prisonerSummary = aValidPrisonerSummary()
-
-  let compiledTemplate: Template
-  let viewContext: Record<string, unknown>
+const njkEnv = nunjucks.configure([
```
nathanrussell-moj-digital, line 8: 👍 Excellent refactoring of this test so that it sets up the nunjucks env in the way our more recent tests do. Nice

integration_tests/mockApis/curiousApi.ts
```diff
@@ -33,10 +33,18 @@ const stubLearnerProfile = (prisonNumber = 'G6115VJ'): SuperAgentRequest =>
           ],
           languageStatus: null,
           plannedHours: null,
-          rapidAssessmentDate: null,
-          inDepthAssessmentDate: null,
-          primaryLDDAndHealthProblem: null,
-          additionalLDDAndHealthProblems: [],
+          prisonId: 'MDI',
```
nathanrussell-moj-digital, line 36: I'm confused by this change
This stub file is the stub of the raw data we get from the Curious API .... and the Curious API hasnt changed, and it certainly does not have fields `prisonName`, `atLeastOnePrisonHasSupportNeeds`, and `hasSupportNeeds`

I can see why you've set values for the Curious API fields `rapidAssessmentDate`, `primaryLDDAndHealthProblem` and `additionalLDDAndHealthProblems`, but not the ones mentioned above 🤔 

And also, I _think_ I'd leave `inDepthAssessmentDate` as `null` rather than `undefined`. I know in respect of how we handle the fields/data downstream of this it makes no difference (null and undefined are both falsey),  but given this stub is to mirror the data we would expect to get from a given API I think it's important to use the same data types/values that we would expect to get from the real API. And I think the real API will return a field with the value `null` rather than `undefined` (or not returning the field at all)

server/routes/overview/supportNeedsController.ts
```diff
@@ -15,20 +15,37 @@ export default class SupportNeedsController {
 
     const supportNeeds = await this.curiousService.getPrisonerSupportNeeds(prisonNumber, req.user.username)
     const prisonNamesById = await this.prisonService.getAllPrisonNamesById(req.user.username)
+    let atLeastOnePrisonHasSupportNeeds = false
 
     // Loop through the healthAndSupport needs array and update the prison name for each need
     if (supportNeeds.healthAndSupportNeeds) {
-      supportNeeds.healthAndSupportNeeds.map(async supportNeed => {
+      supportNeeds.healthAndSupportNeeds = supportNeeds.healthAndSupportNeeds.map(supportNeed => {
```
nathanrussell-moj-digital, line 22: I'm not sure this change is right / I'm not quite sure what it is doing / I think it's now trying to do more than the comment on line 20 suggests 
(though I think you are right with the removal of `async`; and I think there is a minor problem with the current code through the use of `map` but it doesnt actually do anything with the return!

How about this:
```
    let atLeastOnePrisonHasSupportNeeds = false

    if (supportNeeds.healthAndSupportNeeds) {
      // Work out if any of the HealthAndSupportNeeds records contains any assessment data
      atLeastOnePrisonHasSupportNeeds = supportNeeds.healthAndSupportNeeds.some(
        supportNeed =>
          supportNeed.rapidAssessmentDate ||
          supportNeed.inDepthAssessmentDate ||
          supportNeed.primaryLddAndHealthNeeds ||
          (supportNeed.additionalLddAndHealthNeeds && supportNeed.additionalLddAndHealthNeeds.length > 0),
      )

      // Loop through the healthAndSupport needs array and update the prison name for each need
      supportNeeds.healthAndSupportNeeds = supportNeeds.healthAndSupportNeeds.map(supportNeed => {
        const prison = prisonNamesById.get(supportNeed.prisonId)
        return {
          ...supportNeed,
          prisonName: prison,
        }
      })
    }
```

(I've not tested the above, you'll need to play with it to see if it works / does what you are trying to achieve etc)

server/routes/overview/mappers/prisonerSupportNeedsMapper.ts
```diff
@@ -18,6 +18,11 @@ const toHealthAndSupportNeeds = (learnerProfile: LearnerProfile): HealthAndSuppo
       inDepthAssessmentDate: dateOrNull(learnerProfile.inDepthAssessmentDate),
       primaryLddAndHealthNeeds: learnerProfile.primaryLDDAndHealthProblem,
       additionalLddAndHealthNeeds: learnerProfile.additionalLDDAndHealthProblems?.sort() || [],
+      hasSupportNeeds:
+        !!learnerProfile.rapidAssessmentDate ||
```
nathanrussell-moj-digital, line 22: Minor, but why the double bang on rapidAssessmentDate, but not the others?

nathanrussell-moj-digital (dismissed):
LGTM 👍 
(just one minor question, more for my info rather than it needing changing)

server/routes/overview/mappers/prisonerSupportNeedsMapper.ts
lynnette-orr: Added the brackets

nathanrussell-moj-digital (approved):

PR #858, opened 2024-10-11
RR-1003: Overview page changes

lynnette-orr:
Remove padding around 'Add a new goal' in Action card
Slight increase in space above 'Courses and qualifications completed in the last 12 months'

Selected October 11 2024 12:47:09

### DIFF
--- a/assets/scss/local.scss
+++ b/assets/scss/local.scss
@@ -82,6 +82,12 @@ tbody[data-qa="completed-courses-table-body"] .govuk-table__cell:nth-child(2) {
   border-bottom: none;
 }
 
+// Custom margin on completed courses heading in education and training summary card on overview page
+
+h3[data-qa="completed-courses-heading"] {
+  margin-top: 30px !important;
+}
+
 // Custom section break widths
 
 .app-u-section-break-8 {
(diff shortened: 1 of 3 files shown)


nathanrussell-moj-digital:
I'm not sure the styling changes work on the action menu where there are more than 1 item (I hacked the html in chrome dev tools to do this, you might want to try it by adding a new item in the template)
Screenshot 2024-10-11 at 13 57 33

nathanrussell-moj-digital:
Also, I'm not sure your styling change for the courses and quals works in all cases, specifically where there isnt the H3 heading:
Screenshot 2024-10-11 at 14 00 04

Also, I'm not keen on not using the standard govuk class names / styles .... that is what they are there for after all?
Is there not a govuk-heading / govuk-padding type class we can use that gives a 30px space? (the govuk styles are properly responsive as well, so I think we should be using those as preference over and above writing styles like this in the sass)
What do you think?

lynnette-orr:
Removed the custom css. Was able to use padding instead. Modified the Actions card to add some space between multiple items.

Selected October 11 2024 15:10:59

Selected October 11 2024 15:10:16

Selected October 11 2024 15:28:35

Selected October 11 2024 15:29:15

nathanrussell-moj-digital (approved):
LGTM 👍

PR #864, opened 2024-10-14
RR-1002: New tab layout on goals page

lynnette-orr:
Selected October 14 2024 16:22:49


integration_tests/e2e/overview/goal/reviewUpdateGoal.cy.ts
```diff
@@ -44,10 +44,10 @@ context('Review updated goal', () => {
     const goalReference = '10efc562-be8f-4675-9283-9ede0c19dade'
     cy.signIn()
 
-    cy.visit(`/plan/${prisonNumber}/view/goals#in-progress`)
+    cy.visit(`/plan/${prisonNumber}/view/goals/in-progress-goals`)
     const goalsPage = Page.verifyOnPage(GoalsPage)
     const inProgressGoalsPage = goalsPage.clickInProgressGoalsTab()
-    const updateGoalPage = inProgressGoalsPage.clickUpdateButtonForFirstGoal()
+    const updateGoalPage = inProgressGoalsPage.clickUpdateButtonForGoal(goalReference)
```
lynnette-orr, line 50: The reason for these changes is the goal reference in the njk template was being set as the goal number (loop.index) instead of the actual goal reference. It now uses the goal reference (see update to goalsTabContents.njk file below).

nathanrussell-moj-digital:
Blimey! There's a lot going on in this PR. There are a few things I dont think we want to be doing in this PR (eg: changes to the existing archive goals page - remember, we still have the feature toggle so in theory could be asked to switch back to the old views; changes to the service etc); but I can see roughly what you are proposing here.

My question though is are you 100% sure we cant just do this with CSS? Using dev tools I can (on the surface of it) do what we need ; though I accept we'd need to write it as a proper selector and override in the sass somewhere, and we'd need to test the different permutations of how you might arrive on/display the page. If we can do it with CSS this has got to be the easiest option hasn't it?

Screenshot 2024-10-15 at 09 09 21

PR #866, opened 2024-10-15
RR-1002: Use css for tab panel change, change goal ref from index to ref

lynnette-orr:
Using custom css for the tab panel. Changed the goal reference in the goal summary card to use the actual reference rather than the index. Added unit tests for the goals view.

Selected October 15 2024 10:31:56


nathanrussell-moj-digital (approved):
Excellent; that's much simpler! 😁 👍 
Nice work

PR #874, opened 2024-10-17
initial commit of complete goal

stevomcallister:
Added screen flow for complete goal 🍿

### Video demo
https://github.com/user-attachments/assets/c67e5d25-d6e4-42df-a697-48db0d912d8e


server/@types/express/index.d.ts
```diff
@@ -63,6 +64,7 @@ declare module 'express-session' {
     // Goal related forms
     updateGoalForm?: UpdateGoalForm
     archiveGoalForm?: ArchiveGoalForm
+    completeGoalForm?: CompleteGoalForm
```
nathanrussell-moj-digital, line 67: 100% yes, on the prisoner context 🙌

server/@types/forms/index.d.ts
```diff
@@ -56,6 +56,18 @@ declare module 'forms' {
     title: string
   }
 
+  export interface CompleteGoalForm {
+    reference: string
+    title: string
+    note: string
+  }
+
+  export interface CompleteOrArchiveGoalForm {
```
nathanrussell-moj-digital, line 65: We _might_ be able to do this without a form (cant remember what the user journey/screen looks like), but using a form as you have done is fine 👍 . Perhaps we can look at doing it without a form submission some other time 👍

server/data/educationAndWorkPlanClient.ts
```diff
@@ -68,6 +69,13 @@ export default class EducationAndWorkPlanClient {
     })
   }
 
+  async completeGoal(prisonNumber: string, completeGoalRequest: CompleteGoalRequest, token: string): Promise<void> {
```
nathanrussell-moj-digital, line 72: Might be worth considering a unit test for this new method, but we can back fill it in a subsequent PR 👍

server/routes/completeOrArchive/completeOrArchiveGoalController.ts
```diff
@@ -0,0 +1,41 @@
...
+import EducationAndWorkPlanService from '../../services/educationAndWorkPlanService'
+
+export default class CompleteOrArchiveGoalController {
+  constructor(private readonly educationAndWorkPlanService: EducationAndWorkPlanService) {}
+
+  getCompleteOrArchiveGoalView: RequestHandler = async (req, res, next): Promise<void> => {
+    const { prisonNumber, goalReference } = req.params
+    const { prisonerSummary } = res.locals
+
+    const actionPlan = await this.educationAndWorkPlanService.getActionPlan(prisonNumber, req.user.username)
```
nathanrussell-moj-digital, line 14: What you've done here is fine, and follows the patterns/approaches in other similar controllers 👍 

I've been thinking for a while that this pattern of "get the entire action plan, then find the goal in question from the returned array" is a bit meh 🤷‍♂️ and we can perhaps do better.

From memory I think it was because originally we never had an API endpoint to get a goal by its reference - all we could do was get the entire action plan, hence the UI code that we have.
But I _think_ since then we might have implemented a suitable API endpoint to get a single goal by it's reference, so we might be able to use that 🤔 

One for another day, but perhaps we should identify which pages/controllers this might affect, then create some tickets around it

stevomcallister: yes it is a bit overkill grabbing everything when all we want is the title

server/@types/forms/index.d.ts
stevomcallister: yeah I did wonder about this - the form is purely to 'carry' the goal reference during the navigation.

server/services/educationAndWorkPlanService.ts
```diff
@@ -92,6 +94,11 @@ export default class EducationAndWorkPlanService {
     return this.educationAndWorkPlanClient.unarchiveGoal(unarchiveGoalDto.prisonNumber, unarchiveGoalRequest, token)
   }
 
+  async completeGoal(completeGoalDto: CompleteGoalDto, token: string): Promise<unknown> {
```
nathanrussell-moj-digital, line 97: As above, we could do with a unit test for this; lets back fill it in another PR 👍

nathanrussell-moj-digital (approved):
I know this has been a challenge to get here, but this is great stuff 👍 
LGTM 👍

PR #876, opened 2024-10-18
RR-993: Add 2 prisons to tile

lynnette-orr:
Add Deerbolt (DTI) and Manchester (MRI) to tile.


nathanrussell-moj-digital (approved):
LGTM 👍

PR #882, opened 2024-10-21
integration test part deux: Complete a note

stevomcallister:


integration_tests/e2e/overview/goal/completeGoal.cy.ts
```diff
@@ -0,0 +1,152 @@
...
+    cy.task('stubGetFooterComponent')
+    cy.task('stubPrisonerList')
+    cy.task('stubCiagInductionList')
+    cy.task('stubActionPlansList')
+    cy.task('getPrisonerById')
+    cy.task('stubGetInduction')
+    cy.task('getActionPlan')
+    cy.task('stubLearnerProfile')
+    cy.task('stubLearnerEducation')
+    cy.task('archiveGoal')
```
nathanrussell-moj-digital, line 29: I think you want this line here (instead of stubbing `archiveGoal`)
```
cy.task('completeGoal', { prisonNumber, goalReference })
```
And then you don't need it on lines 80 and 115

integration_tests/e2e/overview/goal/completeGoal.cy.ts
```diff
@@ -0,0 +1,152 @@
...
+
+    const completeGoalPage = Page.verifyOnPage(CompleteGoalPage)
+
+    completeGoalPage //
+      .clearNotes()
+      .clickYes()
+    // Then
+    Page.verifyOnPage(OverviewPage) //
+      .hasSuccessMessage('Goal Completed')
+      .hasNumberOfCompletedGoals(1)
```
nathanrussell-moj-digital, line 105: I would remove this part of the assertion (`hasNumberOfCompletedGoals`) because by making this assertion here in the `// Then` we are suggesting that some state has changed via the `// When` block (and we are checking that it has)
The reality is that because we use wiremock to mock our API calls, no state is ever changed. So even though the impl called the API to complete the goal, and we can assert that via the `cy.wiremockVerify` call, the actual state of data has not changed, so the rendering of the Overview page wont have changed from before the `// When` to after it.
If that makes sense?

integration_tests/e2e/overview/goal/completeGoal.cy.ts
```diff
@@ -0,0 +1,152 @@
...
+
+  it('Should be able to complete a goal successfully without entering the optional notes', () => {
+    // Given
+    cy.task('completeGoal', { prisonNumber, goalReference })
+    cy.signIn()
+    cy.visit(`/plan/${prisonNumber}/view/goals`)
+    const goalsPage = Page.verifyOnPage(GoalsPage)
+
+    // When
+    const completeOrArchiveGoalPage = goalsPage //
```
nathanrussell-moj-digital, line 86: I think our `Goals` page driver class has got a bit big / does too much / concerns itself with 3 tabs that are showing fundamentally different goals.
I found lines 87 and 88, and 122 and 123 difficult to read and reason about, because I was expecting the state of the `Goals` page to be on the In-Progress tab showing in-progress goals. But we are apparently checking the archived goals are displayed and there are 2 of them!

I know why it works (because the `Goals` page class knows about all 3 tabs, irrespective of whether they are currently rendered/visible or not), but reading the test is hard

I'd be tempted to remove lines 87 and 88, and 122 and 123 as I dont think they are pertinent to competing a goal?

And I think my PR (and one I will do shortly) will make this easier in the future (I'll do a PR that splits out the `Goals` page drive class into 3, one for each tab. Then it will be much clearer which tab we are on and what functionality we can expect 👍 )

server/views/pages/goal/complete/index.njk
```diff
@@ -30,10 +30,17 @@
                 Add a note (optional)
               </label>
             </h1>
-            <p class="govuk-hint">
-              Learning and work progress plans, including notes, are shared with prisoners and other staff
-            </p>
-            <textarea class="govuk-textarea" id="completed-note" name="note" rows="5" aria-describedby="warning-text">{{ form.note }}</textarea>
+            {{ govukTextarea({
```
nathanrussell-moj-digital, line 33: 👍 Good shout for using the GDS component rather than writing the markup yourself 👍 
but I think the page will end up with 2 "Add a note (optional)" headings (1 from the H1 above, and 1 from the label attribute of this component)

Given that your page already (correctly) has a h1 ("Are you sure you want to mark this goal as completed?"), and we dont want a page with 2 H1's anyway, I'd be tempted to remove the h1 above, and keep the heading in the component (which is being marked up as a h2 👍 )

integration_tests/e2e/overview/goal/completeGoal.cy.ts
stevomcallister: not sure - I looked at what the archive goal test did and did something similar. Although you are right in that the archive test checks before and after that the number of archived goals is 2 - so bit pointless really

server/views/pages/goal/complete/index.njk
stevomcallister: oops - removed the additional h1

PR #893, opened 2024-10-25
RR-1036: Who completed review screen

lynnette-orr:
Adding 'Who completed review' screen.

https://github.com/user-attachments/assets/d6780d09-ca49-48d7-ba8c-14b5433f0c96

### DIFF
--- a/server/@types/forms/index.d.ts
+++ b/server/@types/forms/index.d.ts
@@ -164,7 +164,6 @@ declare module 'reviewPlanForms' {
 
   export interface WhoCompletedReviewForm {
     completedBy: ReviewPlanCompletedByValue
-    completedByOther?: string
     completedByOtherFullName?: string
     completedByOtherJobRole?: string
     'reviewDate-day': string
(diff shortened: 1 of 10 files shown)


server/@types/forms/index.d.ts
```diff
@@ -165,6 +165,8 @@ declare module 'reviewPlanForms' {
   export interface WhoCompletedReviewForm {
     completedBy: ReviewPlanCompletedByValue
     completedByOther?: string
```
nathanrussell-moj-digital, line 167: Really minor! (sorry!),
but did you also mean to remove the original `completedByOther` field from this form object?

If we can get rid of it (as we have done in the DTO) that would be great, as I cant see anywhere else that is using it

nathanrussell-moj-digital (commented):
Looks really good, and the video in the description is fantastic showing all the permutations 👍  Nice 👍 

My only (really minor! - sorry!) question is about whether still need the `completedByOther` field on the main form object

PR #899, opened 2024-10-30
RR-1073 - Fix bug which prevents qualifications being added to an Induction that has no qualifications to start with

nathanrussell-moj-digital:
This PR fixes a user reported bug which prevents qualifications being added to an Induction that has no qualifications to start with.

The basic scenario is that a CIAG has created an Induction for a prisoner, and when they created the Induction that did not enter any qualifications (ie. they answered No to the question "Do you want to add educational qualifications")

The CIAG at some point later comes back to the prisoner's Induction, and on the Education & Training tab, where it correctly displays no qualifications, they click the Change button to retrospectively add the prisoners qualifications.

The bug is that instead of being asked the Qualification Level for the 1st qualification, they are instead redirected back to the Eduction & Training tab.

The problem was that the Change link originally took the user to the `qualification-level` route. This route then did a check to ensure the `EducationDto` was loaded and on the prisoner context .... and in this scenario it wasn't.

The fix I have implemented is to create a new interim route `add-qualifications` which loads the `EducationDto` into the prisoner context, then redirects to `qualification-level`

I have written a cypress test specifically for this scenario to prove the bug fix 👍

### DIFF
--- a/integration_tests/e2e/prePrisonEducaton/createPrePrisonEducation.cy.ts
+++ b/integration_tests/e2e/prePrisonEducaton/createPrePrisonEducation.cy.ts
@@ -104,7 +104,7 @@ context('Create a prisoners pre-prison education', () => {
     const overviewPage = Page.verifyOnPage(OverviewPage)
     overviewPage.selectTab('Education and training')
     const educationAndTrainingPage = Page.verifyOnPage(EducationAndTrainingPage)
-    educationAndTrainingPage.clickToAddEducationalQualifications(HighestLevelOfEducationPage)
+    educationAndTrainingPage.clickToAddEducationHistory()
 
     // When
     // First page is Highest Level of Education
--- a/server/routes/prePrisonEducation/update/index.ts
+++ b/server/routes/prePrisonEducation/update/index.ts
@@ -1,4 +1,4 @@
-import { Router } from 'express'
+import { NextFunction, Request, Response, Router } from 'express'
 import { Services } from '../../../services'
 import asyncMiddleware from '../../../middleware/asyncMiddleware'
 import { checkUserHasEditAuthority } from '../../../middleware/roleBasedAccessControl'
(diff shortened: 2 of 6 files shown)


integration_tests/e2e/prePrisonEducaton/createPrePrisonEducation.cy.ts
```diff
@@ -104,7 +104,7 @@ context('Create a prisoners pre-prison education', () => {
     const overviewPage = Page.verifyOnPage(OverviewPage)
     overviewPage.selectTab('Education and training')
     const educationAndTrainingPage = Page.verifyOnPage(EducationAndTrainingPage)
-    educationAndTrainingPage.clickToAddEducationalQualifications(HighestLevelOfEducationPage)
+    educationAndTrainingPage.clickToAddEducationHistory()
```
nathanrussell-moj-digital, line 107: I've had to rename (and slightly change the semantics) of the existing method `clickToAddEducationalQualifications` because that method name makes more sense in the way I want to use it for this bug fix.
The new name `clickToAddEducationHistory` is a better name for the use case of _this_ cypress test

server/routes/prePrisonEducation/update/index.ts
```diff
@@ -24,6 +24,15 @@ export default (router: Router, services: Services) => {
   const qualificationDetailsUpdateController = new QualificationDetailsUpdateController()
   const qualificationsListUpdateController = new QualificationsListUpdateController(educationAndWorkPlanService)
 
+  router.get('/prisoners/:prisonNumber/education/add-qualifications', [
```
nathanrussell-moj-digital, line 27: This is the new route that sets up the `EducationDto` in the prisoner context, then redirects to `qualification-level` to allow the CIAG start start entering the 1st qualification

lynnette-orr (approved):

PR #902, opened 2024-10-31
Rr-1038: Check your answers for Review

lynnette-orr:
Uploading Full screen November 01 2024 12:55:09.mp4…

### DIFF
--- a/assets/js/backLink.js
+++ b/assets/js/backLink.js
@@ -1,0 +1,11 @@
+const backLink = document.querySelector('.govuk-back-link')
+if (backLink) {
+  backLink.addEventListener(
+    'click',
+    function (e) {
+      window.history.back()
+      e.preventDefault()
+    },
+    false,
+  )
+}
--- a/assets/js/backLink.js
+++ b/assets/js/backLink.js
@@ -1,4 +1,4 @@
-const backLink = document.querySelector('.govuk-back-link')
+const backLink = document.querySelector('.js-back-link')
 if (backLink) {
   backLink.addEventListener(
     'click',
--- a/integration_tests/e2e/reviewPlan/reviewPlan.cy.ts
+++ b/integration_tests/e2e/reviewPlan/reviewPlan.cy.ts
@@ -43,7 +43,7 @@ context(`Review a prisoner's plan`, () => {
 
     // Then
     Page.verifyOnPage(WhoCompletedReviewPage) //
-      .hasBackLinkTo(`/plan/${prisonNumber}/view/overview`)
+      .hasCorrectBackLink()
   })
 
   it('should redirect to overview page given user tries to navigate directly to Review Notes screen - ie. navigate out of sequence', () => {
@@ -67,7 +67,7 @@ context(`Review a prisoner's plan`, () => {
     // When
     // First page is the Who completed the review page
     Page.verifyOnPage(WhoCompletedReviewPage) //
-      .hasBackLinkTo(`/plan/${prisonNumber}/view/overview`)
+      .hasCorrectBackLink()
       .submitPage() // submit the page without answering any questions to trigger a validation error
     Page.verifyOnPage(WhoCompletedReviewPage) //
       .hasErrorCount(2)
@@ -85,7 +85,7 @@ context(`Review a prisoner's plan`, () => {
       .submitPage()
 
     // Next page is Review Notes page
-    Page.verifyOnPage(ReviewNotePage).hasBackLinkTo(`/plan/${prisonNumber}/review`).submitPage() // submit the page without answering any questions to trigger a validation error
+    Page.verifyOnPage(ReviewNotePage).hasCorrectBackLink().submitPage() // submit the page without answering any questions to trigger a validation error
     Page.verifyOnPage(ReviewNotePage)
       .hasErrorCount(1)
       .hasFieldInError('notes')
--- a/integration_tests/e2e/reviewPlan/reviewPlan.cy.ts
+++ b/integration_tests/e2e/reviewPlan/reviewPlan.cy.ts
@@ -42,8 +42,7 @@ context(`Review a prisoner's plan`, () => {
     cy.visit(`/plan/${prisonNumber}/review`)
 
     // Then
-    Page.verifyOnPage(WhoCompletedReviewPage) //
-      .hasCorrectBackLink()
+    Page.verifyOnPage(WhoCompletedReviewPage)
   })
 
   it('should redirect to overview page given user tries to navigate directly to Review Notes screen - ie. navigate out of sequence', () => {
@@ -67,7 +66,6 @@ context(`Review a prisoner's plan`, () => {
     // When
     // First page is the Who completed the review page
     Page.verifyOnPage(WhoCompletedReviewPage) //
-      .hasCorrectBackLink()
       .submitPage() // submit the page without answering any questions to trigger a validation error
     Page.verifyOnPage(WhoCompletedReviewPage) //
       .hasErrorCount(2)
@@ -85,7 +83,8 @@ context(`Review a prisoner's plan`, () => {
       .submitPage()
 
     // Next page is Review Notes page
-    Page.verifyOnPage(ReviewNotePage).hasCorrectBackLink().submitPage() // submit the page without answering any questions to trigger a validation error
+    Page.verifyOnPage(ReviewNotePage) //
+      .submitPage() // submit the page without answering any questions to trigger a validation error
     Page.verifyOnPage(ReviewNotePage)
       .hasErrorCount(1)
       .hasFieldInError('notes')
--- a/server/routes/reviewPlan/reviewCheckYourAnswersController.test.ts
+++ b/server/routes/reviewPlan/reviewCheckYourAnswersController.test.ts
@@ -1,0 +1,58 @@
+import { Request, Response } from 'express'
+import ReviewCheckYourAnswersController from './reviewCheckYourAnswersController'
+import ReviewCheckYourAnswersView from './reviewCheckYourAnswersView'
+import { getPrisonerContext } from '../../data/session/prisonerContexts'
+import aValidPrisonerSummary from '../../testsupport/prisonerSummaryTestDataBuilder'
+import { addCurrentPageToHistory } from '../pageFlowHistory'
+import ReviewPlanCompletedByValue from '../../enums/reviewPlanCompletedByValue'
+
+jest.mock('../pageFlowHistory', () => ({
+  addCurrentPageToHistory: jest.fn(),
+}))
+
+describe('ReviewCheckYourAnswersController', () => {
+  const controller = new ReviewCheckYourAnswersController()
+
+  const prisonNumber = 'A1234BC'
+  const prisonerSummary = aValidPrisonerSummary(prisonNumber)
+  const userName = 'Larry David'
+
+  let req: Request
+  let res: Response
+  let next: jest.Mock
+
+  beforeEach(() => {
+    req = {
+      params: { prisonNumber },
+      session: {},
+    } as unknown as Request
+
+    res = {
+      render: jest.fn(),
+      locals: { prisonerSummary, user: { name: userName } },
+    } as unknown as Response
+
+    jest.clearAllMocks()
+  })
+
+  describe('getReviewCheckYourAnswersView', () => {
+    it('should render the "Check Your Answers" page', async () => {
+      // Given
+      const reviewPlanDto = {
+        completedBy: ReviewPlanCompletedByValue.MYSELF,
+        reviewDate: '2024-03-09',
+        notes: 'Progress noted in review.',
+      }
+      getPrisonerContext(req.session, prisonNumber).reviewPlanDto = reviewPlanDto
+
+      const expectedViewData = new ReviewCheckYourAnswersView(prisonerSummary, reviewPlanDto, userName).renderArgs
+
+      // When
+      await controller.getReviewCheckYourAnswersView(req, res, next)
+
+      // Then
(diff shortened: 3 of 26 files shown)


server/routes/reviewPlan/reviewCheckYourAnswersController.ts
```diff
@@ -1,11 +1,18 @@
...
 import ReviewCheckYourAnswersView from './reviewCheckYourAnswersView'
+import { getPrisonerContext } from '../../data/session/prisonerContexts'
+import { addCurrentPageToHistory } from '../pageFlowHistory'
 
 export default class ReviewCheckYourAnswersController {
   getReviewCheckYourAnswersView: RequestHandler = async (req, res, next): Promise<void> => {
+    const { prisonNumber } = req.params
     const { prisonerSummary } = res.locals
+    const userName = res.locals.user.name
+    addCurrentPageToHistory(req)
```
lynnette-orr, line 11: Add the check your answers page to history to enable navigation from the review or notes pages if they've been reached from a Change link.

server/routes/reviewPlan/reviewNoteController.ts
```diff
@@ -19,7 +19,11 @@ export default class ReviewNoteController {
 
     getPrisonerContext(req.session, prisonNumber).reviewNoteForm = undefined
 
-    const view = new ReviewNoteView(prisonerSummary, reviewNoteForm)
+    const backlinkUrl = req.session.pageFlowHistory
```
lynnette-orr, line 22: Set backlink url based on PageFlowHistory, which will include the check your answers page if arriving from there.

server/views/pages/reviewPlan/checkYourAnswers.njk
```diff
@@ -12,7 +12,75 @@
...
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
-        <h1 class="govuk-heading-l">Check your answers</h1>
+      <h1 class="govuk-heading-l" data-qa="page-heading">Check
+        {{prisonerSummary.firstName + " " + prisonerSummary.lastName}}'s review information</h1>
+
+      <div class="app-u-heading-with-change-link">
+        <h2 class="govuk-heading-m">
+          {% if reviewPlanDto.completedBy === 'MYSELF' %}
+            Who completed the review?
```
nathanrussell-moj-digital, line 21: Are we saying the heading text is different based on who did the review?
That sounds wrong / overly complicated for the user - we should check this with the designers.

server/views/pages/reviewPlan/checkYourAnswers.njk
```diff
@@ -12,7 +12,75 @@
...
+            </dd>
+          </div>
+        {% endif %}
+      </dl>
+      <div class="app-u-heading-with-change-link">
+        <h2 class="govuk-heading-m">Review note</h2>
+        <span class="govuk-body-m">
+          <a class="govuk-link" href="/plan/{{ prisonerSummary.prisonNumber }}/review/notes" data-qa="review-note-change-link" style="white-space: nowrap;">
+            Change<span class="govuk-visually-hidden">
+              who completed the review</span>
```
nathanrussell-moj-digital, line 60: Minor - the visually hidden text is wrong here - this is the change link for the note 

Also, what is the inline style for white-space: nowrap for? (and in the change link above)

I think we should probably look to avoid inline styles 🤔 , but I'm wondering why we need it at all? Is it fixing a layout problem that we can perhaps fix some other way? Do our other check your answers pages have similar problems with their change links? 🤔

server/views/pages/reviewPlan/checkYourAnswers.njk
```diff
@@ -12,7 +12,75 @@
...
+          </a>
+        </span>
+      </div>
+      <dl class="govuk-summary-list govuk-!-margin-bottom-8">
+        <div class="govuk-summary-list__row">
+          <dt class="govuk-summary-list__key govuk-!-width-one-quarter">
+            {% if reviewPlanDto.completedBy === 'MYSELF' %}
+              Note added
```
nathanrussell-moj-digital, line 68: As above; do we really have different heading/label text for the note, depending on who added it?
Lets check this with the designers

nathanrussell-moj-digital (commented):
I've left a couple of comments/queries about some of the content - would be worth checking with the design team.

Also I'm not sure about the use of the pageFlowHistory .... 🤔 
I can see why you've thought to use it, as we use in in the Induction journey. But as you've probably seen the Induction journey is far more complex than this journey, and the pageFlowHistory stuff in the Induction journey is complex and messy!

I know we've not spoken about this, but the pageFlowHistory stuff (in Induction) is a bit of a mess that I'd like to get rid of one day!, and it's held on the session which is something we are trying to move away from (we want things on the prisoner context instead if we can / if it makes sense)

So with this in mind I'd rather we were not using PFH for this journey if we can avoid it, and I certainly don't want us adding extra methods to it (like addCurrentPageToHistory, especially as that might be confusing vs. methods of the same name in the Induction flow) You can see how I'm concerned that its got messy and I dont want us making it any more messy if we can help it 🤣 👍 

Anyway, I'm wondering if there is a simpler way for the "Who completed review" and "Review notes" page to know that the previous page was Check Your Answers, and therefore for them to know what the correct back link should be .... 🤔 

Let me give that some though ... but in the mean time if you want to check with Karen & co. re: the questions over the differing content/headings etc, that would be good 👍

server/routes/reviewPlan/reviewNoteController.ts
lynnette-orr: Changed this to a boolean to make things simpler.

server/views/pages/reviewPlan/checkYourAnswers.njk
lynnette-orr: Yes it's supposed to be different, confirmed by Karen.

server/views/pages/reviewPlan/checkYourAnswers.njk
lynnette-orr: Confirmed correct with Karen.

nathanrussell-moj-digital:
@lynnette-orr Re: the idea of doing the back links some other way .... this is what I had in mind ... 🤔 
https://dsdmoj.atlassian.net/browse/RR-849

NOTE: I am not asking you to do the above ticket, and your PR here should not change the way back links work for the Induction flow. But I wonder if we can take this screen journey (Reviews) as our first use case where we do back links differently, and effectively use it as the blue print for what we will do in RR-849 (as and when we pick that up)

Have a read of the above ticket and get a feel for what is being proposed there, but basically we are proposing using the javascript history API (which then means the Back links will do exactly what the browser Back button does 👍 rather than being a "forward" action). The only areas of complexity are that the cypress tests that test the back link will need changing (you wont be able to test the back link location anymore - as above, we are only doing Reviews at this stage), and because it uses javascript you'll need to hide the link for non-js browsers (in fact, the proposal is to hide it for everyone, and show it for js enabled browsers; but you get the idea)

Have a look / play and see what you think. Let me know how you get on 👍

server/routes/reviewPlan/reviewCheckYourAnswersController.ts
```diff
@@ -1,11 +1,17 @@
 import { RequestHandler } from 'express'
 import ReviewCheckYourAnswersView from './reviewCheckYourAnswersView'
+import { getPrisonerContext } from '../../data/session/prisonerContexts'
 
 export default class ReviewCheckYourAnswersController {
   getReviewCheckYourAnswersView: RequestHandler = async (req, res, next): Promise<void> => {
+    const { prisonNumber } = req.params
     const { prisonerSummary } = res.locals
+    const userName = res.locals.user.name
```
nathanrussell-moj-digital, line 9: I've just noticed in the screenshot in the jira ticket, that when the user did the review themselves, then on the Check Your Answers page, we should render "I did the review myself" rather than their name:
Screenshot 2024-11-01 at 08 51 39

This is a good thing as it means we dont need this code here to get the username, and we dont need it as a property on the `ReviewCheckYourAnswersView` class either 👍

server/views/pages/reviewPlan/checkYourAnswers.njk
nathanrussell-moj-digital: Have a look at the Education & Training tab , specifically the Educational Qualifications heading and change link:
Screenshot 2024-11-01 at 09 03 08

I think we've done essentially the same thing there, but without inline css:
```
            <div class="app-summary-card__change-link">
              <h3 class="govuk-heading-s app-summary-card__change-link__heading">Educational qualifications</h3>
              {%- set educationalQualificationsChangeLinkUrl -%}
                {# The target of the change link is different depending on whether the education record has qualifications already recorded on not #}
                {%- if education.educationDto.qualifications | length -%}
                  {# If the education record has qualifications the change link should be to the qualifications list screen, which allows the user to add/remove qualifications #}
                  /prisoners/{{ prisonerSummary.prisonNumber }}/education/qualifications
                {%- else -%}
                  {# If the education record does not have any qualifications the change link should be to a route that sets up the DTO in the prisoner context then redirects to the qualification level screen, which allows the user enter the first qualification #}
                  /prisoners/{{ prisonerSummary.prisonNumber }}/education/add-qualifications
                {%- endif -%}
              {%- endset -%}
              <a class="govuk-link app-summary-card__change-link__link govuk-!-display-none-print" href="{{ educationalQualificationsChangeLinkUrl }}" data-qa="educational-qualifications-change-link">
                Change<span class="govuk-visually-hidden"> qualifications by adding or removing qualifications</span>
              </a>
            </div>
```
The above has some complexity you wont need re: working out the URL to link to; and the above example is a h3 heading whereas you have yours as a h2 (which I think is correct for this page 👍 )
But you get the idea, and hopefully we can apply it to this page too 👍 🤞 
Basically:
* we have a parent `div` with class `app-summary-card__change-link`
* the heading element has a class `app-summary-card__change-link__heading`
* the change link anchor tag element has a class `app-summary-card__change-link__link`

Hopefully that's all you need 🤞

lynnette-orr: Refactored to use the formatting above, negating the need for inline style.

server/routes/reviewPlan/reviewCheckYourAnswersController.ts
lynnette-orr: Ah, I had been following the ACs for this one. Checked with Karen and the screenshot is correct.

Selected November 01 2024 14:18:37

nathanrussell-moj-digital (approved):

assets/js/backLink.js
```diff
@@ -1,4 +1,4 @@
-const backLink = document.querySelector('.govuk-back-link')
+const backLink = document.querySelector('.js-back-link')
```
nathanrussell-moj-digital, line 1: Using the class `.js-back-link` as the selector is better, both here for the JS and next file for the css, because if we use the standard govuk style as the selector then the back links on the Induction journey will be affected, and we dont want to do that (in this PR)

integration_tests/e2e/reviewPlan/reviewPlan.cy.ts
```diff
@@ -42,8 +42,7 @@ context(`Review a prisoner's plan`, () => {
     cy.visit(`/plan/${prisonNumber}/review`)
 
     // Then
-    Page.verifyOnPage(WhoCompletedReviewPage) //
-      .hasCorrectBackLink()
```
nathanrussell-moj-digital: I couldn't really see what value the new `hasCorrectBackLink` assertion had (what is "correct" about it?); so I've removed it

server/routes/reviewPlan/reviewCheckYourAnswersController.test.ts
```diff
@@ -49,7 +49,7 @@ describe('ReviewCheckYourAnswersController', () => {
       await controller.getReviewCheckYourAnswersView(req, res, next)
 
       // Then
-      expect(res.render).toHaveBeenCalledWith('pages/reviewPlan/checkYourAnswers', expectedViewData)
+      expect(res.render).toHaveBeenCalledWith('pages/reviewPlan/checkYourAnswers/index', expectedViewData)
```
nathanrussell-moj-digital, line 52: I should have spied this is an earlier PR, but I've moved the nunjucks template into a folder called `checkYourAnswers`, rather than it being called `checkYourAnswers.njk` (so it's consistent with the other Review based templates 👍 )

server/routes/reviewPlan/whoCompletedReviewController.test.ts
```diff
@@ -139,28 +139,5 @@ describe('whoCompletedReviewController', () => {
       expect(getPrisonerContext(req.session, prisonNumber).whoCompletedReviewForm).toBeUndefined()
       expect(getPrisonerContext(req.session, prisonNumber).reviewPlanDto).toEqual(reviewPlanDto)
     })
-
-    it(`should set backlinkUrl to Check Your Answers view given the previous page was Check Your Answers`, async () => {
```
nathanrussell-moj-digital: I dont think this test was testing anything over and above other tests in the class, and certainly nothing to do with the back link, so I've removed it

server/views/pages/reviewPlan/checkYourAnswers/index.njk
```diff
@@ -21,12 +21,7 @@
 
       <div class="app-summary-card__change-link">
         <h2 class="govuk-heading-m app-summary-card__change-link__heading govuk-!-margin-bottom-3">
-          {% if reviewPlanDto.completedBy === 'MYSELF' %}
```
nathanrussell-moj-digital: Based on this conversation (which I forgot to add you to! Sorry!), we don't need this logic 👍 
https://moj.enterprise.slack.com/archives/C07V4CVGV5W

PR #905, opened 2024-11-05
RR-991 - Added feature toggle function for enabled prisons

nathanrussell-moj-digital:
This PR adds a feature toggle function to determine if a given prison is one of the prisons enabled for the service.

We need this for the middleware (which I've also written) which will be used to set a flag as to whether to show the "prison is not onboarded" banner or not (to be developed in RR-1091)

I've also refactored the `reviewsPrisonsEnabled` feature toggle to be a function that is called in a similar way.

### DIFF
--- a/server/config.ts
+++ b/server/config.ts
@@ -177,8 +177,19 @@ export default {
   ),
   featureToggles: {
     // someToggleEnabled: toBoolean(get('SOME_TOGGLE_ENABLED', false)),
-    reviewsPrisonsEnabled: get('REVIEWS_PRISONS_ENABLED', ''),
     completedGoalsEnabled: toBoolean(get('COMPLETED_GOALS_ENABLED', false)),
     archiveGoalNotesEnabled: toBoolean(get('ARCHIVE_GOAL_NOTES_ENABLED', false)),
+    reviewJourneyEnabledForPrison: (prisonId: string): boolean => {
+      const reviewsPrisonsEnabled = get('REVIEWS_PRISONS_ENABLED', '')
+        .split(',')
+        .map(id => id.trim())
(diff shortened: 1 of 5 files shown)


server/config.ts
```diff
@@ -177,8 +177,19 @@ export default {
   ),
   featureToggles: {
     // someToggleEnabled: toBoolean(get('SOME_TOGGLE_ENABLED', false)),
-    reviewsPrisonsEnabled: get('REVIEWS_PRISONS_ENABLED', ''),
     completedGoalsEnabled: toBoolean(get('COMPLETED_GOALS_ENABLED', false)),
     archiveGoalNotesEnabled: toBoolean(get('ARCHIVE_GOAL_NOTES_ENABLED', false)),
+    reviewJourneyEnabledForPrison: (prisonId: string): boolean => {
```
nathanrussell-moj-digital, line 182: `reviewJourneyEnabledForPrison` is a refactor of the previous property `reviewsPrisonsEnabled`, and it means that the client code can simply call the function, rather than having to worry about do the splitting etc itself (eg. see the change to `server/routes/reviewPlan/index.ts` below 👇 )

server/config.ts
```diff
@@ -177,8 +177,19 @@ export default {
...
+      const reviewsPrisonsEnabled = get('REVIEWS_PRISONS_ENABLED', '')
+        .split(',')
+        .map(id => id.trim())
+      return reviewsPrisonsEnabled.includes(prisonId) || reviewsPrisonsEnabled.includes('***')
+    },
+    prisonIsEnabledForService: (prisonId: string): boolean => {
```
nathanrussell-moj-digital, line 188: This function is the new feature toggle function property, and can be called to determine whether a given prison is one of the enabled prisons or not.

lynnette-orr (approved):

PR #911, opened 2024-11-07
Rr-1092: remove check for viewer role

lynnette-orr:
Remove checkUserHasViewAuthority and ApplicationRoles.ROLE_EDUCATION_WORK_PLAN_VIEWER, update tests.

### DIFF
--- a/integration_tests/e2e/induction/inductionSecurityScenarios.cy.ts
+++ b/integration_tests/e2e/induction/inductionSecurityScenarios.cy.ts
@@ -20,7 +20,7 @@ context('Security tests for creating and updating Inductions', () => {
   describe('Creating Inductions', () => {
     it('should redirect to auth-error page given user does not have edit authority', () => {
       // Given
-      cy.task('stubSignIn')
+      cy.task('stubSignInAsReadOnlyUser')
       cy.signIn()
 
       const prisonNumber = 'G6115VJ'
@@ -49,7 +49,7 @@ context('Security tests for creating and updating Inductions', () => {
 
     it('should redirect to auth-error page given user does not have edit authority', () => {
       // Given
-      cy.task('stubSignIn')
+      cy.task('stubSignInAsReadOnlyUser')
       cy.signIn()
 
       const prisonNumber = 'G6115VJ'
--- a/integration_tests/e2e/induction/inductionSecurityScenarios.cy.ts
+++ b/integration_tests/e2e/induction/inductionSecurityScenarios.cy.ts
@@ -20,7 +20,7 @@ context('Security tests for creating and updating Inductions', () => {
   describe('Creating Inductions', () => {
     it('should redirect to auth-error page given user does not have edit authority', () => {
       // Given
-      cy.task('stubSignInAsUserWithViewAuthority')
+      cy.task('stubSignIn')
       cy.signIn()
 
       const prisonNumber = 'G6115VJ'
@@ -49,7 +49,7 @@ context('Security tests for creating and updating Inductions', () => {
 
     it('should redirect to auth-error page given user does not have edit authority', () => {
       // Given
-      cy.task('stubSignInAsUserWithViewAuthority')
+      cy.task('stubSignIn')
       cy.signIn()
 
       const prisonNumber = 'G6115VJ'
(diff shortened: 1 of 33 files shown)


integration_tests/e2e/induction/inductionSecurityScenarios.cy.ts
```diff
@@ -18,20 +18,6 @@ context('Security tests for creating and updating Inductions', () => {
   })
 
   describe('Creating Inductions', () => {
-    it('should redirect to auth-error page given user does not have any authorities', () => {
```
nathanrussell-moj-digital, line 21: Has this test started failing? which is why you're removed it? 🤔 
It shouldn't be failing, and is still a valid test - if you are a user with no authorities (you are effectively a read-only user), you should not be able to get to the `/create-induction/....` or `/induction/....` pages

I'm hoping both of these tests still pass and we can keep them

integration_tests/e2e/overview/goal/goalsPage.cy.ts
```diff
@@ -6,7 +6,7 @@ context('View goals', () => {
 
   beforeEach(() => {
     cy.task('reset')
-    cy.task('stubSignInAsUserWithEditAuthority')
+    cy.task('stubSignInAsUserWithViewAuthority')
```
nathanrussell-moj-digital, line 9: For all of these cypress tests, where we are saying its now a readonly page where you dont need any roles, I think we should change this to:
```
cy.task('stubSignIn')
```

In addition, I think you want to make an edit to `cypress.config.ts`:
```
    setupNodeEvents(on) {
      on('task', {
        reset: resetStubs,
        ...auth,

        stubSignIn: (roles: []) => auth.stubSignIn(roles),
        stubSignInAsUserWithEditAuthority: () => auth.stubSignIn(['ROLE_EDUCATION_WORK_PLAN_EDITOR']),
        stubSignInAsUserWithViewAuthority: () => auth.stubSignIn(['ROLE_EDUCATION_WORK_PLAN_VIEWER']),
```
and remove the `stubSignInAsUserWithViewAuthority` setup altogether, because we are basically saying that that role no longer exists any more 👍

integration_tests/e2e/overview/workAndInterestsTab.cy.ts
```diff
@@ -307,5 +409,21 @@ context('Prisoner Overview page - Work and Interests tab', () => {
       // Then
       Page.verifyOnPage(InPrisonWorkPage)
     })
+
+    it(`should link to auth-error page given user has read-only authority`, () => {
```
nathanrussell-moj-digital, line 413: Ooooh, I think this is wrong, but is actually a sign of a different problem which I think we should fix first ....

So, all these Change links should only be shown if you have the EDITOR role ... but at the moment we show them regardless 😬 

I think we need to part this PR for the time being, and create a new branch and PR to fix up all the Change and Add links on the Education & Training, and Work Interests pages.

Basically we need to change all instances of this type of thing:
```
              <a class="govuk-link app-summary-card__change-link__link govuk-!-display-none-print" href="{{ educationalQualificationsChangeLinkUrl }}" data-qa="educational-qualifications-change-link">
                Change<span class="govuk-visually-hidden"> qualifications by adding or removing qualifications</span>
              </a>
```
and wrap it in an if block, like this:
```
            {% if hasEditAuthority %}
              <a class="govuk-link app-summary-card__change-link__link govuk-!-display-none-print" href="{{ educationalQualificationsChangeLinkUrl }}" data-qa="educational-qualifications-change-link">
                Change<span class="govuk-visually-hidden"> qualifications by adding or removing qualifications</span>
              </a>
            {% endif %}
```
You don't need to add the `hasEditAuthority` to all the View classes, it will already be there. The if statement above should work 🤞

nathanrussell-moj-digital:
Ahhh, we probably didnt want to merge your other branch into this one.
Firstly I'm not sure about some of the proposed changes in your other branch/PR (I've left comments there)

But also it means this PR is now massive! It's a 48 file PR ! 😬  And the whole point of fixing the Change/Add links on the Education & Training and Work Interests tabs (your other PR) was to make _this_ one simpler, not make it bigger! 🤣 

Lets talk about both of the branches/PRs tomorrow 👍

integration_tests/e2e/induction/inductionSecurityScenarios.cy.ts
nathanrussell-moj-digital: Ah, I now understand why you deleted this test, and I agree its the right thing to have done 👍

nathanrussell-moj-digital (approved):

PR #920, opened 2024-11-19
Rr-1045: "Exemption reason" screen

lynnette-orr:
https://github.com/user-attachments/assets/750ea813-907f-4864-bb07-988f2afdc549


server/views/pages/reviewPlan/exemption/index.njk
```diff
@@ -0,0 +1,207 @@
...
+          hint: {
+            html: warningTextHtml
+          },
+          items: [
+            {
+              value: "EXEMPT_PRISONER_DRUG_OR_ALCOHOL_DEPENDENCY",
+              text: "EXEMPT_PRISONER_DRUG_OR_ALCOHOL_DEPENDENCY" | formatExemptionReasonValue,
+              checked: form.exemptionReason === "EXEMPT_PRISONER_DRUG_OR_ALCOHOL_DEPENDENCY",
+              conditional: {
+                html: exemptionReasonDetailsInput(
```
nathanrussell-moj-digital, line 62: Neat use of the macro 👍

server/views/pages/reviewPlan/exemption/index.njk
```diff
@@ -0,0 +1,207 @@
+{% extends "../../../partials/layout.njk" %}
+
+{% set pageTitle = "Exemption" %}
+{% set pageId = "exemption" %}
```
nathanrussell-moj-digital, line 4: I think the pageId and pageTitle need to be a bit more explicit. In our other pages we have this type of thing:

```
{% set pageTitle = "Who completed the prisoner's plan review?" %}

{% set pageId = "review-plan-who-completed-review" %}
```

So here perhaps the pageId  should be `review-plan-exemption-reason`; and the pageTitle should be something like `Put the prisoner's review on hold` or something? 🤷‍♂️  (we're not allowed to put the prisoner name/id in the `pageTitle` here, so we can't just copy/duplicate the H1)

Also (thinking about the other pages we are about to write for this feature) I think the folder would be better off being called `exemptionReason` (because the other pages might be in folders `exemptionConfirm` (or `exemptionAreYouSure`), and `exemptionComplete`)

nathanrussell-moj-digital (commented):
You've made a good start on this, but a few things to have a think about / consider.
Let me know what you think - happy to chat it through if it helps

PR #922, opened 2024-11-21
Rr-1046: Are you sure you want to put review on hold screen

lynnette-orr:
https://github.com/user-attachments/assets/f68ede5b-be20-448d-ac57-38b0bc8caf86


nathanrussell-moj-digital (approved):
I'm going to approve this so you can merge it which will allow us to move on .... but ....

I think there are elements of this PR that have taken us, or are about to take us, down the wrong road, and we need to address those before we consider the next screen.

Specifically I think we need a DTO, which we're going to need to think a little bit about ... 🤔 , and it's likely that we'll end up refactoring/replacing a few parts of what you have in this PR.

Once you've merged this do not start on the next screen, and instead can you think about a DTO please ... I think the DTO itself will be pretty simple, possibly just having 2 fields. The complexity will be around designing when and where it is created and checked. Populating it in the exemptionReasons controller should be easy, and reading it back in your new confirmExemptionController and adding to the new view class should be easy.

Anyway, happy to talk through ideas/thoughts etc, but I'm not free until tomorrow morning, so if you can have a think/play (small, iterative PRs are good 👍 😁 ), then we can chat tomorrow if you need it.

PR #940, opened 2024-11-28
RR-1156 - Wire up the calls to the CreateReview service; unit tests; integration tests

nathanrussell-moj-digital:
This PR wires up the call to the CreateReview service at the end of the Review journey, along with associated unit tests.

It also updates the existing Review journey integration tests to assert the API was called correctly.

There were also a small handful of changes that were necessary to the controllers and nunjucks templates in order to create the fully joined up end-to-end, as tested by the integration tests. (These were not possible without the service call, so it is right that they were picked up as part of this)


integration_tests/e2e/reviewPlan/reviewComplete.cy.ts
nathanrussell-moj-digital, line 1: I dont think we need this test, as the other 2 Review cypress tests cover everything that was in this test file

server/@types/dto/index.d.ts
```diff
@@ -84,6 +84,9 @@ declare module 'dto' {
     completedByOtherJobRole?: string
     reviewDate: Date
     notes?: string
+    wasLastReviewBeforeRelease?: boolean
```
nathanrussell-moj-digital, line 87: Needed to add these to the DTO, to carry the properties from the response data of calling the API (IE. The Review Complete page needs to know whether this was the last review and the next review dates)

server/@types/viewModels/index.d.ts
```diff
@@ -133,7 +133,7 @@ declare module 'viewModels' {
     problemRetrievingData: boolean
   }
 
-  export interface CreatedActionPlan {
+  export interface CreatedActionPlanReview {
```
nathanrussell-moj-digital, line 136: Minor naming change to better reflect its use

lynnette-orr (approved):

PR #942, opened 2024-11-28
Rr 1028: Action card updates

lynnette-orr:
Action plan review status 'completed'
Screenshot 2024-11-28 at 17 25 35

Review date from is in the future
Screenshot 2024-11-28 at 17 23 34

Review is due
Screenshot 2024-11-28 at 17 01 11

Review is overdue
Screenshot 2024-11-28 at 16 59 59

No induction
Screenshot 2024-11-28 at 16 58 27


assets/scss/local.scss
```diff
@@ -49,11 +49,34 @@
...
-   white-space: nowrap; 
-   padding-right: 17px;
+.govuk-tag {
+  &--sessions-tag {
+    white-space: nowrap;
+    padding-right: 1.5ch;
+  }
+}
+
+.actions-card-reviews-tag-yellow {
```
nathanrussell-moj-digital, line 61: Minor; but would these 3 classes logically sit better within `_action-card.scss`

server/views/components/actions-card/template.njk
```diff
@@ -20,15 +20,32 @@
         {% for action in params.actions %}
 
           {% if action['render-if'] == null or action['render-if'] === true %}
-            {#- Set classes for this action #}
-            {%- set classNames = "govuk-link" -%}
-            {%- if action.classes %}
-              {% set classNames = classNames + " " + action.classes %}
+            {%- if action.notAUrl %}
```
nathanrussell-moj-digital, line 23: This looks a little complicated, especially where it is being used in the main Overview template (that is particularly hard to read 😬 )

The idea with these components is that they are as generic as we can make them, and have no knowledge of how they are being called. I know you've only used the term "review" in the comment, but the class names (heavy-border, border, no-border) are all from a specific design requirement of the reviews layout.
I think we can do this better .... 🤔

server/views/pages/overview/partials/overviewTab/overviewTabContents.njk
```diff
@@ -44,9 +44,90 @@ Data supplied to this template:
...
+    {% set dueOrOnTimeTagTitle = '<span>' ~ dueOrOnTimeTagContent ~ '</span>' %}
+    {% set overdueTagContent = 'Review was due by ' ~ reviewDateTo | formatDate('D MMM YYYY') %}
+    {% set overdueTagTitle = '<span>' ~ overdueTagContent ~ '</span>' %}
+    {% set releaseDateContent = 'Release on ' ~ releaseDate | formatDate('D MMMM YYYY') %}
+    {% set releaseDateTitle = '<span>' ~ releaseDateContent ~ '<span>' %}
+
     <div class="govuk-grid-column-one-third app-u-print-full-width">
       {{ actionsCard({
         actions: [
+          {
```
nathanrussell-moj-digital, line 57: I can see why you've done this (because the design calls for elements in the action card that are not actions!, such as headings and tags). This kinda sucks (not what you've done; the design and how it impacts our actions-card component sucks)
I think we need to think about this a little ... 🤔 

One thing that might make this easier / neater, and you've not addressed this in your PR here, is that all these review based actions are only to be displayed if the reviews feature toggle is enabled for the prison.
We could add a load more logic to each action-item's `render-if` statements 😢 ; or we could treat it as 2 components entirely, where the original one is shown if the feature toggle is not enabled and it is post-induction, else show the other one.

server/views/pages/overview/partials/overviewTab/overviewTabContents.njk
```diff
@@ -44,9 +44,90 @@ Data supplied to this template:
...
+            }
+          },
+           {
+            title: releaseDateTitle | safe,
+            'render-if': induction.isPostInduction and reviewStatus === 'noReviewsDue',
+            notAUrl: true,
+            heavyBorder: true
+          },
+          {
+            title: '<span class="goal-action"><img src="/assets/images/icon-mark-review-complete.svg" role="presentation" alt="" width="30px" height="24px" class="goal-icon" />Mark review as complete</span>' | safe,
```
nathanrussell-moj-digital, line 108: The more I think about it, the more I think the macro/actions-card component is not actually helping us.
There's so much "code" here, and where some of them have markup and href's within them .... I think we'd be better off not using the action card component at all on this page, and just writing the markup in the page.

server/routes/overview/overviewView.ts
```diff
@@ -51,6 +52,9 @@ type RenderArgs = {
     problemRetrievingData: boolean
     isPostInduction: boolean
   }
+  reviewStatus: string
+  reviewDateTo: Date | null
+  releaseDate: Date
```
nathanrussell-moj-digital, line 57: I don't think we need `releaseDate` as a separate property here because `prisonerSummary` is already part of the rendered data and that already includes `releaseDate`. I don't think we need to duplicate it do we?

Also, I would put the other 2 fields in an object which contains `problemRetrievingData` (because we might have had a problem retrieving the action plan data and therefore should not attempt to show Review based stuff on the Overview page because we don't know the data). I would also use a enum-like values for the status.
Something like this:
```
    ....
    actionPlanReview: {
      reviewStatus: 'DUE' | 'OVERDUE' | 'NO_SCHEDULED_REVIEW'
      reviewDueDate: Date | null
    }
    ....
```

server/routes/overview/overviewView.ts
```diff
@@ -103,6 +107,24 @@ export default class OverviewView {
       prisonerHasHadInduction && !this.actionPlanReviews?.problemRetrievingData && mostRecentReviewSession == null
     const prisonerHasHadInductionAndAtLeastOneReview = prisonerHasHadInduction && mostRecentReviewSession != null
 
+    const reviewDateFrom = this.actionPlanReviews?.latestReviewSchedule?.reviewDateFrom || null
+    const reviewDateTo = this.actionPlanReviews?.latestReviewSchedule?.reviewDateTo || null
+    const noReviewsDue = this.actionPlanReviews?.latestReviewSchedule?.status === ActionPlanReviewStatusValue.COMPLETED
+    const releaseDate = this.prisonerSummary.releaseDate || null
+
+    const currentDate = new Date()
```
nathanrussell-moj-digital, line 115: I would call this `today` and init as with date-fns:
```
const today = startOfToday()
```

Then, in your logic, I wonder if using some of the date-fns functions such as [isBefore](https://date-fns.org/v4.1.0/docs/isBefore) (and/or isAfter etc) might make the code more readable?

nathanrussell-moj-digital:
Closed, as implemented in 2 separate PRs

PR #954, opened 2024-12-09
RR-227 - Introduced client and service methods to be able to call the createActionPlan API

nathanrussell-moj-digital:
This PR introduces the client and service methods to be able to call the `createActionPlan` API (rather than the UI calling the `createGoals` method when it wants to create the action plan)

This PR only introduces the methods, it does not wire them in or use them.

My next PR will wire the call to the service method into the relevant points in the user journey 👍

### DIFF
--- a/server/@types/educationAndWorkPlanApi/index.d.ts
+++ b/server/@types/educationAndWorkPlanApi/index.d.ts
@@ -4,6 +4,54 @@
  */
 
 export interface paths {
+  '/queue-admin/retry-dlq/{dlqName}': {
+    parameters: {
+      query?: never
+      header?: never
+      path?: never
+      cookie?: never
+    }
+    get?: never
+    put: operations['retryDlq']
+    post?: never
+    delete?: never
+    options?: never
+    head?: never
+    patch?: never
+    trace?: never
+  }
+  '/queue-admin/retry-all-dlqs': {
+    parameters: {
+      query?: never
+      header?: never
+      path?: never
+      cookie?: never
+    }
+    get?: never
+    put: operations['retryAllDlqs']
+    post?: never
+    delete?: never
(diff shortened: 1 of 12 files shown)


server/@types/educationAndWorkPlanApi/index.d.ts
```diff
@@ -4,6 +4,54 @@
  */
 
 export interface paths {
+  '/queue-admin/retry-dlq/{dlqName}': {
```
nathanrussell-moj-digital, line 7: Re-imported the swagger spec, which has brought in with it lots of unrelated change such as the DLQ endpoints etc.

The change we really wanted to bring in by importing the swagger spec was the removal of the `reviewDate` field, which is below 👇

server/data/educationAndWorkPlanClient.ts
```diff
@@ -30,16 +31,27 @@ export default class EducationAndWorkPlanClient {
...
-      data: createGoalsRequest,
+  async getActionPlan(prisonNumber: string, token: string): Promise<ActionPlanResponse> {
+    return EducationAndWorkPlanClient.restClient(token).get<ActionPlanResponse>({
+      path: `/action-plans/${prisonNumber}`,
     })
   }
 
-  async getActionPlan(prisonNumber: string, token: string): Promise<ActionPlanResponse> {
-    return EducationAndWorkPlanClient.restClient(token).get<ActionPlanResponse>({
+  async createActionPlan(
```
nathanrussell-moj-digital, line 40: The diff looks bigger than it is because I moved/re-ordered the methods `getActionPlan` and createGoals`; but really what we have here is the introduction of the new `createActionPlan` method

stevomcallister (approved):

PR #957, opened 2024-12-09
RR-227 - Tidied and simplified uses of getActionPlan

nathanrussell-moj-digital:
This PR tidies and simplifies the uses of `getActionPlan` in preparation for changing the API to return a 404 when no action plan exists.

`getActionPlan` is an API that exists to return the prisoner's action plan (collection of goals).
There is a TODO on the API service method to make it return a 404 when the UI is calling the new `createActionPlan` API endpoint:
```
  /**
   * Retrieves a Prisoner's [ActionPlan] based on their prison number.
   * Returns a new [ActionPlan] if the [ActionPlan] cannot be found.
   */
  fun getActionPlan(prisonNumber: String): ActionPlan {
    log.debug { "Retrieving Action Plan for prisoner [$prisonNumber]" }
    // TODO RR-227 - return 404 if not found
    val actionPlan = persistenceAdapter.getActionPlan(prisonNumber)
      ?: ActionPlan(
        reference = UUID.randomUUID(),
        prisonNumber = prisonNumber,
        goals = emptyList(),
      )
    actionPlan.goals.onEach { it.notes = goalNotesService.getNotes(it.reference) }
    return actionPlan
  }
```

This PR has brought together and simplified all of the calls that the UI makes to `getActionPlan` so that we know where we need to make the changes to support a 404 response.

### DIFF
--- a/server/routes/archiveGoal/archiveGoalController.ts
+++ b/server/routes/archiveGoal/archiveGoalController.ts
@@ -1,5 +1,6 @@
 import type { RequestHandler, Request } from 'express'
 import type { ArchiveGoalForm } from 'forms'
+import type { Goal } from 'viewModels'
 import createError from 'http-errors'
 import EducationAndWorkPlanService from '../../services/educationAndWorkPlanService'
 import ArchiveGoalView from './archiveGoalView'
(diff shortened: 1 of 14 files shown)


server/routes/archiveGoal/archiveGoalController.ts
```diff
@@ -18,18 +19,19 @@ export default class ArchiveGoalController {
...
   getArchiveGoalView: RequestHandler = async (req, res, next): Promise<void> => {
     const { prisonNumber, goalReference } = req.params
-    const { prisonerSummary } = res.locals
+    const { prisonerSummary, allGoalsForPrisoner } = res.locals
 
     let { archiveGoalForm } = getPrisonerContext(req.session, prisonNumber)
     if (!archiveGoalForm || archiveGoalForm.reference !== goalReference) {
-      const actionPlan = await this.educationAndWorkPlanService.getActionPlan(prisonNumber, req.user.username)
-      if (actionPlan.problemRetrievingData) {
+      if (allGoalsForPrisoner.problemRetrievingData) {
```
nathanrussell-moj-digital, line 26: Basically all of the changes in this PR are the same as this ....
Instead of the controller method calling the service method I have changed it so that the route definition (`index.ts` 👇 ) calls the service via a middleware function. The controller is simpler as it can safely assume the middleware has been called and prisoner goals are on `res.locals.allPrisonerGoals`

But in respect of being able to better support when the API starts returning a 404, the change will only be in 1 or 2 places (probably the service method and the middleware)

PR #958, opened 2024-12-09
RR-227 - Make use of retrieveGoals by status

nathanrussell-moj-digital:
This PR builds upon my recent PRs re: how the UI gets the action plan & goals.
The changes in this PR are to make use of `retrieveGoals` by status middleware where it makes more sense to do so, rather than getting the entire action plan.

Specifically this is on the controller routes for archiving, completing, unarchiving and updating goals. My previous PR made changes to these routes so that they retrieved all goals for the prisoner then filtered the list to get the specific goal in question.
The changes in this PR add a minor efficiency to this by using the `retrieveGoals` (by status) middleware. So each of these controller routes now only gets a subset of the goals (based on the status that we know the goal should be in for the given operation), and then filters to get the specific goal.

A longer term improvement for these routes would be a middleware that gets a specific goal by it's ID, but that's one for another day.
The real purpose of this PR was to differentiate the calls to get the prisoner's action plan vs. those places that only need the goals.

### DIFF
--- a/server/routes/archiveGoal/index.ts
+++ b/server/routes/archiveGoal/index.ts
@@ -3,7 +3,8 @@ import { Services } from '../../services'
 import ArchiveGoalController from './archiveGoalController'
 import { checkUserHasEditAuthority } from '../../middleware/roleBasedAccessControl'
 import asyncMiddleware from '../../middleware/asyncMiddleware'
-import retrieveAllGoalsForPrisoner from '../routerRequestHandlers/retrieveAllGoalsForPrisoner'
+import retrieveGoals from '../routerRequestHandlers/retrieveGoals'
+import GoalStatusValue from '../../enums/goalStatusValue'
 
 /**
  * Route definitions for the pages relating to Archiving A Goal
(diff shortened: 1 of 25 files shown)


server/routes/archiveGoal/index.ts
```diff
@@ -14,7 +15,7 @@ export default (router: Router, services: Services) => {
 
   router.use('/plan/:prisonNumber/goals/:goalReference/archive', [checkUserHasEditAuthority()])
   router.get('/plan/:prisonNumber/goals/:goalReference/archive', [
-    retrieveAllGoalsForPrisoner(services.educationAndWorkPlanService),
+    retrieveGoals(services.educationAndWorkPlanService, GoalStatusValue.ACTIVE),
```
nathanrussell-moj-digital, line 18: Similar changes in the other controller route definitions - rather than retrieve all prisoner goals, we'll just get the goals in the status relevant to the operation in question.
For example, here for the "archive goal" route, we only want to get the ACTIVE goals, because we know you can't archive a completed or already archived goal.

Basically the changes in the other route defs are the same kind of idea 👍

stevomcallister (approved):

PR #960, opened 2024-12-10
Correct the logging for when the Education and Induction APIs return a 404

nathanrussell-moj-digital:
This PR tidies up and corrects some logging which is causing noise in the app insights logs. At the moment it's hard to see the wood for the trees in the UI logs.

The problem area is how the UI calls the API to get the prisoner's Induction, and also the prisoner's Education.
The API implementation (correctly) returns a 404 for both of these endpoints when the requested resource does not exist. IE. if the prisoner has no Induction it returns a 404. Equally, when the prisoner has no Education record it returns a 404.

As far as the UI is concerned though, these are not error conditions. It is valid for the prisoner to not have an Induction, and valid for them not to have an Education record. Indeed, for both of these use cases the absence of the resource drives the UI journey down certain paths.

But what is currently happening is that, whilst the UI journey is correct, the 404 error response is logged as an error. So we are seeing lots (and lots!) of errors logged to app insights (via `log.error`) for what is actually normal/expected processing.

### DIFF
--- a/server/routes/routerRequestHandlers/createEmptyInductionIfNotInSession.test.ts
+++ b/server/routes/routerRequestHandlers/createEmptyInductionIfNotInSession.test.ts
@@ -36,15 +36,7 @@ describe('createEmptyInductionIfNotInSession', () => {
     // Given
     req.session.inductionDto = undefined
 
-    const educationServiceError = {
-      status: 404,
-      data: {
-        status: 404,
-        userMessage: `Education not found for prisoner [${prisonNumber}]`,
-        developerMessage: `Education not found for prisoner [${prisonNumber}]`,
-      },
-    }
-    educationAndWorkPlanService.getEducation.mockRejectedValue(educationServiceError)
+    educationAndWorkPlanService.getEducation.mockResolvedValue(undefined)
 
     const expectedInduction = { prisonNumber }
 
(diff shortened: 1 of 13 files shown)


server/routes/routerRequestHandlers/createEmptyInductionIfNotInSession.test.ts
```diff
@@ -36,15 +36,7 @@ describe('createEmptyInductionIfNotInSession', () => {
...
-    const educationServiceError = {
-      status: 404,
-      data: {
-        status: 404,
-        userMessage: `Education not found for prisoner [${prisonNumber}]`,
-        developerMessage: `Education not found for prisoner [${prisonNumber}]`,
-      },
-    }
-    educationAndWorkPlanService.getEducation.mockRejectedValue(educationServiceError)
+    educationAndWorkPlanService.getEducation.mockResolvedValue(undefined)
```
nathanrussell-moj-digital, line 39: The basic change is that the service layer now checks the status code from the client response. If it's a 404 then it returns undefined, otherwise it continues to throw back the error.

server/routes/routerRequestHandlers/retrieveEducation.ts
```diff
@@ -16,31 +16,12 @@ const retrieveEducation = (educationAndWorkPlanService: EducationAndWorkPlanServ
         educationDto: await educationAndWorkPlanService.getEducation(prisonNumber, req.user.username),
       }
     } catch (error) {
-      res.locals.education = { ...gracefullyHandleException(error, prisonNumber), educationDto: undefined }
+      logger.debug('Error retrieving Education data', error)
```
nathanrussell-moj-digital, line 19: The code that calls the service method no longer needs to handle any exception (gracefully or otherwise 🤣 ) because it is now handled in the service method itself

nathanrussell-moj-digital:
Closing as I've thought of a much simpler approach to achieve the same result

PR #968, opened 2024-12-16
RR-1027 - Correct the count in the Session history box

nathanrussell-moj-digital:
This PR corrects the count in the Session history box on the main Overview screen.

The session count is the number of sessions held with a prisoner, where a "session" is their initial Induction, plus any Reviews.

The initial development of this count considered the Induction as simply being that the Induction record exists.
However, the Induction is only considered complete when there is both the Induction + at least 1 goal.

This PR corrects that logic.

### DIFF
--- a/server/routes/overview/overviewController.test.ts
+++ b/server/routes/overview/overviewController.test.ts
@@ -125,7 +125,7 @@ describe('overviewController', () => {
         counts: {
           inductionSessions: 1,
           reviewSessions: 1,
-          totalSessions: 2,
+          totalCompletedSessions: 2,
         },
         lastSessionConductedAt: parseISO('2023-06-19T09:39:44.000Z'),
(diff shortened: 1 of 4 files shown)


server/routes/overview/overviewController.test.ts
```diff
@@ -125,7 +125,7 @@ describe('overviewController', () => {
         counts: {
           inductionSessions: 1,
           reviewSessions: 1,
-          totalSessions: 2,
+          totalCompletedSessions: 2,
```
nathanrussell-moj-digital, line 128: As well as fixing the bug, I've renamed the terminology to be the total completed sessions
(because just having an Induction record does not mean the Induction is complete - you need at least 1 goal as well for it to be considered complete)

stevomcallister (approved):

PR #974, opened 2024-12-18
RR-1028 - No review actions displayed when the prisoner has no scheduled review

nathanrussell-moj-digital:
Currently there is a small bug with the Reviews UI stuff we have been doing recently with the Actions menu; in that if the prisoner does not yet have a Review Schedule setup for any reason (perhaps they've not yet had their Induction etc), then we display "no reviews due". Whilst arguably this might be correct, its not what the business intended. They intended the "no reviews due" content to show when the prisoner has had their last review before release, and hence another review is not due. This is why it also shows the release date

### Current screen when the prisoner has no Review Schedule set up
Screenshot 2024-12-18 at 19 10 02

### Fix, where the prisoner has no Review Schedule set up
Screenshot 2024-12-18 at 19 11 58

### DIFF
--- a/server/routes/overview/overviewView.ts
+++ b/server/routes/overview/overviewView.ts
@@ -111,23 +111,19 @@ export default class OverviewView {
       prisonerHasHadInduction && !this.actionPlanReviews?.problemRetrievingData && mostRecentReviewSession == null
     const prisonerHasHadInductionAndAtLeastOneReview = prisonerHasHadInduction && mostRecentReviewSession != null
 
-    const scheduledReviewExists = this.actionPlanReviews?.latestReviewSchedule != null
-    const hasHadLastReview =
-      scheduledReviewExists &&
-      this.actionPlanReviews.latestReviewSchedule.status === ActionPlanReviewStatusValue.COMPLETED
+    const reviewScheduleDoesNotExist = this.actionPlanReviews?.latestReviewSchedule == null
+    const latestReviewScheduleStatus = this.actionPlanReviews?.latestReviewSchedule?.status
+    const hasHadLastReview = latestReviewScheduleStatus === ActionPlanReviewStatusValue.COMPLETED
     const reviewOnHold =
-      scheduledReviewExists &&
-      this.actionPlanReviews.latestReviewSchedule.status !== ActionPlanReviewStatusValue.SCHEDULED &&
-      this.actionPlanReviews.latestReviewSchedule.status !== ActionPlanReviewStatusValue.COMPLETED
-    const reviewDateFrom = scheduledReviewExists
-      ? this.actionPlanReviews.latestReviewSchedule.reviewDateFrom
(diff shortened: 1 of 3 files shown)


server/routes/overview/overviewView.ts
```diff
@@ -111,21 +111,26 @@ export default class OverviewView {
       prisonerHasHadInduction && !this.actionPlanReviews?.problemRetrievingData && mostRecentReviewSession == null
     const prisonerHasHadInductionAndAtLeastOneReview = prisonerHasHadInduction && mostRecentReviewSession != null
 
-    const noScheduledReview =
-      this.actionPlanReviews?.latestReviewSchedule == null ||
-      this.actionPlanReviews?.latestReviewSchedule.status === ActionPlanReviewStatusValue.COMPLETED
+    const scheduledReviewExists = this.actionPlanReviews?.latestReviewSchedule != null
+    const hasHadLastReview =
```
stevomcallister, line 115: This logic doesn't look right to me. I feel like it should be  

`!scheduledReviewExists & this.actionPlanReviews.latestReviewSchedule.status === ActionPlanReviewStatusValue.COMPLETED`

But I can see that your screen shot is displaying as it should so it must be right.

stevomcallister (approved):

server/routes/overview/overviewView.ts
nathanrussell-moj-digital: Ah, the logic _is_ correct, but maybe the problem is the variable name `scheduledReviewExists`

What that variable is there to do is determine whether _any_ Review Schedule object exists for the  prisoner, by looking at the presence of the `latestReviewSchedule` property of their Action Plan Reviews (which comes from [this endpoint](https://learningandworkprogress-api-dev.hmpps.service.justice.gov.uk/swagger-ui/index.html#/Action%20Plan%20Reviews/get-action-plan-review))
For the most part it's to save a load of complicated null checks in the next few lines of code (where it checks for various statuses), but it's also used as part of the core fix for this bug, so that we can differentiate between a prisoner having no Review Schedules setup at all (ie. prisoner who has not even had their induction yet), vs. a prisoner who has had their last review and is therefore not scheduled for another.

I think the problem is that of perhaps slightly confusing terminology .... The object is called `ReviewSchedule` and it has a status field to indicate the status of the schedule (scheduled, exempt etc). I guess it's kinda tenuous as to whether an exempted Review Schedule is still scheduled or not ? .... and I guess it's becomes even more tenuous when we say a Review Schedule is Complete ? 🤔 

Anyway, the logic is right, and perhaps the new variable that I introduced would be better being called something like `noScheduledReview` as before (but then all the places that use it below would need to negate it. Maybe `prisonerHasAReviewScheduleInAnyState` .... but it's a bit wordy ...

What do you think?

stevomcallister (approved):

PR #997, opened 2025-01-21
RR-1179 - Refactor actionCards component

nathanrussell-moj-digital:
This PR refactors the `actionCards` UI template into a proper nunjucks component/macro. This is a better pattern because it means we can write proper unit tests for it (which I have done), and also allows for easier future maintenance (such as adding the Induction based actions, which is what we need to do in this story)

The PR also refactors (renames) the nunjucks filter that maps Review Exemption reasons into human readable labels. The reason we need to do this is because it was slightly vaguely named before (it didn't explicitly mention "reviews" in its name), and in the next PR for this story we will need to introduce a very similar filter that maps Induction Excemption reasons into human readable labels.


server/@types/viewComponents/index.d.ts
```diff
@@ -2,13 +2,8 @@
  * Module declaring types that are used by the nunjucks view components
  */
 declare module 'viewComponents' {
-  import type { Goal } from 'viewModels'
-
-  export interface ActionCardParams {
```
nathanrussell-moj-digital, line 7: `ActionCardParams` was an early version of a type used in an early version of an action card component. It has not been used for some time, so tidying it up 👍

stevomcallister (approved):

PR #999, opened 2025-01-22
RR-1179 - Add Induction actions to actionCards component

nathanrussell-moj-digital:
PR to add the new Inductions actions to actionCards component.

The [ticket](https://dsdmoj.atlassian.net/browse/RR-1179) describes 7 permutations of how the Induction actions should look, based on various states of the Induction Schedule etc.
These are all done, and proven with unit tests.

They are not so easy to see/visualise at the moment though - whilst we have the prison.admission etc listeners running in dev which is creating new InductionSchedules, most of them require the data "tweaking" 😁  in order to drive the different states etc.
I have however done it for the following 2 scenarios, which along with the unit tests that cover all cases, should be enough to give us the early confidence that this is correct:

### Overdue induction
Screenshot 2025-01-22 at 16 11 36

### Induction due
Screenshot 2025-01-22 at 16 13 14


stevomcallister:
> PR to add the new Inductions actions to actionCards component.
> 
> The [ticket](https://dsdmoj.atlassian.net/browse/RR-1179) describes 7 permutations of how the Induction actions should look, based on various states of the Induction Schedule etc. These are all done, and proven with unit tests.
> 
> They are not so easy to see/visualise at the moment though - whilst we have the prison.admission etc listeners running in dev which is creating new InductionSchedules, most of them require the data "tweaking" 😁 in order to drive the different states etc. I have however done it for the following 2 scenarios, which along with the unit tests that cover all cases, should be enough to give us the early confidence that this is correct:

Once it's in DEV we can put messages on the queues and see it come to life 🤞

stevomcallister (approved):

PR #1006, opened 2025-01-23
Update dependencies from template

nathanrussell-moj-digital:
This PR updates the dependencies and associated code in the PLP UI project, from the `typescript-template` project.

Whilst we update the dependencies regularly based on dependabot and renovate scans, there are some changes that have happened in the typescript-template that we've not kept up to date with, and by not updating those there are a few dependencies that dependabot & renovate could not easily update (specifically things like using `esbuild` instead of `eslint`, and not using `cookie-session`)

Whilst it looks like a fairly big PR, the vast majority of changes has come straight from the same files in `typescript-template`. For any changed file that is not from `typescript-template` I've added a comment to the PR to explain the change


assets/js/applicationinsights.js
```diff
@@ -1,5 +1,5 @@
 /* global Microsoft */
-window.applicationInsights = (function () {
+window.applicationInsights = (() => {
```
nathanrussell-moj-digital, line 2: The new linter didn't like anonymous functions like we had them previously, so I've changed them to ES5 function expressions

assets/js/applicationinsights.js
```diff
@@ -9,13 +9,9 @@ window.applicationInsights = (function () {
...
-            contentName: function (element) {
-              // If there is a id, use this as the content name
-              if (element.dataset.id) return element.dataset.id
-
-              // If this element is in the header or footer it could contain
-              // personal data so we use a default value instead
-              if (!element.closest('main')) return 'Unknown'
+            contentName: element => {
+              // If there is an id, use this as the content name, else return 'Unknown'
+              return element.dataset.id ? element.dataset.id : 'Unknown'
```
nathanrussell-moj-digital, line 14: The new linter did not like the fact that the function could (in theory) not have a return value (in practice it was never a problem)

assets/js/govukFrontendInit.js
nathanrussell-moj-digital, line 1: This is now initialised in the main layout.njk template, similar to how it is done in the typescript-template project

assets/js/print.js
```diff
@@ -1,24 +1,24 @@
 /**
  * Expand all the Overview notes details elements on print
  */
-function showDetailsElementOnPrint() {
+const showDetailsElementOnPrint = () => {
```
nathanrussell-moj-digital, line 4: More changes to make the linter happy 😁

assets/js/govukFrontendInit.js
brightonsbox, line 1: Just wondered why you've got rid of this, but not https://github.com/ministryofjustice/hmpps-education-and-work-plan-ui/blob/update-dependencies-from-template/assets/js/mojFrontendInit.js as was done in the template?

nathanrussell-moj-digital: Good shout, I'd not noticed mojFrontendInit had been removed, I'll do it now 👍 

(the only reason I knew to remove govukFrontendInit was because the linter was moaning about something that I could not work out how to fix, so I looked at what the template did 🤷‍♂️ )

brightonsbox (approved):
LGTM

PR #1011, opened 2025-01-28
RR-1228 - Refactor to get DPS header & footer from DPS library component

nathanrussell-moj-digital:
This PR is massive (sorry! 😬 ), but is [based on the blueprint PR](https://github.com/ministryofjustice/hmpps-welcome-people-into-prison-ui/pull/491) we were given in order to do this work.

This PR removes all of the code that we had in respect of calling the `front-end-components` API to get the header and footer, including managing error/fallback conditions, and replaces it all with a call to a new DPS library dependancy (`@ministryofjustice/hmpps-connect-dps-components`) that does it all for us 👍 
 
One of the reasons that this PR is so big is because we need to change pretty much every cypress test. Previously in our cypress tests we setup the stubs to get the header & footer. This is no longer necessary (as it's all managed by the library), so we can remove all of the calls to the stubs, and the stubs themselves.

---

The reason we are doing this is to have a simpler header and footer that does not contain the dynamic links to other DPS services (so that it will work better on the Education networks).
I'm not convinced it does actually do that!, as the resultant header and footer for me look identical to previous, complete with links to other DPS services!
Having said that:
* It is exactly what we've been asked to do
* I believe it is an improvement over our previous code, and massively simplifies our codebase (by having all of that logic and processing that we previously had containing in a library)


stevomcallister (approved):
Looks lovely have a tick.

PR #1013, opened 2025-01-28
RR-1184 - Correct the status on the update goal review page

nathanrussell-moj-digital:
PR to fix the bug re: the Goal status field on the Update Goal review page

### Screenshot showing bug
Screenshot 2025-01-28 at 15 21 02

### Screenshot showing fix
Screenshot 2025-01-28 at 15 17 25

### DIFF
--- a/server/@types/dto/index.d.ts
+++ b/server/@types/dto/index.d.ts
@@ -4,6 +4,8 @@ declare module 'dto' {
   import QualificationLevelValue from '../../enums/qualificationLevelValue'
   import SessionCompletedByValue from '../../enums/sessionCompletedByValue'
   import ReviewPlanExemptionReasonValue from '../../enums/reviewPlanExemptionReasonValue'
+  import GoalStatusValue from '../../enums/goalStatusValue'
+  import StepStatusValue from '../../enums/stepStatusValue'
 
   export interface CreateActionPlanDto {
     prisonNumber: string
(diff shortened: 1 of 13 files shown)


server/@types/dto/index.d.ts
```diff
@@ -28,14 +30,15 @@ declare module 'dto' {
     goalReference: string
     title: string
     steps: Array<UpdateStepDto>
+    status: GoalStatusValue
```
nathanrussell-moj-digital, line 33: The core problem is that the DTO and form objects for updating a goal did not previously include the goal status, hence it could not be displayed onscreen.

The fix is to add the fields to the DTO and form object. In doing so I have also taken the opportunity to use an enum for the statuses for Goal and Step, and ensure they are populated via the mappers etc.

server/routes/updateGoal/mappers/goalToUpdateGoalFormMapper.ts
```diff
@@ -22,12 +25,12 @@ const toUpdateStepForm = (step: Step): UpdateStepForm => {
...
     title: step.title,
     stepNumber: step.sequenceNumber,
-    status: step.status,
+    status: StepStatusValue[step.status as keyof typeof StepStatusValue],
   }
 }
 
 const toDateString = (date: Date): string => {
-  return moment(date).format('YYYY-MM-DD')
+  return format(date, 'yyyy-MM-dd')
```
nathanrussell-moj-digital, line 33: As well as the mappers mapping the status field, they also map dates. 
Previously we used the `moment` date library, but we have a ticket on the backlog to migrate from `moment` to `date-fns`. Given that I was making a change to these mappers anyway I took the opportunity to use `date-fns` date functions here 👍

stevomcallister (approved):

PR #1019, opened 2025-02-04
RR-1240 - Initial refactoring to better support RBAC

nathanrussell-moj-digital:
PR to introduce better support for RBAC

### Background
At the moment we only have one role `ROLE_EDUCATION_WORK_PLAN_EDITOR`, and we have 2 bits of code (a middleware `checkUserHasEditAuthority`, and a view model property `hasEditAuthority`) that explicitly check for the role. Basically you either have the role, or you don't.

This approach has worked up until now, but it will need some love in order to support the new roles.

#### Option 1
We could create new equivalents of the above - eg: `checkUserHasManagerAuthority`, `checkUserHasContributorAuthority`, `hasManagerAuthority` and `hasContributorAuthority` - and then use these in all the route definitions and view logic as appropriate.
But it might get messy. For example in the view layer we have logic around the Add Goals button:
```
{% if hasEditAuthority %}
  ... markup for Add Goals button ...
{% endif %}
```
We would need to change it to 
```
{% if hasEditAuthority or hasManagerAuthority or hasContributorAuthority %}
  ... markup for Add Goals button ...
{% endif %}
```
and we would have to do this type of thing everywhere where we currently check permissions ... 🤔 

#### Option 2
A different approach would be to have the concept of "actions" that the user could perform (create goals, update goals, create induction etc), and a smaller set of functions that can be used to ask if the user has permission to do a thing. The above example could become:
```
{% if userHasPermissionTo('CREATE_GOALS') %}
  ... markup for Add Goals button ...
{% endif %}
```
The resultant code is easier to read and easier to maintain. It's also one place to maintain the roles for a given action, rather than having to hunt down every place where we allow someone to add goals to change the logic surrounding it.

### This PR
This PR introduces the basic code for Option 2, and uses it for the RBAC controls to do with exempting (and clearing) Inductions, and performing Reviews and exempting (and clearing) them.
This allows us to make sure that it works and we are happy with the approach without risking production functionality (because Reviews and Induction exemptions etc are not currently enabled in prod)
If we are happy with this approach and testing proves no regressions, then subsequent PRs will refactor the RBAC controls protecting other pages and functionality to use these new methods.


server/@types/viewComponents/index.d.ts
```diff
@@ -26,6 +26,7 @@ declare module 'viewComponents' {
     inductionSchedule: InductionScheduleView
     actionPlanReview: ActionPlanReviewScheduleView
     hasEditAuthority: boolean
+    userHasPermissionTo: () => boolean
```
nathanrussell-moj-digital, line 29: This is the new view model property - `userHasPermissionTo` is a function that returns a boolean.
It's use can be seen in a couple of places below

server/enums/applicationAction.ts
```diff
@@ -0,0 +1,13 @@
+enum ApplicationAction {
```
nathanrussell-moj-digital, line 1: An enum describing actions that a user can do in the application.
I _think_ this is most of them, but should be enough to get us going / serve as an example; and if we discover other actions that we need to protect with RBAC then this is where we will define them

server/enums/dpsRole.ts
```diff
@@ -0,0 +1,7 @@
+enum DpsRole {
```
nathanrussell-moj-digital, line 1: This enum previously existed in `roleBasedAccessControl` below 👇 
I've moved it here into the `enum` folder as that is a better / more consistent place for it (rather than it being buried in another file)

server/enums/applicationRole.ts
```diff
@@ -0,0 +1,8 @@
+enum ApplicationRole {
```
nathanrussell-moj-digital, line 1: This enum previously existed in `roleBasedAccessControl` below 👇 
I've moved it here into the `enum` folder as that is a better / more consistent place for it (rather than it being buried in another file)

I've also added the 2 new roles that we will be creating

server/middleware/checkWhetherToShowServiceOnboardingBanner.ts
```diff
@@ -15,7 +15,7 @@ const checkWhetherToShowServiceOnboardingBanner = async (req: Request, res: Resp
   )
 
   res.locals.showServiceOnboardingBanner =
-    !userRoles.includes(ApplicationRoles.ROLE_EDUCATION_WORK_PLAN_EDITOR) &&
+    !userRoles.includes(ApplicationRole.ROLE_EDUCATION_WORK_PLAN_EDITOR) &&
```
nathanrussell-moj-digital, line 18: This will need to change with the move away from this role name, but it is not the subject of this PR

server/middleware/populateUserAuthorities.ts
```diff
@@ -1,13 +1,18 @@
...
+import { userHasPermissionTo } from './roleBasedAccessControl'
 
 export default function populateUserAuthorities(): RequestHandler {
   return async (req, res, next) => {
     if (res.locals?.user?.token) {
       const { authorities: roles = [] } = jwtDecode(res.locals.user.token) as { authorities?: string[] }
 
-      res.locals.hasEditAuthority = roles.includes(ApplicationRoles.ROLE_EDUCATION_WORK_PLAN_EDITOR)
+      // hasEditAuthority is deprecated - use userHasPermissionTo instead
+      res.locals.hasEditAuthority = roles.includes(ApplicationRole.ROLE_EDUCATION_WORK_PLAN_EDITOR)
```
nathanrussell-moj-digital, line 13: `hasEditAuthority` will be nuked once we've migrated everything over to the new `userHasPermissionTo` function
But at the moment it is used in a lot of places, so we need to keep it for now

server/middleware/roleBasedAccessControl.ts
```diff
@@ -1,15 +1,76 @@
...
+import ApplicationRole from '../enums/applicationRole'
+import ApplicationAction from '../enums/applicationAction'
 
-enum ApplicationRoles {
-  ROLE_EDUCATION_WORK_PLAN_EDITOR = 'ROLE_EDUCATION_WORK_PLAN_EDITOR',
-}
+/**
+ * Deprecated - use `checkUserHasPermissionTo` instead
+ */
+const checkUserHasEditAuthority = () => authorisationMiddleware([ApplicationRole.ROLE_EDUCATION_WORK_PLAN_EDITOR])
```
nathanrussell-moj-digital, line 8: `checkUserHasEditAuthority` will be nuked once we've migrated everything over to the new `checkUserHasPermissionTo` function
But at the moment it is used in a lot of places, so we need to keep it for now

server/middleware/roleBasedAccessControl.ts
```diff
@@ -1,15 +1,76 @@
...
-enum DpsRoles {
-  ROLE_GLOBAL_SEARCH = 'ROLE_GLOBAL_SEARCH',
-  ROLE_INACTIVE_BOOKINGS = 'ROLE_INACTIVE_BOOKINGS',
-  ROLE_POM = 'ROLE_POM',
+/**
+ * A map of [ApplicationAction] to [ApplicationRole]s, to determine which role is required for any given action.
+ * The list of [ApplicationRole]s should be considered an "or" list. Users require any one of the listed roles for each
+ * action.
+ */
+const rolesForAction = {
```
nathanrussell-moj-digital, line 15: This is the crux of how it all works ... a map of actions to the roles required to perform that action.

At the moment I've added all 3 of our roles to every action, which will mean stuff will continue to work whilst we get the roles created and migrate users to the new roles.
Once we've done that we can remove `ROLE_EDUCATION_WORK_PLAN_EDITOR`, and set the right role(s) to each action 👍

server/routes/induction/exemption/exemption/index.ts
```diff
@@ -20,30 +22,36 @@ export default (router: Router, services: Services) => {
   const exemptionRecordedController = new ExemptionRecordedController()
 
   router.get('/prisoners/:prisonNumber/induction/exemption', [
+    checkUserHasPermissionTo(ApplicationAction.EXEMPT_INDUCTION),
```
nathanrussell-moj-digital, line 25: Call the new middleware `checkUserHasPermissionTo`
I quite like this - it is clear and expressive, saying "have you got permission to do this thing", rather than "do you have a specific role" 👍

server/routes/induction/exemption/index.ts
```diff
@@ -3,16 +3,15 @@ import createError from 'http-errors'
...
 import config from '../../../config'
-import { checkUserHasEditAuthority } from '../../../middleware/roleBasedAccessControl'
 import exemptInductionRoutes from './exemption'
 import removeInductionExemptionRoutes from './removeExemption'
 
 /**
  * Route definitions for exempting, and removing the exemption of a prisoner's Induction
  */
 export default (router: Router, services: Services) => {
-  router.get('/prisoners/:prisonNumber/induction/exemption', [checkPrisonIsEnabled(), checkUserHasEditAuthority()])
```
nathanrussell-moj-digital, line 14: Remove the old role check

server/views/components/actions-card/_inductionActions.njk
```diff
@@ -40,7 +40,7 @@
     <ul class="govuk-list govuk-!-margin-0" data-qa="induction-action-items">
       {% if params.inductionSchedule.inductionStatus == 'ON_HOLD' %}
 
-        {% if params.hasEditAuthority %}
+        {% if params.userHasPermissionTo('REMOVE_INDUCTION_EXEMPTION') %}
```
nathanrussell-moj-digital, line 43: And this is how we use the new `userHasPermissionTo` view model function.
Remove the previous role based check, replacing it with a clear and expressive check whether you have permission to do a named thing 👍

PR #1028, opened 2025-02-10
Fix async code pattern to not use `.then` callbacks

nathanrussell-moj-digital:
Small PR to make a small fix to a coding style re: async method calls.

When a method returns a `Promise`, there are 2 styles / approaches to call it.
The original way was to use `.then` callback methods. The problem with this approach is that it can create "callback hell", where you can end up chaining and nesting several `.then` calls.

A more modern and generally recommended approach is to use `await`, where the `await` keyword basically says "wait for this promise to be resolved"

This PR refactors a coupe of places where we'd implemented both patterns in the same call! 😬 EG: this type of thing:
```
    const prisoners: Prisoner[] = await this.prisonerSearchClient
      .getPrisonersByPrisonId(prisonId, page, pageSize, systemToken)
      .then(pagedCollectionOfPrisoners => pagedCollectionOfPrisoners.content)
```
where we are using `await` and a `.then` callback !

Unrelated to the above, this PR also corrects a couple of places where we were passing the system token into the service method to get prison names, when in fact we should have been passing in the username.

### DIFF
--- a/server/routes/functionalSkills/functionalSkillsController.ts
+++ b/server/routes/functionalSkills/functionalSkillsController.ts
@@ -1,8 +1,9 @@
 import type { Assessment, FunctionalSkills } from 'viewModels'
-import { Request, RequestHandler } from 'express'
+import { RequestHandler } from 'express'
 import FunctionalSkillsView from './functionalSkillsView'
 import PrisonService from '../../services/prisonService'
 import { functionalSkillsByType } from '../functionalSkillsResolver'
+import logger from '../../../logger'
 
 export default class FunctionalSkillsController {
   constructor(private readonly prisonService: PrisonService) {}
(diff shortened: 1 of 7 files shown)


server/routes/functionalSkills/functionalSkillsController.ts
```diff
@@ -34,14 +35,25 @@ export default class FunctionalSkillsController {
...
-        return {
-          ...assessment,
-          prisonName: allPrisonNamesById.get(assessment.prisonId),
-        }
-      })
+  private setPrisonNamesOnAssessments = async (
+    assessments: Array<Assessment>,
+    username: string,
+  ): Promise<Array<Assessment>> => {
+    const allPrisonNamesById = await this.getAllPrisonNamesByIdSafely(username)
```
nathanrussell-moj-digital, line 42: Refactored to not use `.then`

stevomcallister (approved):

PR #1043, opened 2025-02-13
RR-1024 - Added the sortable column headings to the 3 session list pages

nathanrussell-moj-digital:
This PR carries on with the iterative development of the new Session List pages (due/overdue/on hold)

This PR adds the basic table heading (with `xx` as a placeholder for the number of records) ; and adds the basic table with sortable column headings (different for each page)

https://github.com/user-attachments/assets/8a6352ef-64e9-4eaa-beb2-0c98c7640a9c


server/views/components/sortable-table-header/macro.njk
```diff
@@ -0,0 +1,11 @@
+{#  -------------------------------------------------------------
```
nathanrussell-moj-digital, line 1: This (and the next file) are a move and refactor of `sortableTableHeader.njk` from the prisoner list page
I've made it generic so it can be used on both the Prisoner List page, and also the 3 new Session List pages 👍

stevomcallister (approved):

PR #1045, opened 2025-02-14
RR-1024 - Refactor how we call prisoner search

nathanrussell-moj-digital:
This PR refactors how we call prisoner-search-api to get the list of all prisoners in a prison.

The reason we need to do this is so that the new Session List screens (due/overdue/on-hold) can get the list of prisoners to then call our API to get sessions.
At the moment we take this approach on the prisoner-list screen, but its currently a bit wedded/coupled to prisoner-list , hence the refactoring so that we can call it in other places.

It looks like a massive PR (and it is in respect of number of files changed 😬 ), but the vast majority of the change is how we call the `aValidPrisonerSummary` test data builder function, because that had to have a minor change to its signature.


server/middleware/checkPrisonerInCaseloadMiddleware.test.ts
```diff
@@ -16,7 +16,7 @@ describe('checkPrisonerInCaseloadMiddleware', () => {
   beforeEach(() => {
     jest.resetAllMocks()
     res.locals.user = hmppsUser()
-    res.locals.prisonerSummary = aValidPrisonerSummary(prisonNumber, prisonersPrisonId)
+    res.locals.prisonerSummary = aValidPrisonerSummary({ prisonNumber, prisonId: prisonersPrisonId })
```
nathanrussell-moj-digital, line 19: The vast majority of changes in this PR are actually how we call `aValidPrisonerSummary`
Because of the nature of the main changes in the PR I needed to change the function sig of `aValidPrisonerSummary`, hence it has meant that everywhere we call it I've needed to make this type of change

server/services/prisonerSearchService.ts
```diff
@@ -14,6 +15,50 @@ export default class PrisonerSearchService {
...
   ) {}
 
+  /**
+   * Returns the [PrisonerSummary]s for the specified prisonId
+   *
+   * The HMPPS `prisoner-search` API is a paged API. This function calls the API starting from page 0 until there are no
+   * more pages remaining. The cumulative array of Prisoner records from all API calls are mapped and
+   * grouped into an array of `PrisonerSummary` within the returned `PrisonerSummaries` object.
+   */
+  async getPrisonersByPrisonId(prisonId: string, username: string): Promise<PrisonerSummaries> {
```
nathanrussell-moj-digital, line 25: This is the main objective of this PR - to create a proper service method in the service class (should have been there before 🙄 ) to call prisoner-search to get all prisoners in a prison, and to also account for the fact that it is a paged API (previously we set a page size of 9999 and assumed it would always come back in one page!)

stevomcallister (approved):

PR #1050, opened 2025-02-17
RR-1024 - pagedPrisonerSummaryPrisonerSession

nathanrussell-moj-digital:
This PR is part of the client side search & filter capability for the session list pages.

Because (at the moment) our search pages need to get data from 2 sources (prisoner list in order to get prisoner summary data, plus our own API for the sessions for those prisoners), **the filtering, sorting and pagination is all done client side** (this will change if we start storing prisoner summary data in our own database/API). But at the moment that's how it has to work 🤷‍♂️ 

This PR introduces the classes and code to support the client side filtering, sorting and pagination (wiring it up into the controller and view will come in my next PR)

The basic idea is that we have a view model class which represents each "row" in the search results. This is the snappily named `PrisonerSummaryPrisonerSession` which is a combination of `PrisonerSummary` and `PrisonerSession` (ie. has all the fields of both types.

We then have a class called `PagedPrisonerSummaryPrisonerSession` which is where all the work happens.
It's constructor takes an array of `PrisonerSummaryPrisonerSession` records (that the controller will provide based on calls to prisoner-search and our own API). This will contain all records, not just those that have been filtered, or the current page etc.
The `PagedPrisonerSummaryPrisonerSession` class then has various methods to allow you to change page, sort and filter, with the associated accompanying logic (eg: filtering will implicitly reset you back to page 1, because the result of the filter might leave you with fewer pages than your current page)

This is how the existing prisoner list page works, and this follows the exact same pattern.

### example screenshot of how the session list page will operate
image

### DIFF
--- a/server/routes/sessionList/pagedPrisonerSummaryPrisonerSession.ts
+++ b/server/routes/sessionList/pagedPrisonerSummaryPrisonerSession.ts
@@ -144,73 +144,45 @@ export default class PagedPrisonerSummaryPrisonerSession {
     (left: PrisonerSummaryPrisonerSession, right: PrisonerSummaryPrisonerSession): number => {
       switch (sortBy) {
         case SortBy.NAME: {
-          if (sortableName(left) === sortableName(right)) {
-            return 0
-          }
-          if (sortableName(left) > sortableName(right)) {
-            return sortOrder === SortOrder.ASCENDING ? 1 : -1
-          }
-          return sortOrder === SortOrder.ASCENDING ? -1 : 1
+          return compare(sortableName(left), sortableName(right), sortOrder)
         }
         case SortBy.LOCATION: {
-          if (left.location === right.location) {
-            return 0
-          }
-          if (left.location > right.location) {
-            return sortOrder === SortOrder.ASCENDING ? 1 : -1
-          }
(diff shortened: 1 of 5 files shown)


server/routes/sessionList/pagedPrisonerSummaryPrisonerSession.ts
```diff
@@ -0,0 +1,259 @@
...
+  /**
+   * Function that returns a [PrisonerSummaryPrisonerSession] comparator function that operates on the specified sort field
+   * and sort order.
+   */
+  private prisonerSearchSummariesComparator =
+    (sortBy: SortBy, sortOrder: SortOrder) =>
+    (left: PrisonerSummaryPrisonerSession, right: PrisonerSummaryPrisonerSession): number => {
+      switch (sortBy) {
+        case SortBy.NAME: {
+          if (sortableName(left) === sortableName(right)) {
```
stevomcallister, line 147: Looks like each of these case statements does the same thing... Maybe (and I don't know how) a function can be extracted that does the common comparison bits?

stevomcallister (approved):

stevomcallister (approved):

PR #1051, opened 2025-02-17
RR-1024 - Wired pagedPrisonerSummaryPrisonerSession into controller and view

nathanrussell-moj-digital:
PR to wire up the controller and view to be able to sort/filter/paginate and render the `pagedPrisonerSummaryPrisonerSession`

Unit and cypress tests will be in my next PR, as will the translating the exemption reason into a human readable value

https://github.com/user-attachments/assets/b2d8a2a5-9933-46ce-b331-4e0c29ebcd43

### DIFF
--- a/helm_deploy/values-dev.yaml
+++ b/helm_deploy/values-dev.yaml
@@ -25,6 +25,7 @@ generic-service:
     COMPONENT_API_URL: 'https://frontend-components-dev.hmpps.service.justice.gov.uk'
     ACTIVITIES_API_URL: 'https://activities-api-dev.prison.service.justice.gov.uk'
     ENVIRONMENT_NAME: 'DEV'
+    SESSION_LIST_UI_DEFAULT_PAGINATION_PAGE_SIZE: "4"
 
     # Comma delimited list of prison IDs that our service is enabled for in `dev`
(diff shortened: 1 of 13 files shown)


helm_deploy/values-dev.yaml
```diff
@@ -25,6 +25,7 @@ generic-service:
     COMPONENT_API_URL: 'https://frontend-components-dev.hmpps.service.justice.gov.uk'
     ACTIVITIES_API_URL: 'https://activities-api-dev.prison.service.justice.gov.uk'
     ENVIRONMENT_NAME: 'DEV'
+    SESSION_LIST_UI_DEFAULT_PAGINATION_PAGE_SIZE: "4"
```
nathanrussell-moj-digital, line 28: I've deliberately set the pagination size in dev to 4. This is to allow the team to test the pagination controls properly. 
There are only 15 overdue records in Wakefield, so with a pagesize of 50 they'd never be able to test pagination properly. 4 might seem like a weird number, but it is better than 3 or 5 which would result in exact page sizes. With 4 we should expect 3 pages of 4 records, plus a 4th page with 3 records on it.

stevomcallister (approved):

PR #1058, opened 2025-02-21
RR-1265 - Use a dedicated client for curious API calls

nathanrussell-moj-digital:
This PR makes the required changes to use a dedicated client in order to call Curious API.
The client is still managed, issued and validated by HMPPS Auth; it's just that its not the main UI client and only carries the role `ROLE_CURIOUS_API`

The HAAR team have already created the client in dev and setup the relevant secrets in our dev namespace. This PR makes the required changes to our codebase to use it 👍


helm_deploy/values-dev.yaml
```diff
@@ -34,6 +34,12 @@ generic-service:
 
     REVIEWS_ENABLED: "true"
 
+  # TODO - move to parent `values.yaml` when we've requested and the HAAR team have created the client in preprod and prod
```
nathanrussell-moj-digital, line 37: If we had this in `values.yaml` before the client and secrets exist in preprod and prod, if we tried to deploy this build to preprod or prod it would not start up because the secrets do not exist yet

server/services/curiousService.ts
```diff
@@ -18,11 +18,11 @@ export default class CuriousService {
   ) {}
 
   async getPrisonerSupportNeeds(prisonNumber: string, username: string): Promise<PrisonerSupportNeeds> {
-    const systemToken = await this.hmppsAuthClient.getSystemClientToken(username)
     const prisonNamesById = await this.prisonService.getAllPrisonNamesById(username)
+    const curiousClientToken = await this.hmppsAuthClient.getCuriousClientToken()
```
nathanrussell-moj-digital, line 22: This is the key change in respect of the Curious Service class. Instead of getting the UI's system token via `hmppsAuthClient.getSystemClientToken`, we get the Curious token via the new method `this.hmppsAuthClient.getCuriousClientToken`

server/data/hmppsAuthClient.ts
```diff
@@ -73,4 +93,20 @@ export default class HmppsAuthClient {
 
     return newToken.body.access_token
   }
+
+  async getCuriousClientToken(): Promise<string> {
```
nathanrussell-moj-digital, line 97: Method to get the Curious client token from HMPP Auth.
It works in exactly the same way as the `getSystemClientToken` method, in that it caches it in the token store, only requesting HMPPS Auth to issue it when we need it (ie. when it's not in the token store / already expired)

stevomcallister (approved):

PR #1075, opened 2025-03-03
RR-1279 - Removed prisoner name from page titles

nathanrussell-moj-digital:


server/views/pages/induction/additionalTraining/index.njk
```diff
@@ -1,8 +1,7 @@
 {% extends "../../../partials/layout.njk" %}
 
 {% set pageId = "induction-additional-training" %}
-{% set title = "Does " + prisonerSummary.firstName + " " + prisonerSummary.lastName + " have any other training or vocational qualifications?" %}
-{% set pageTitle = title %}
+{% set pageTitle = "Do they have any other training or vocational qualifications?" %}
```
nathanrussell-moj-digital, line 4: All the changes in this PR are basically the same thing.
In a handful of nunjucks templates we were previously setting a variable `title` that included the prisoner's name; and we were using it for the main `<h1>` of the page (usually within one of the govuk form macros as in this case); but also within the html `<title>` element (via the `pageTitle`) variable.

We were told a while back not to do this type of thing (but clearly these pages slipped through the net! 🙄 ) because the page `<title>` element is logged externally in 3rd party storage (typically Google Analytics which we dont use, but also App Insights which we do use)
We were told that logging prisoner names externally like this was not OK (it's essentially PII 🤷‍♂️ ), but obviously it is OK to use prisoner names etc within the markup of the rendered page.

This is what is currently logged to App Insights for example:
image

The changes being made in all of these templates basically keeps the `<h1>` etc with the same content, but makes the `<title>` element more generic and not referencing a specific person by using terms such as "they"

PR #1085, opened 2025-03-10
RR-1296 - Use the latter of ACTION_PLAN_CREATED event or INDUCTION_CREATED event

nathanrussell-moj-digital:
This PR updates the rendering of the timeline in respect of the ACTION_PLAN_CREATED and INDUCTION_CREATED events

Basically the API records and serves to the UI both an `ACTION_PLAN_CREATED` and `INDUCTION_CREATED` event. They are kinda the same, but subtly different!
* The `INDUCTION_CREATED` event is created when the Induction (question set) is created
* The `ACTION_PLAN_CREATED` event is created when the first goal is created

Clearly there are a number of permutation of what a prisoner's timeline data might contain:
* No `INDUCTION_CREATED` or `ACTION_PLAN_CREATED` event if the induction and goals have not yet been created
* The `INDUCTION_CREATED` event, but no `ACTION_PLAN_CREATED` when the Induction has been created, but no goals yet
  * In this case the "Action Plan" is not considered created yet
* No `INDUCTION_CREATED` event, but the `ACTION_PLAN_CREATED` event.
  * This is the case where Goals are created before the Induction. There will be cases where we only have a `ACTION_PLAN_CREATED` event.
* Both the `INDUCTION_CREATED` and `ACTION_PLAN_CREATED` events
  * In the case where we have both events, we will have cases where the `INDUCTION_CREATED` event was created before the `ACTION_PLAN_CREATED` event; and cases where the `ACTION_PLAN_CREATED` was created first.

The UI needs to render a `ACTION_PLAN_CREATED` event in the UI (which renders as "Learning and work progress plan created"), but only if both `INDUCTION_CREATED` and `ACTION_PLAN_CREATED` events exist.
* If the `INDUCTION_CREATED` event is created before the `ACTION_PLAN_CREATED` event (our main use case), then we need to use the dates etc from the `ACTION_PLAN_CREATED` event but with the contextual info from the `INDUCTION_CREATED` event
* If the `ACTION_PLAN_CREATED` event is created before the `INDUCTION_CREATED` event (ie. Goals were created before the Induction), then we need to use the `INDUCTION_CREATED` event (because it has the correct dates and contextual info), but change it's type to `ACTION_PLAN_CREATED`


server/routes/overview/prepareTimelineForView.ts
```diff
@@ -134,4 +135,68 @@ const modifyActionPlanCreatedTimestamp = (timelineEvents: Array<TimelineEvent>):
...
+// This function corrects these events:
+//   * If BOTH the INDUCTION_CREATED and ACTION_PLAN_CREATED events exist:
+//     * If the INDUCTION_CREATED event is after the ACTION_PLAN_CREATED event, use the induction created event (but with an
+//       event type of ACTION_PLAN_CREATED)
+//     * If the ACTION_PLAN_CREATED event is after or the same time as the INDUCTION_CREATED, use the ACTION_PLAN_CREATED
+//       event but with the contextual info from the INDUCTION_CREATED event
+//  * If both events are not present, return the timeline with those event types removed
+//    IE. in case there was one but not the other, remove them so there are neither and the timeline is unabiguous in this
+//    regard.
+const updateActionPlanCreatedEventWithLatestOfActionPlanCreatedOrInductionCreatedEvents = (
```
nathanrussell-moj-digital, line 159: This method is the key change that fixes the "Learning and work plan created" timeline event 👍

stevomcallister: holy long const Batman!

stevomcallister (approved):

server/routes/overview/prepareTimelineForView.ts
nathanrussell-moj-digital: It does what it says in the tin 😁 
(and the arguably overly verbose method comment 😁 🤷‍♂️ )

PR #1088, opened 2025-03-11
RR-849 - Refactored back links on In Prison Work and In Prison Training pages

nathanrussell-moj-digital:
This PR refactors the "back" links on the Induction "In Prison Work" and "In Prison Training" pages:
Screenshot 2025-03-12 at 08 23 23

### Background
The designs for the Induction screens from our designers have a "< Back" link at the top of each form page. (GDS guidelines say that form/wizard style pages should have a "back" link to allow users to go back and correct data on previous screens)

To produce a true browser back navigation link that goes back to the previous page in the browser history you need to use the javascript `history` API. And making elements of the page dependant on JS is potentially a bit of a GDS no-no because you cannot guarantee that users have JS enabled .... 🙄  (yes, I know it's not a public website, and that our users are known users with known and managed PC and browser configurations .... but them's the rules !)

To get round this, rather than produce a JS back link, we implemented a solution by where the node application manages state in respect of the page you were on previously, and then produces a regular anchor link that is labelled "< Back" and that goes to the page URL that the application knows you were on previously.

It's fiendishly complicated because each page you can get to via 3 routes (the page immediately preceding it in the wizard style journey; or the Check Your Answers page; or the main Overview page (in the case of editing a previously saved Induction)). And some pages where there is conditional routing in the wizard flow (ie. questions such as Have you worked before?" whose answer determines the next page are even more complicated!

Over and above that, as a solution it's fundamentally flawed (IMHO) and I've never been happy with it. Semantically it's wrong because the link is labelled "< Back" and the user is expecting to go (back?) to the previous page (which they do), though they don't go "back" anywhere. Because it's a regular anchor link they are actually going forwards in the browser history! This causes a real problem and confusion when the user uses the browser Back button (because that really does go "back"), but they end up on a page that is genuinely their previous page (as per the history API), but is not always the page they expect!

### Solution
We agreed a while back to change these "< Back" links to JS links, despite the concerns over whether users would have JS enabled or not.
We decided the risk was low because the users and their browser configurations are a known quantity.
We also decided to implement these "< Back" links using a progressive enhancement technique, which basically means hiding the link via CSS, and then using a CSS rule that displays it only if the browser has JS enabled. By doing this we would only be presenting the "< Back" link if the user had JS enabled; and if they didnt (unlikely) the link would not be displayed. And given that back navigation is not key functionality (you can complete the Induction without needing to use the Back button), and that the browser Back button would still work anyway, this felt like an acceptable and pragmatic solution.

This solution was agreed with the team a while ago and it has been implemented in other form/wizard based screens in our service already - we have already proven it 👍 . So now we are at the stage of implementing it in the Induction journey, hence these PRs.

### Changes
The Induction journey is made up of many screens, and implementing this change for all screens in the journey would result is a massive PR! For each page the change typically touches the nunjucks template, the page controller superclass, the page controller for Create, the page controller for Update, the unit tests for both Create and Update, the page view model class, and a couple of cypress tests! It's a lot of change!
So I've opted to implement this change over several PRs that do it page by page (or related groups of pages)

This PR implements these changes for the In Prison Work Interests and In Prison Training pages.

### DIFF
--- a/integration_tests/e2e/induction/createCheckYourAnswersChangeLinks.cy.ts
+++ b/integration_tests/e2e/induction/createCheckYourAnswersChangeLinks.cy.ts
@@ -48,7 +48,6 @@ context(`Change links on the Check Your Answers page when creating an Induction`
     // Change in-prison training interests
     Page.verifyOnPage(CheckYourAnswersPage)
       .clickInPrisonTrainingInterestsChangeLink()
-      .hasBackLinkTo(`/prisoners/${prisonNumber}/create-induction/check-your-answers`)
       .deSelectInPrisonTraining(InPrisonTrainingValue.FORKLIFT_DRIVING)
       .selectInPrisonTraining(InPrisonTrainingValue.BARBERING_AND_HAIRDRESSING)
       .selectInPrisonTraining(InPrisonTrainingValue.RUNNING_A_BUSINESS)
@@ -57,7 +56,6 @@ context(`Change links on the Check Your Answers page when creating an Induction`
     // Change in-prison work interests
     Page.verifyOnPage(CheckYourAnswersPage)
       .clickInPrisonWorkInterestsChangeLink()
-      .hasBackLinkTo(`/prisoners/${prisonNumber}/create-induction/check-your-answers`)
       .deSelectWorkType(InPrisonWorkValue.PRISON_LIBRARY)
       .selectWorkType(InPrisonWorkValue.MAINTENANCE)
       .selectWorkType(InPrisonWorkValue.TEXTILES_AND_SEWING)
--- a/server/routes/induction/common/checkYourAnswersController.ts
+++ b/server/routes/induction/common/checkYourAnswersController.ts
@@ -7,6 +7,16 @@ import { buildNewPageFlowHistory } from '../../pageFlowHistory'
  * Abstract controller class defining functionality common to both the Create and Update Induction journeys.
  */
 export default abstract class CheckYourAnswersController extends InductionController {
+  getBackLinkUrl(_req: Request): string {
+    // Default implementation - the back link is not displayed on the Check Your Answers page
+    return undefined
+  }
+
+  getBackLinkAriaText(_req: Request): string {
+    // Default implementation - the back link is not displayed on the Check Your Answers page
+    return undefined
+  }
+
   /**
    * Returns the Check Your Answers view; suitable for use by the Create and Update journeys.
    */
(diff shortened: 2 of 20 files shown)


integration_tests/e2e/induction/createCheckYourAnswersChangeLinks.cy.ts
```diff
@@ -48,7 +48,6 @@ context(`Change links on the Check Your Answers page when creating an Induction`
     // Change in-prison training interests
     Page.verifyOnPage(CheckYourAnswersPage)
       .clickInPrisonTrainingInterestsChangeLink()
-      .hasBackLinkTo(`/prisoners/${prisonNumber}/create-induction/check-your-answers`)
```
nathanrussell-moj-digital, line 51: Because the back link is now a JS back link (`window.history.back()`) we cannot assert what the back link would be / where it would take you.
(And TBH, these assertions were only there to help prove the implementation of the complex rules re: tracking the previous page etc)

server/routes/induction/common/checkYourAnswersController.ts
```diff
@@ -7,6 +7,16 @@ import { buildNewPageFlowHistory } from '../../pageFlowHistory'
  * Abstract controller class defining functionality common to both the Create and Update Induction journeys.
  */
 export default abstract class CheckYourAnswersController extends InductionController {
+  getBackLinkUrl(_req: Request): string {
```
nathanrussell-moj-digital, line 10: The Induction page superclass defines the abstract methods `getBackLinkUrl` and `getBackLinkAriaText` , and all implementations needed to implement these methods (which was then in turn used in the nunjucks templates to render the "back" links)

Anyway, once we have migrated all Induction pages to use JS back links, these methods can be nuked from the superclass and implementations 👍  (I'm looking forward to that PR 😁 )

server/routes/induction/common/inPrisonTrainingController.ts
```diff
@@ -18,19 +28,10 @@ export default abstract class InPrisonTrainingController extends InductionContro
 
     this.addCurrentPageToFlowHistoryWhenComingFromCheckYourAnswers(req)
 
-    if (req.session.pageFlowHistory) {
```
nathanrussell-moj-digital, line 21: The "pageFlowHistory" is part of the complex implementation of how the Induction journey knows what the previous page is.
Again, when all Induction pages have been migrated to the JS back link solution this is another thing I'm looking forward to nuking completely 😁

server/routes/induction/create/inPrisonTrainingCreateController.test.ts
```diff
@@ -97,47 +93,6 @@ describe('inPrisonTrainingCreateController', () => {
       expect(req.session.inPrisonTrainingForm).toBeUndefined()
       expect(req.session.inductionDto).toEqual(inductionDto)
     })
-
-    it('should get the In Prison Training view given the previous page was Check Your Answers', async () => {
```
nathanrussell-moj-digital, line 101: We can no longer do this type of test because the controller no longer manages state of what the previous page was

PR #1108, opened 2025-03-19
RR-1031 - Add filtering options on API client

nathanrussell-moj-digital:
This PR adds support to allow the client method that calls the Timeline API to be able to pass the relevant query string params to apply the filtering.
At the moment the params are all sent as `false`, with the filtering still being applied post-response. Subsequent PRs will start to wire this all up 👍

### DIFF
--- a/integration_tests/mockApis/educationAndWorkPlanApi.ts
+++ b/integration_tests/mockApis/educationAndWorkPlanApi.ts
@@ -245,7 +245,7 @@ const stubGetTimeline = (prisonNumber = 'G6115VJ'): SuperAgentRequest =>
   stubFor({
     request: {
       method: 'GET',
-      urlPattern: `/timelines/${prisonNumber}`,
+      urlPathPattern: `/timelines/${prisonNumber}`,
     },
     response: {
       status: 200,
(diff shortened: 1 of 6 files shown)


integration_tests/mockApis/educationAndWorkPlanApi.ts
```diff
@@ -245,7 +245,7 @@ const stubGetTimeline = (prisonNumber = 'G6115VJ'): SuperAgentRequest =>
   stubFor({
     request: {
       method: 'GET',
-      urlPattern: `/timelines/${prisonNumber}`,
+      urlPathPattern: `/timelines/${prisonNumber}`,
```
nathanrussell-moj-digital, line 248: Needed to make this change to the wiremock stubs, because now the API is called with the query string params applied (all as false at the moment) - but it means as far as wiremock matching is concerned, it is a different URL so we need to use `urlPathPattern` rather than `urlPattern`

server/data/timelineApiFilterOptions.ts
```diff
@@ -0,0 +1,30 @@
+export default class TimelineApiFilterOptions {
```
nathanrussell-moj-digital, line 1: This is the crux of how this is all going to work.
This new class `TimelineApiFilterOptions` is now a method arg to the API method.
The constructor is an optional map of optional flags; and it creates a property `queryParams` that has all the query string options with `false` defaults

The service method will new one of these up with the relevant filtering options and pass it into the API method; and the API method will use it for the `query` object.

In this PR the service method news up a `TimelineApiFilterOptions` but with no constructor arg/filter options, so the resultant filtering when we call the API is all the filter options set to `false` 👍

stevomcallister (approved):

PR #1132, opened 2025-04-03
RR-1372 - Add journeyId to create induction routes

nathanrussell-moj-digital:
This PR adds a journey ID to the path params of the "create induction" routes.
The reason we want to do this is because (in a future PR) we will use the journey ID as a key to store the Induction data in Redis, rather than holding it on the session as at present.

The pattern adopted is the same as several other DPS services including CSIP, BVLS and others.
The basic idea is that when the user starts the journey (ie. the first page of the Induction flow in our case), a unique ID (journey ID) is generated and inserted into the URL, followed by a redirect to that new URL.
All route patterns of the Induction flow have been changed to include the journey ID
All controllers in the Induction flow (and their unit tests) have been changed to use the journey ID from "this" page when constructing the URL for the users "next" page in the flow.

All this PR does is introduce the journey ID and pass it from page to page. At the moment it is not used for storing the Induction data in Redis (that will be a future PR)

### DIFF
--- a/server/middleware/auditMiddleware.ts
+++ b/server/middleware/auditMiddleware.ts
@@ -58,31 +58,37 @@ const pageViewEventMap: Record<string, Page> = {
   '/plan/:prisonNumber/in-prison-courses-and-qualifications': Page.IN_PRISON_COURSES_AND_QUALIFICATIONS,
 
   // Create induction
-  '/prisoners/:prisonNumber/create-induction/hoping-to-work-on-release':
+  '/prisoners/:prisonNumber/create-induction/hoping-to-work-on-release': null, // route without the journeyId does not raise an audit event because it redirects to the route with a journeyId
+  '/prisoners/:prisonNumber/create-induction/:journeyId/hoping-to-work-on-release':
     Page.INDUCTION_CREATE_HOPING_TO_WORK_ON_RELEASE,
-  '/prisoners/:prisonNumber/create-induction/want-to-add-qualifications': Page.INDUCTION_CREATE_ADD_QUALIFICATION,
-  '/prisoners/:prisonNumber/create-induction/qualifications': Page.INDUCTION_CREATE_QUALIFICATIONS,
-  '/prisoners/:prisonNumber/create-induction/highest-level-of-education':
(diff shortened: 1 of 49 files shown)


server/middleware/auditMiddleware.ts
```diff
@@ -58,31 +58,37 @@ const pageViewEventMap: Record<string, Page> = {
   '/plan/:prisonNumber/in-prison-courses-and-qualifications': Page.IN_PRISON_COURSES_AND_QUALIFICATIONS,
 
   // Create induction
-  '/prisoners/:prisonNumber/create-induction/hoping-to-work-on-release':
+  '/prisoners/:prisonNumber/create-induction/hoping-to-work-on-release': null, // route without the journeyId does not raise an audit event because it redirects to the route with a journeyId
+  '/prisoners/:prisonNumber/create-induction/:journeyId/hoping-to-work-on-release':
```
nathanrussell-moj-digital, line 62: The journey ID path param has been added to all these pages that are audited

server/routes/induction/create/index.ts
```diff
@@ -61,163 +62,159 @@ export default (router: Router, services: Services) => {
...
-  router.post('/prisoners/:prisonNumber/create-induction/**', [
+  router.use('/prisoners/:prisonNumber/create-induction/**', [
     checkUserHasPermissionTo(ApplicationAction.RECORD_INDUCTION),
     createEmptyInductionIfNotInSession(educationAndWorkPlanService),
     setCurrentPageInPageFlowQueue,
+    insertJourneyIdentifier({ insertIdAfterElement: 3 }), // insert journey ID immediately after '/prisoners/:prisonNumber/create-induction' - eg: '/prisoners/A1234BC/create-induction/473e9ee4-37d6-4afb-92a2-5729b10cc60f/hoping-to-work-on-release'
   ])
 
-  router.get('/prisoners/:prisonNumber/create-induction/hoping-to-work-on-release', [
+  router.get('/prisoners/:prisonNumber/create-induction/:journeyId/hoping-to-work-on-release', [
```
nathanrussell-moj-digital, line 72: And the other part of the magic is to change all the Induction routes to include the journey ID as a path parameter

PR #1139, opened 2025-04-09
RR-1402 - remove Review Actions when there is no scheduled review

nathanrussell-moj-digital:
PR to fix the problem where users can conduct reviews and review exemptions when the prisoner has no scheduled review.

If the prisoner has no scheduled review, we now no longer display the links to perform these actions:
Screenshot 2025-04-09 at 14 36 46


stevomcallister (approved):
A fine solution

PR #1147, opened 2025-04-16
RR-1413 - Fix "clear filters" link for all user types

nathanrussell-moj-digital:
PR to fix the href of the "clear filters" link on the prisoner-list page so that it works for all user types

### DIFF
--- a/integration_tests/e2e/prisonerList/prisonerList.cy.ts
+++ b/integration_tests/e2e/prisonerList/prisonerList.cy.ts
@@ -48,8 +48,6 @@ context(`Display the prisoner list screen`, () => {
       // Given
       if (authority === 'VIEW') {
         cy.task('stubSignInAsReadOnlyUser')
-      } else if (authority === 'EDIT') {
-        cy.task('stubSignInAsUserWithManagerRole')
       } else if (authority === 'CONTRIBUTOR') {
         cy.task('stubSignInAsUserWithContributorRole')
       }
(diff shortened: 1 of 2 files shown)


integration_tests/e2e/prisonerList/prisonerList.cy.ts
```diff
@@ -136,28 +134,38 @@ context(`Display the prisoner list screen`, () => {
...
-
-      // When
-      prisonerListPage.clearFilters()
-
-      // Then
-      prisonerListPage //
-        .hasResultsDisplayed(expectedResultCount)
-        .hasNoSearchTerm()
-        .hasNoStatusFilter()
+    ;['VIEW', 'CONTRIBUTOR', 'MANAGER'].forEach(authority => {
```
nathanrussell-moj-digital, line 137: This test is a rework of the previous version, but iterates through 3 different user types (read only users, contributors, and managers)
For each of them the test gets to the prisoner-list page, applies a filter, and then clicks the "Clear" filter link.
The expectation is that for all user types they all land back on the prisoner-list page with the filters removed and a full set of results displayed.

server/views/pages/prisonerList/partials/searchBox.njk
```diff
@@ -39,8 +39,11 @@
           {# Reset the Prisoner List page filters using an anchor link with blank searchTerm and statusFilter field values.
               The link also contains the current sort order to ensure the current sort field and order are maintained after
               reseting the filter.
+              The link url is "/search" which works for all user types. A link url of "/" only works for readonly and users
+              with the Contributor role because their default landing page ("/") is this prisoner-list page. Users with
+              the Manager role have a different landing page ("/"), so using "/search" will work for all user types.
           #}
-          <a class="govuk-link govuk-link--no-visited-state" href="/?searchTerm=&statusFilter=&sort={{ sortBy }},{{ sortOrder }}" data-qa="clear-filters">
+          <a class="govuk-link govuk-link--no-visited-state" href="/search?searchTerm=&statusFilter=&sort={{ sortBy }},{{ sortOrder }}" data-qa="clear-filters">
```
nathanrussell-moj-digital, line 46: This is the fix. Previously we linked to "/". IE. the root/default page
This worked for read only users and users with the Contributor role, because their default landing page ("/") is this prisoner-list page.
But for users with the Manager role, their default landing page is a different page, so when they clicked the "Clear" filters link they were redirected to their landing page, rather than redisplaying this page with the filters cleared.

Fortunately the fix is pretty easy. We have a route "/search" that renders this prisoner-list page that works for all user types 👍

yasinmustafa (approved):
lgtm

PR #1149, opened 2025-04-23
RR-1375 - Introduced journey data handling (type, store and middlewares)

nathanrussell-moj-digital:
This PR introduces the initial code we will need to support the concept of "journey data" being stored in redis rather than the session.

In a previous PR we have already introduced the middleware that adds the journey ID to the path parameter of the Induction journey. The journeyID exists as a path parameter, but we do not use it yet (the `InductionDto` is still held on the session)

This PR introduces:
* The `JourneyData` type, currently comprised of a property to hold the `InductionDto`
* The `journeyData` property on the `Request`
* A redis backed store class to store, get, and remove `JourneyData`
* A middleware to store journeyData via the store class
* A middleware to retrieve journeyData from the store class

This PR introduces all of these things, but they are not wired in or used at the moment. That will come in subsequent PRs (I have a branch that does most of this, but where the Induction flow is quite large it touches ~90 files !!, so I am trying to introduce the change in smaller chunks)

### DIFF
--- a/server/data/journeyDataStore/journeyDataStore.ts
+++ b/server/data/journeyDataStore/journeyDataStore.ts
@@ -1,0 +1,57 @@
+import { isValid, parseISO } from 'date-fns'
+import { RedisClient } from '../redisClient'
+import logger from '../../../logger'
+
+const DATE_FIELDS = ['createdAt', 'updatedAt']
+
+export default class JourneyDataStore {
+  constructor(private readonly client: RedisClient) {
+    client.on('error', error => {
+      logger.error(error, `Redis error`)
+    })
+  }
+
+  private async ensureConnected() {
+    if (!this.client.isOpen) {
+      await this.client.connect()
+    }
+  }
+
+  async setJourneyData(
+    username: string,
+    journeyId: string,
+    journeyData: Express.JourneyData,
+    durationHours = 1,
+  ): Promise<string> {
+    await this.ensureConnected()
+    return this.client.set(`journey.${username}.${journeyId}`, JSON.stringify(journeyData), {
+      EX: durationHours * 60 * 60,
+    })
+  }
+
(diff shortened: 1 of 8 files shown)


server/data/journeyDataStore/journeyDataStore.ts
```diff
@@ -0,0 +1,57 @@
...
+    await this.ensureConnected()
+    return this.client.set(`journey.${username}.${journeyId}`, JSON.stringify(journeyData), {
+      EX: durationHours * 60 * 60,
+    })
+  }
+
+  async getJourneyData(username: string, journeyId: string): Promise<Express.JourneyData> {
+    await this.ensureConnected()
+    const serializedJourneyData = await this.client.get(`journey.${username}.${journeyId}`)
+    return JSON.parse(serializedJourneyData ?? '{}', dataParsingReviver(DATE_FIELDS))
```
nathanrussell-moj-digital, line 35: This is probably the most significant thing of note ... the use of the "reviver" function on `JSON.parse` ....

Basically, unless you give it some help with a reviver function, `JSON.parse` does not instantiate `Date` properties properly (at all!). This means that if your target object contains `Date` properties (which `InductionDto` does), these properties are not instantiated properly, and you end up with an object that contains the property name (eg: `createdAt`), but the value is the ISO string (straight from when it was serialised), rather than a `Date` object.

This is a pretty good write up of the problem and solution:
https://medium.com/@meplafi/date-parsing-problem-in-json-with-json-parses-lesser-known-parameter-3fc6cca0c53b#:~:text=This%20happens%20because%20JSON%20does,date%20object%20automatically%20for%20us.

(not sure if this might be a problem for you @iterative-it in CSIP - not sure if any of your journey objects contain `Date` properties?)

iterative-it (approved):
LGTM

PR #1153, opened 2025-04-25
RR-1436 - Refactor Induction Exemption journey to use journeyData to hold DTO

nathanrussell-moj-digital:
The PR refactors the Induction Exemption journey to use `journeyData` to hold it's DTO (InductionExemptionDto)

Previously the DTO was held in the `prisonerContext` which was an early attempt at simplifying our session usage / storing data. Whilst `prisonerContext` helped with some problems (it is keyed on prisonNumber meaning you can work on 2 different prisoners in different tabs 👍 ), it is still held on the session.

This journey is the first journey to start to use the new journeyData process, so is the first to have wired in the `setupJourneyData` middleware into it's routes 👍


server/routes/induction/exemption/exemption/confirmExemptionController.test.ts
```diff
@@ -28,15 +27,15 @@ describe('ConfirmExemptionController', () => {
...
-      session: {},
+      journeyData: {},
     } as unknown as Request
     res = {
       render: jest.fn(),
       redirect: jest.fn(),
       locals: { prisonerSummary },
     } as unknown as Response
 
-    getPrisonerContext(req.session, prisonNumber).inductionExemptionDto = undefined
```
nathanrussell-moj-digital, line 39: The changes to the controllers and their tests are essentially the same - everywhere where we used the prisonerContext and session to hold/reference the inductionExceptionDto has been changed to use req.journeyData.

server/routes/induction/exemption/exemption/exemptionReasonController.test.ts
```diff
@@ -87,7 +88,7 @@ describe('exemptionReasonController', () => {
       // Then
       expect(res.redirect).toHaveBeenCalledWith(`/prisoners/A1234BC/induction/${journeyId}/exemption/confirm`)
       expect(getPrisonerContext(req.session, prisonNumber).inductionExemptionForm).toBeUndefined()
-      expect(getPrisonerContext(req.session, prisonNumber).inductionExemptionDto).toEqual(inductionExemptionDto)
+      expect(req.journeyData.inductionExemptionDto).toEqual(inductionExemptionDto)
```
nathanrussell-moj-digital, line 91: Worth noting that this PR is only about moving the DTO (the object that is passed from page to page) from prisonerContext/session to the new journeyData.

The individual page forms are still held on the prisonerContext/session - these are less of a problem for what we are trying to solve, but they will get moved/refactored, but will involve some very different refactoring so I wanted to keep them separate.

server/routes/induction/exemption/exemption/index.ts
```diff
@@ -1,56 +1,61 @@
...
+  router.use([
+    //
+    checkUserHasPermissionTo(ApplicationAction.EXEMPT_INDUCTION),
+    setupJourneyData(journeyDataService),
+  ])
 
   router.get('/', [
     checkInductionIsScheduled(inductionService), // Induction Schedule must be SCHEDULED in order to exempt it
-    createEmptyInductionExemptionDtoIfNotInPrisonerContext,
+    createEmptyInductionExemptionDtoIfNotInJourneyData,
```
nathanrussell-moj-digital, line 35: The "create...." and "check...." middlewares have been renamed, hence the changes here (previously they referenced the prisonerContext, now they reference the journeyData)

PR #1156, opened 2025-04-28
RR-1442 - Refactor Update Induction journey to use journeyData to hold DTO

nathanrussell-moj-digital:


server/routes/induction/common/additionalTrainingController.ts
```diff
@@ -17,7 +17,7 @@ export default abstract class AdditionalTrainingController extends InductionCont
     res: Response,
     next: NextFunction,
   ): Promise<void> => {
-    const { inductionDto } = req.session
+    const inductionDto = req.session.inductionDto ?? req.journeyData?.inductionDto
```
nathanrussell-moj-digital, line 20: These controllers in `induction/common` are used for both the Create and Update journey
In this PR we are changing the Update journey to store and reference the DTO in journeyData, but the Create journey still uses the session; hence this logic in each controller.

(soon we will migrate the Create Induction journey to use the journeyData approach, so this code will be simplified again at that point)

server/routes/induction/update/additionalTrainingUpdateController.ts
```diff
@@ -20,7 +20,7 @@ export default class AdditionalTrainingUpdateController extends AdditionalTraini
     next: NextFunction,
   ): Promise<void> => {
     const { prisonNumber, journeyId } = req.params
-    const { inductionDto } = req.session
+    const { inductionDto } = req.journeyData
```
nathanrussell-moj-digital, line 23: The changes to all the Update controllers is basically the same ...  get the `InductionDto` from journeyData instead of the session; and either put it back on journeyData rather than the session, or clear the journeyData (depending on the flow in the existing code)

server/routes/induction/update/index.ts
```diff
@@ -43,9 +44,12 @@ export default (router: Router, services: Services) => {
   const additionalTrainingUpdateController = new AdditionalTrainingUpdateController(inductionService)
 
   router.use('/prisoners/:prisonNumber/induction', [
-    insertJourneyIdentifier({ insertIdAfterElement: 3 }), // insert journey ID immediately after '/prisoners/:prisonNumber/induction' - eg: '/prisoners/A1234BC/induction/473e9ee4-37d6-4afb-92a2-5729b10cc60f/hoping-to-work-on-release'
     checkUserHasPermissionTo(ApplicationAction.UPDATE_INDUCTION),
-    retrieveInductionIfNotInSession(services.inductionService),
+    insertJourneyIdentifier({ insertIdAfterElement: 3 }), // insert journey ID immediately after '/prisoners/:prisonNumber/induction' - eg: '/prisoners/A1234BC/induction/473e9ee4-37d6-4afb-92a2-5729b10cc60f/hoping-to-work-on-release'
+  ])
+  router.use('/prisoners/:prisonNumber/induction/:journeyId', [
+    setupJourneyData(journeyDataService),
```
nathanrussell-moj-digital, line 51: This is the call to the middleware to setup the journeyData on each page in the Update Induction flow

nathanrussell-moj-digital, line 51: This is the call to the middleware to setup the journeyData on each page in the Update Induction flow

stevomcallister (approved):

PR #1162, opened 2025-04-29
RR-1305 - Fix back nav from Skills back to Previous Work Experience

nathanrussell-moj-digital:
PR to fix the problem with Back navigation losing it's state.
Specifically its going Back from the Skills page to the previous page, which would be one of the prisoner's Previous Work Experience pages.
As far as I know it's only this particular Back navigation that is a problem, and it's because Previous Work Experience is a "bit special" ! 🙄 

Basically the journey is:
* Has the prisoner worked before? Yes/No/Not Relevant
* If the answer is No or Not Relevant then the next screen is Skills
* If the user presses Back on Skills it correctly goes back to "Has the prisoner worked before"

If however on the "Has the prisoner worked before" screen, the user answers Yes then the journey becomes dynamic:
* The next screen is a list of checkboxes asking what types/industries/sectors the prisoner has previously worked in
* Then for each checked work experience type, the user is presented with a detail screen asking "what was the prisoners role and duties in the construction industry", then "what was the prisoners role and duties in the health industry" etc

This dynamic page flow causes some problems, so basically at the end of "what types/industries/sectors have they worked in?" we create a `PageFlow` object which acts like a queue - a queue of pages that we need to present to the user. And the queue has some state and logic to determine if you are at the end of the queue in which case the next screen is Skills.

It works in a forward flow, and thats great!

But the problem here is that the user has moved forwards through this dynamic flow, and landed on the Skills page. The dynamic flow has ended and we are back in the regular static Create Induction flow. 
And then the user clicks Back. This takes them back through their browser history, and correctly displays the previous page, which would be the last "previous work detail" screen from the dynamic flow.

And then they click the green Continue button to move forwards again. But because the dynamic flow has ended the code doesnt know where you are in the dynamic flow, so doesnt know which page to route you to. And so you get the error page. Sad times 😢

### DIFF
--- a/integration_tests/e2e/induction/createInductionNavigateBackFromSkillsToPreviousWorkExperience.cy.ts
+++ b/integration_tests/e2e/induction/createInductionNavigateBackFromSkillsToPreviousWorkExperience.cy.ts
@@ -1,0 +1,211 @@
+import { startOfToday, sub } from 'date-fns'
+import Page from '../../pages/page'
+import OverviewPage from '../../pages/overview/OverviewPage'
+import HopingToGetWorkValue from '../../../server/enums/hopingToGetWorkValue'
+import AffectAbilityToWorkPage from '../../pages/induction/AffectAbilityToWorkPage'
+import AbilityToWorkValue from '../../../server/enums/abilityToWorkValue'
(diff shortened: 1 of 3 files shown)


stevomcallister (approved):
Nice

PR #1209, opened 2025-06-26
RR-1635 - Remove delay between characters when cypress "types" into fields

nathanrussell-moj-digital:
PR to remove the delay between characters when cypress "types" into fields - this has reduced the time it takes for the cypress tests to run by over 1 minute 👍

### DIFF
--- a/integration_tests/pages/goal/ArchiveGoalPage.ts
+++ b/integration_tests/pages/goal/ArchiveGoalPage.ts
@@ -17,7 +17,7 @@ export default class ArchiveGoalPage extends Page {
   }
 
   enterReason(value: string): ArchiveGoalPage {
-    this.reasonOther().type(value)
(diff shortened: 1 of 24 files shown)


integration_tests/pages/goal/ArchiveGoalPage.ts
```diff
@@ -17,7 +17,7 @@ export default class ArchiveGoalPage extends Page {
   }
 
   enterReason(value: string): ArchiveGoalPage {
-    this.reasonOther().type(value)
+    this.reasonOther().type(value, { delay: 0 })
```
nathanrussell-moj-digital, line 20: All changes in this PR are the same thing - in every place where we call the cypress `.type` method we are now specifying the optional options object with the delay (between each "key press") set to 0

PR #1273, opened 2025-09-17
RR-1849 - Simplify empty states for Functional Skills

nathanrussell-moj-digital:
This PR is part of the story that reads Functional Skills from Curious APIs as recorded in Curious 1 and Curious 2

Specifically this PR is about simplifying the empty states as thankfully the designs are a little simpler in this respect now.

Previously, when displaying Functional Skills, there were some pages (but not all! 🙄 ) that would ALWAYS display English and Maths, even if the prisoner had not taken those assessments. IE. they were "special" empty states - rather than simply saying "no functional skills recorded", we would render "English - not recorded in Curious" and "Maths - not recorded in Curious"

This odd empty state logic was slightly complicated, not least because it wasnt consistent on all screens! Some screens did it, some didnt!

The good news is that the new designs now simplify this - on all screens where we show Functional Skills, there is a simple "no functional skills" empty state message.

This PR implements the new simpler empty states, and a couple of minor content changes that are also in the new Functional Skills designs.


srowlands-moj (approved):
LGTM

PR #1286, opened 2025-09-22
RR-1851 - Refactor types and mapper to map ALN assessments from Curious

nathanrussell-moj-digital:
PR to refactor the mapper that maps Curious Assessment data to LDD and ALN assessments

The refactoring covers some naming changes (of the types, mapper function names, service function name, middleware function name) to better reflect the purpose/content of the data, and to bring it inline with the naming from SAN (which essentially is all being modelled on now)

The mapper now maps the Curious Assessments to both LDD and ALN assessments, in exactly the same way that SAN does.

At the moment only the LDD assessments are rendered on screen. I'll update the nunjucks to render both the LDD and ALN assessments in my next PR


server/@types/viewModels/index.d.ts
```diff
@@ -64,9 +65,9 @@ declare module 'viewModels' {
   /**
    * Collates the Additional Learning Needs (ALN) and Learning Difficulties and Disabilities (LDD) assessments that have been recorded in Curious
    */
-  export interface PrisonerSupportNeeds {
```
nathanrussell-moj-digital, line 67: `PrisonerSupportNeeds` was the original name for this type when
a) the LWP screen was called Support Needs
b) it only contained LDD assessments
c) ALN assessments were not a thing (this was written long before Curious 2 introduced ALN assessments)

It was a bad name because the data never described a "support need". It described assessments, which is not a "need" in itself

srowlands-moj (approved):
LGTM

PR #1572, opened 2026-04-27
RR-2429 - Corrected print rendering of links & sensitive info

nathanrussell-moj-digital:
PR to corrected print rendering of links & sensitive info (specifically staff names)

### DIFF
--- a/assets/scss/print.scss
+++ b/assets/scss/print.scss
@@ -13,6 +13,14 @@
     display: inline;
   }
 
+  a {
+    text-decoration: none !important;
+    color: inherit !important;
+  }
+  a[href]:after {
+    content: none !important;
+  }
+
   .govuk-hint,
   .govuk-tag,
   .govuk-caption-l,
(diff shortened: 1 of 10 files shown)


assets/scss/print.scss
```diff
@@ -13,6 +13,14 @@
     display: inline;
   }
 
+  a {
```
nathanrussell-moj-digital, line 16: In the print style sheet, change how anchor links are rendered on the page. Basically we need to override the styles from framework (and browser defaults), by not decorating links with an underline, not making the blue, and not displaying the href in brackets after the link text.

server/views/pages/lrsQualifications/index.njk
```diff
@@ -100,7 +100,7 @@
           <p class="govuk-body" data-qa="not-matched-in-lrs-message">
             No results. Unique learner number (ULN) has not yet been matched or learner declined to share their LRS data.
             {% if userHasPermissionTo('USE_DPS_LEARNER_RECORD_MATCHING_SERVICE') %}
-              <a href="{{ matchLearnerRecordUi }}" target="_blank" class="govuk-link" data-qa="link-to-lrs">Check a record or match a learner in LRS (opens in a new tab).</a>
+              <a href="{{ matchLearnerRecordUi }}" target="_blank" class="govuk-link govuk-!-display-none-print" data-qa="link-to-lrs">Check a record or match a learner in LRS (opens in a new tab).</a>
```
nathanrussell-moj-digital, line 103: Add the `govuk-!-display-none-print` class, so that the whole thing is not printed at all (ie. in this case its more than just not styling it like a link, in this case we dont want to print it at all because the text content is an instruction that makes no sense on a printed page)

ayoMoj (approved):

PR #1590, opened 2026-05-21
RR-2618 - Fix rendering of zero counts for education

nathanrussell-moj-digital:
PR to fix a small bug re: the rendering of educational qualification counts where the prisoner does not yet have an induction (which is where the education records are stored)

Before the bug fix:
image
notice that "Other educational qualifications" does not have a count at all

With this fix:
Screenshot 2026-05-21 at 09 03 48

### DIFF
--- a/server/views/pages/overview/partials/overviewTab/_newEducationAndTrainingSummaryCard.njk
+++ b/server/views/pages/overview/partials/overviewTab/_newEducationAndTrainingSummaryCard.njk
@@ -108,7 +108,7 @@ Data supplied to this template:
             <div class="summary-count-container">
               {% if verifiedQualifications.isFulfilled() %}
                 {% set verifiedQualifications = verifiedQualifications.value %}
-                {% set qualifications = verifiedQualifications.qualifications | filterArrayOnProperty('source', 'AO') %}
+                {% set qualifications = verifiedQualifications.qualifications | default([]) | filterArrayOnProperty('source', 'AO') %}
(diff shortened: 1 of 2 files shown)


server/views/pages/overview/partials/overviewTab/_newEducationAndTrainingSummaryCard.njk
```diff
@@ -108,7 +108,7 @@ Data supplied to this template:
             <div class="summary-count-container">
               {% if verifiedQualifications.isFulfilled() %}
                 {% set verifiedQualifications = verifiedQualifications.value %}
-                {% set qualifications = verifiedQualifications.qualifications | filterArrayOnProperty('source', 'AO') %}
+                {% set qualifications = verifiedQualifications.qualifications | default([]) | filterArrayOnProperty('source', 'AO') %}
```
nathanrussell-moj-digital, line 111: As shown in the unit test below, `verifiedQualifications`, `curiousInPrisonCourses` and `educationDto` all have the potential to be null, which is how this bug manifests itself.

These 3 properties are all populated by the route handler from calls to the relevant APIs. For each of the API calls we handle a 404 as meaning the prisoner has no data. IE. it is not an error, it simply means there is no data in the relevant API for the prisoner. The route handlers / service calls honour this - if the API returns a 404, it hands it back up the stack as a null, but crucially NOT an error.

The bug was that in each of the 3 lines modified here, if the property was null, it was essentially a NPE. But the nunjucks processing doesn't error in this case, it silently returns it as `undefined`

The fix is simply to use the nunjucks `default` filter 👍

ayoMoj (approved):

PR #1602, opened 2026-06-03
RR-2657 - Addition of feature toggle to enable PES specific features

nathanrussell-moj-digital:
PR to add a new feature toggle to the LWP UI service to enable/disable features related to the CIAG education contracts (PEF vs PES)

In keeping with the toggle in the API (which exists for the same reason), I've used the helm env var name `CIAG_KPI_PROCESSING_RULE`. This should help future teams enable the PES features on 01/09/26 as they will be dealing with the same env var in both the API and UI (it is likely we will have rolled off by then, so am trying to make life easier for the next team 👍 )

At the moment nothing is using this new feature toggle in the UI - this is to support the next piece of work that I am about to start


helm_deploy/hmpps-education-and-work-plan-ui/values.yaml
```diff
@@ -59,6 +59,7 @@ generic-service:
     AUDIT_SERVICE_NAME: "hmpps-education-and-work-plan"
     AUDIT_ENABLED: "true"
     LRS_INTEGRATION_ENABLED: true
+    CIAG_KPI_PROCESSING_RULE: "PEF" # Set to PEF or PES depending on Curious support
```
nathanrussell-moj-digital, line 62: Same env var as the API toggle - will make it easier for the next team as they will need to set the same env var to the same value in both the API and UI (rather than messing about with 2 differently named env vars that are for the same purpose)

ayoMoj (approved):

